PR #19, opened 2019-03-20
Electric Coin Company

rex4539:


gmale:
Thanks, that's a good catch. Some of that wording was written before the name change.

As you're probably aware, the tip of development is in the `preview` branch and [I've made this change there](https://github.com/zcash/zcash-android-wallet-sdk/commit/af998e132d63b4305ed4cc56b0a83d518e30c431#diff-04c6e90faac2675aa89e2176d2eec7d8R24) but it is good to have the change on `master` as well.

gmale:
LGTM. Merging [per comments](https://github.com/zcash/zcash-android-wallet-sdk/pull/19#issuecomment-478112279).

PR #37, opened 2019-05-24
Rename JniConverter to Librustzcash

gmale:
Turns out JNI classes are hard to mock due to the way sytem libraries
are loaded. So it became easier to put the JniConverter class behind an
interface. Once that refactor was necessary, it was a good time to
update the name of this class because it retained its original PoC name
for far too long.


str4d:
I don't think this will work, because the backend JNI methods were not renamed. I'll address this in #29.

PR #40, opened 2019-07-26
Configure mainnet SDK flavours to use mainnet Rust feature flag

str4d:


build.gradle
```diff
@@ -169,6 +169,16 @@ cargo {
...
+
+    variants {
+        zcashmainnetDebug {
+            defaultFeaturesAnd("mainnet")
+        }
+
+        zcashmainnetRelease {
+            defaultFeaturesAnd("mainnet")
+        }
+    }
```
gmale: what do you think about a dsl along the lines of:

```
variants {
    zcashmainnet {
         features "mainnet"
    }
}
```

in other words, allow for configuring just a `productFlavor` or just a `buildType` and then also assuming that the default feature is used, by default (so we can maybe shorten the name), and if you want to do something other than the common behavior, THEN you have to jump through hoops?

str4d: That's definitely a doable DSL, it just takes a lot more work to implement (both in terms of engineering, and defining how overrides interact or replace each other - currently it's a straight override, with the per-variant config completely replacing the equivalent general config). See https://gitlab.com/creati8e/badge for the kind of DSL I'd use for this.

As for the feature configuration, what I actually want to end up with is this kind of DSL:

```
// Feature "std" is set by default, matching how the Rust ecosystem handles no-std
cargo {
    features {
        defaultAnd("bar")
    }
    variants {
        nostd {
            features {
                noDefaultBut("bar")
            }
        }
    }
}
```

i.e. the per-variant or per-flavour configuration is exactly the same syntax you'd use at the top level. But that requires nested DSL inside a `NamedDomainObjectContainer`, which I haven't figured out yet.

str4d: Regarding the features DSL itself, there are three underlying flags we are trying to represent:
```
$ cargo help build
...
        --features <FEATURES>       Space-separated list of features to activate
        --all-features              Activate all available features
        --no-default-features       Do not activate the `default` feature
```

`--all-features` will always appear on its own, while `--features` will appear with or without `--no-default-features`. This leads to three options in the current DSL: `all`, `defaultAnd(...)` and `noDefaultBut(...)`.

The plugin upstream uses a `features { ... }` block to contain the three options, but currently doesn't prevent you from setting multiple options at once. AFAICT whichever one you set last takes precedence.

I'm not sure how I feel about moving the options out of a `features { ... }` block generally. On the one hand, a fully-implemented setup would ensure we only use one of the options at a time, so the nesting isn't doing much. On the other hand, what it _is_ doing is providing better local context about the setting. Also, if we used a bare `features ...` like you suggest, then we'd need a separate setting for no default features. Upstream may have opinions about this as well.

Here are some alternatives that we can stare at for a bit:
```
cargo {
    defaultFeaturesAnd("bar")

    productFlavors {
        nostd {
            noDefaultFeaturesBut("bar", "spam")
        }
    }
}
```
```
# Have variants replace general "features" setting with their own?
cargo {
    features("bar") // "std" and "bar" enabled

    productFlavors {
        nostd {
            noDefaultFeatures
            features("spam") // "spam" enabled
        }
    }
}
```
```
# Have variants be purely-additive, which might be too limiting?
cargo {
    features("bar") // "std" and "bar" enabled

    productFlavors {
        nostd {
            noDefaultFeatures
            features("spam") // "bar" and "spam" enabled
        }
    }
}
```

PR #43, opened 2019-08-28
[WIP] Use a TLS connection.

defuse:


gmale:
Per discussion earlier this week, we'll add this change to the latest code and include gradle support to enforce TLS on mainnet builds.

PR #47, opened 2019-08-30
add empty swift_prefix option value to .proto files

pacu:
Motivation.

proto-swift-gen script adds the package prefix on a snake case fashion to Swift Struct files. 
This makes the code difficult to read and has no convenience to us.

Example:
`CompactBlock` -> `Cash_Z_Wallet_SDK_RPC_CompactBlock`

When adding `option swift_prefix = "";` the resulting struct Name is `CompactBlock`

### DIFF
--- a/src/main/proto/WalletData.proto
+++ b/src/main/proto/WalletData.proto
@@ -1,4 +1,5 @@
 package cash.z.wallet.sdk.proto;
+option swift_prefix = "";
 message WalletData {
   required string name = 1;
   required int32 id = 2;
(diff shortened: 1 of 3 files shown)


gmale:
Looks good to me. Only question, do you think an empty string prefix would introduce any risk for collisions with wallet maker's existing name spaces? Or are swift dependencies combined in a way where that can't happen?

pacu:
> Looks good to me. Only question, do you think an empty string prefix would introduce any risk for collisions with wallet maker's existing name spaces? Or are swift dependencies compiled in a way where that can't happen?

If their name collide they would refer to the Module's name first themselves. 

Let's say that we chose that our SDK is branded "SwiftZCash". 
they would have to go with SwiftZCash.CompactBlock

Issue #30, opened 2019-05-15
Internal-Security-Finding #1: API Secure-Usability Suggestions


defuse:
Here are some suggestions for making the SDK's API harder to use in an insecure way:

1. Suppose light clients are connecting to `lightwalletd` over Tor. `Synchronizer`'s constructor allows you to specify `batchSize`, the number of compact blocks to download at a time, and `blockPollFrequency`, how often to poll for new blocks. If different wallets set these differently, it fragments the anonymity set of Tor users connecting to that `lightwalletd`, since the different wallets will have different externally-visible behavior.
2. The API could avoid returning stale balances if, for example, by balance could return an enum-like type CurrentBalance(balance) or StaleBalance(balance), or whatever equivalent feature Kotlin has, or even just another field for stale balance in `WalletBalance`. This way it's impossible to accidentally ignore the `isStale` flag.
3. This probably isn't practical to do because, as the README mentions, wallets will want to manage their own seeds, but it would be safer if it were impossible to construct objects with insecure seeds (i.e. all instances should either be created via secure randomness or via a restore of an export of one created via secure randomness).
4. The SDK's user is in control of `paramDestinationDir` and `dataDbPath`. They *could* choose a really insecure location for these things, for example a shared directory on the SD card. They should be in a directory that's readable/writable only by the app, otherwise other things on the system could backdoor the parameters or mess with the database. It might be worth checking the permissions, like how `ssh` checks the permissions on `~/.ssh/`.

lindanlee:
Actual specifics of this might be out of date.

gmale:
From security team, "This is about making the easiest way to do things the most secure way."

gmale:
All items have been addressed in the latest code on the `preview` branch.

1. `bacthSize` and `blockPollFrequency` are no longer configurable
2. `isStale` flag no longer exists and the synchronizer now scans and validates before showing data
3. librustzcash has made changes to enforce seed length (https://github.com/zcash/librustzcash/pull/128)
4. `paramDestinatationDir` and `dataDbPath` are no longer configurable and are restricted to directories that are private to the app

defuse:
Nice! We can close this ticket and open another one to do a second review for API usability. For that it'll help to have a sample app that uses the newest API.

PR #49, opened 2019-09-26
Reduce configurability per Security Finding 1.

gmale:
These changes largely reduce the amount of configuration that can be tweaked in order to prevent fragmentation of the anonymity set of users. If wallet makers change certain properties than it can become easy to detect which network requests are coming from which client. The goal is for clients to be as anonymous as possible.


gmale:
Changes address security finding 1 (https://github.com/zcash/zcash-android-wallet-sdk/issues/30)

Issue #95, opened 2020-01-16
Timing of block downloads leaks whether or not any notes belonging to the wallet were discovered


defuse:
This issue was originally discovered and reported by Florian Tramèr. I'm making it public since no users are currently at risk.

The issue is that the SDK will:

1. Download new blocks,
2. process those blocks, taking an amount of time that depends on how many notes belong to this wallet,
3. wait a fixed length of time,
4. make a request to lightwalletd for more new blocks.

A passive network adversary can measure the time between (1) and (4) to determine whether any notes in the blocks belonged to the wallet. To fix this, either

a. Decouple the network code from the block processing code, such that downloaded blocks are queued up to be processed on a separate thread and the processing won't affect the timing of network requests,
b. Florian's suggestion, to request blocks at absolute time intervals, i.e. instead of sleeping for 20 seconds between requests, request new blocks at times \*:00, \*:20, \*:40. *In order for this to not leak information, the block processing MUST always complete before the next request is due to be made.*

I recommend (b) since it's easier and we're about to run into even harder side-channel challenges with memo downloading and mempool access.

gmale:
Until we change the way data access works, I'm concerned about option A because it could lead to concurrent access of the SQLite file holding the cached blocks.

Since the Rust side of the logic (scanning) and the Kotlin side both use the `*.db` file and a write occuring at the same time as a read leads to "corrupt database" errors, it has been safer to keep all DB access sequential.

I'd also like to detect blocks as soon as possible after their availability. Would random time be acceptable? Perhaps with some fixed minimum? I'm thinking at least 10 but up to blocktime/2 seconds.

So instead of:

`delay(FIXED_TIME)`

`delay(FIXED_TIME + Random.nextInt(FIXED_TIME))`

Where FIXED_TIME is around 20 seconds or so.

@defuse thoughts?

ftramer:
I wouldn't recommend the randomized delay approach, due to the following attack:
Say the adversary obtains a public key and wants to know if this key belongs to a certain mobile user. The adversary can create many payments for this public key (even payments of 0 value work), and then time the SDK's delay when processing each of them. This averages out the randomized delay and reveals the delay incurred when processing a payment.

gmale:
Thank you for the feedback @ftramer this information is **_very_** helpful! I'd like to implement changes immediately but I have some questions, to help me understand this correctly:

> and then time the SDK's delay when processing each of them

1. How might this work in practice on a mobile device?

[...]

PR #96, opened 2020-01-17
Bugfix/block timing

gmale:
Address #95 by not exposing computation time while polling the network for changes.

Note, the issue is most prevalent while "in the steady state" of polling for the latest block. The app downloads all missing blocks on launch and then processes them without downloading again until the next interval boundary.

### DIFF
--- a/src/main/java/cash/z/wallet/sdk/block/CompactBlockProcessor.kt
+++ b/src/main/java/cash/z/wallet/sdk/block/CompactBlockProcessor.kt
@@ -265,7 +265,20 @@ class CompactBlockProcessor(
         downloader.rewindToHeight(lowerBound)
     }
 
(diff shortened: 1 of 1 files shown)


gmale:
@tframer We'd love to get your feedback on this.

ftramer:
This looks reasonable to me.
I'd make to include tests for some edge cases, e.g., if you were to somehow land on an interval boundary but at a glance it seems like this is handled fine.
I guess the polling interval is chosen with a sufficient margin so that the SDK would always be able to process all transactions it receives well before the beginning of the next interval?

gmale:
The related [changes for this live on master](https://github.com/zcash/zcash-android-wallet-sdk/blob/master/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt#L489-L503) and the basic approach has been deemed reasonable. Based on that, I'll close this and we can always re-open a new issue if the problem is not sufficiently addressed.

Issue #99, opened 2020-02-03
Update Demo Apps to Work with Current Version of SDK


braddmiller:
The changes to the SDK for the most recent version of the dogfooding app have broken some of the existing demo apps. We should update the demo apps to be fully functional to support the developer community and allow creation of documentation.

Suspected broken elements
- [ ] Convert from "testalice..." seed to mnemonic seed (3)
- [ ] Should be moved to the latest librustzcash (note-spending-v7) and SDK (8) **This needs research and broken down**
- [ ] Change how initialization works (step one, get the seed, step two, feed into SDK) (5)

PR #108, opened 2020-02-26
Feature/demo repair

gmale:
Restore the demo app to working order and also add some new features, most notably, mainnet support.


gmale:
closes #99

PR #109, opened 2020-02-26
Make update_sapling_tree use TLS

defuse:


defuse:
At first I tried to use `tls_api_native_tls`, but grpc [depends on tls-api = "0.1"](https://github.com/stepancheg/grpc-rust/blob/v0.6/grpc/Cargo.toml), and trying to use tls-api-native-tls 1.0 lead to an OpenSSL-related compliation error:

```
error: failed to run custom build command for `openssl v0.9.24`

Caused by:
  process didn't exit successfully: `/home/firexware/zcash/zcash-android-wallet-sdk/target/release/build/openssl-b7cc50664799de27/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'Unable to detect OpenSSL version', /home/firexware/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.9.24/build.rs:16:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

warning: build failed, waiting for other jobs to finish...
```

rustls 1.0 worked so I went with that.

In any case @str4d should review this because I have little clue what I'm doing when it comes to rust dependency graphs :-)

PR #112, opened 2020-02-27
Rust fixes

str4d:


gmale:
looks reasonable, I'll test this in the morning.

gmale:
Works against `lightwalletd.z.cash:9067` and @adityapk00's `lightd-main.zecwallet.co:443` but does not work against a server running locally. Just fails with error:

>Error: Grpc(Http(ClientDied(None)))

I'm okay with this because at least I can make checkpoints with TLS servers. If the local server issue is low-hanging fruit, then maybe it is worth addressing. Otherwise, LGTM.

gmale:
After chatting with @str4d it seems likely that my local issue is a result of me being lazy and not adding a certificate with a root of trust. So this code is fine to merge.

Summary of our chat:

> str4d  1:30 PM
> And are you using TLS locally?
> 
> Kevin 1:30 PM
> yeah (I just dropped the -no-tls-very-insecure flag)
> 
> str4d  1:31 PM
> Then the problem is likely that your local TLS certificate doesn't have a root of trust
> (in the set used by rustls, which is what Taylor set things up to use)
> 
> Kevin  1:32 PM
> ok then that sounds acceptable
> 
> str4d  1:32 PM
> You might have luck using https://mkcert.dev (edited)

PR #122, opened 2020-04-22
Initial ci.md

benzcash:
Initial targets for CI stages.

We'll get more definition on the details once the goals look good.


benzcash:
Any secrets required to perform the action should also be noted.

PR #172, opened 2020-10-21
Refactor to updated data access API.

nuttycom:


src/main/rust/lib.rs
```diff
@@ -749,7 +782,7 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_createToAdd
     env: JNIEnv<'_>,
     _: JClass<'_>,
     db_data: JString<'_>,
-    consensus_branch_id: jlong,
+    _: jlong, // was: consensus_branch_id; this is now derived from target/anchor height
```
gmale, line 785: :clap:

gmale:
This looks great. After a quick visual scan, I think this will compile and run, directly, in the app. I'll give that a try.

From there, one of the next steps will be to move [these databases](https://github.com/zcash/zcash-android-wallet-sdk/pull/172/files#diff-f04fcd56deba3d69c1c149fe6fffcc49c46ef45d1b6289ee0124789e0df519ecR86-R94) out of rust and into kotlin and slowly get rid of the sqlite crate :crossed_fingers:

src/main/rust/lib.rs
```diff
@@ -89,6 +77,22 @@ fn print_debug_state() {
     debug!("Release enabled (congrats, this is NOT a debug build).");
 }
 
+#[cfg(feature = "mainnet")]
+pub const NETWORK: MainNetwork = MAIN_NETWORK;
+
+#[cfg(not(feature = "mainnet"))]
+pub const NETWORK: TestNetwork = TEST_NETWORK;
+
+fn wallet_db(env: &JNIEnv<'_>, db_data: JString<'_>) -> Result<WalletDB, failure::Error> {
```
gmale, line 86: no need to change anything (because the goal is to get rid of these DBs on the rust side) but just for future reference:

the convention on the sdk side has been to call the `block_db` here "cache_db" and the `wallet_db` here the "data_db"

nuttycom: Yeah, I was aware of the existing terminology; this is kind of a suggestion for how to improve it. The `BlockSource` trait is intended to not be specific to being backed by a database - it's callback-oriented, so it could be backed by a stream, for example.

PR #173, opened 2020-10-30
Add ability to erase by alias.

gmale:


gmale:
Addresses partner feedback. Tests pass. Merging.

PR #187, opened 2020-12-01
Clear the warns build

herou:
Added the ExperimentalCoroutinesApi in build.grandle in order to clear some warning builds

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -136,10 +136,11 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
 
-    // ExperimentalCoroutinesApi helps to clear the build warnings
+    // ExperimentalCoroutinesApi and FlowPreview help to clear the build warnings
     kotlinOptions {
         jvmTarget = "1.8"
         freeCompilerArgs += "-Xopt-in=kotlinx.coroutines.ExperimentalCoroutinesApi"
+        freeCompilerArgs += "-Xopt-in=kotlinx.coroutines.FlowPreview"
     }
 
     packagingOptions {
--- a/build.gradle
+++ b/build.gradle
@@ -135,6 +135,13 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+
+    // ExperimentalCoroutinesApi helps to clear the build warnings
+    kotlinOptions {
+        jvmTarget = "1.8"
+        freeCompilerArgs += "-Xopt-in=kotlinx.coroutines.ExperimentalCoroutinesApi"
+    }
+
     packagingOptions {
         exclude 'META-INF/DEPENDENCIES'
         exclude 'META-INF/LICENSE'
--- a/src/main/java/cash/z/ecc/android/sdk/ext/CurrencyFormatter.kt
+++ b/src/main/java/cash/z/ecc/android/sdk/ext/CurrencyFormatter.kt
@@ -19,6 +19,7 @@ import java.util.*
 
 //TODO: provide a dynamic way to configure this globally for the SDK
 // For now, just make these vars so at least they could be modified in one place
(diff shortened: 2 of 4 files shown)


gmale:
Nice work @herou and quick turn-around!

This cuts the warnings down by a lot. After compiling this, it looks like if you also add the similar [`FlowPreview` compiler arg](https://github.com/zcash/zcash-android-wallet/blob/master/app/build.gradle#L90), then it will clear all the remaining coroutine "opt-in" warnings.

From there, if we next add this line:
```
@file:Suppress("NOTHING_TO_INLINE")
```
 to the top of the [CurrencyFormatter](https://github.com/zcash/zcash-android-wallet-sdk/blob/master/src/main/java/cash/z/ecc/android/sdk/ext/CurrencyFormatter.kt) and [Twig](https://github.com/zcash/zcash-android-wallet-sdk/blob/master/src/main/java/cash/z/ecc/android/sdk/ext/Twig.kt) files then it gets rid of about a dozen other warnings (since these are top-level utilities, the inlining is intentional).

Once those things are all done, I'd say look through the rest of the warnings and see if you can spot any others that you think are worth fixing. Anything that doesn't change the API for downstream wallets or require regression testing would be ideal. For instance, there's a warning about [these files](https://github.com/zcash/zcash-android-wallet-sdk/blob/79184c3cdd23087ed1835ced46df500834a56a84/src/main/java/cash/z/ecc/android/sdk/transaction/WalletTransactionEncoder.kt#L178), technically, being nullable. Changing that line to `if(file.parentFile?.exists() == true)` would clear the warning without breaking anything or requiring regression tests.

Fixing anything along those lines is a huge help because it's exactly the kind of thing that falls through the cracks. Once that's complete, we can start looking at some other tech debt that's more interesting!

herou:
@gmale I committed the required changes, please let me know if it is okay.

gmale:
Hi @herou!
This is great. Thanks for making the contribution!

One tiny thing that I noticed is that you added the `@Suppress("NOTHING_TO_INLINE")` at the object level. If you added it at the file level, instead, it would clear many more of these warnings. For these two utility classes, the inlining is intentional so I think clearing the warnings at the file level makes sense and also signals that we've done this on purpose. Just add `file` and put it as the first line, above the package declaration:

```
@file:Suppress("NOTHING_TO_INLINE")
```

Then, tackle some of the other warnings that you think would be helpful without breaking things. After that, we can merge this and then discuss some ideas of what to tackle next.

Thank you so much for you help. I know this seems small but it's a good warm up and also very sorely needed. Those warnings have existed for over a year!

gmale:
I'll do the final math once the PR is merged but so far it looks like warnings have been reduced by well over 50%!!

Originally, there were around 90. I think when this is done, more than 80% will be cleared.

herou:
@gmale You have it, please give more tickets to work on.

herou:
@gmale Please, give me another ticket to work on.

gmale:
Hi @herou! I appreciate the enthusiasm and ever since you asked, I've been thinking about this a lot.

I have been working on a list of potential things that might be good candidates for you. Given that you have RxJava experience, it might make sense for you to take the lead on adding an [SDK module that is more Rx-friendly](https://github.com/zcash/zcash-android-wallet-sdk/projects/1#card-50867892). Or if you want to do something fun, you could work on [useful utilities](https://github.com/zcash/zcash-android-wallet-sdk/discussions/192). Or if you only have a little time, bugs and tests always need attention.

It's really going to depend on what you're most interested in doing and roughly how much time you have available.

But first, we should find a better place to chat than this ticket! I created a new [discussion area](https://github.com/zcash/zcash-android-wallet-sdk/discussions/193). Let's try that out and see if we find it useful.

I also added a [project board](https://github.com/zcash/zcash-android-wallet-sdk/projects) to this repo to begin brain dumping and organizing potential work.

Issue #205, opened 2021-01-16
Add a function to SaplingParamTool for verifying the checksum of sapling parameter files


gmale:
This issue is blocked by #204 because the tool needs to exist first before we can add to it.

**As an** app developer, **I'd like to** validate the checksum of sapling parameter files **so that** I can create a better experience for users when these files are invalid and need to be fetched by informing them.

This issue falls under [the parent issue for param files](https://github.com/zcash/zcash-android-wallet-sdk/issues/201).

### Context
If anything goes wrong while downloading the param files, then the simplest fix is to fetch them again. Users can achieve this by clearing the app cache or app developers can do this automatically. Either way, having a function to determine that the files are corrupt (or truncated or otherwise invalid) is useful.

### Implementation Details

- [ ] Expose a function in the RustWelding layer that can perform a checksum of the param files
    - [ ] optional: be able to confirm each file separately so that wallets don't always download both when one is bad
- [ ] Add a function (or two functions such as `validateOutputParamChecksum()` and `validateSpendingParamChecksum()`) to the `SaplingParamTool` that calls this new rust function
- [ ] Update the `SaplingParamTool.validate()` function so that it fails when one or both of the param files have an invalid checksum

Note: Ensuring that the RustBackend is properly loaded at the right time and available to the tool could get tricky.
Note: Hopefully, there is a suitable function in librustzcash that we can call that returns helpful errors that we can map in the SDK. Then this may be as simple as calling `validateSaplingParams` or something similar in the RustBackend.

### Criteria
- A function exists in SaplingParamTool that returns false (or throws an exception) when the param files have an invalid checksum
- Behavior mirrors DerivationTool--functions are called statically
- The `SaplingParamTool.validate()` function fails when the param files exist but have an invalid checksum. Currently, [the function only checks that the files exist](https://github.com/zcash/zcash-android-wallet-sdk/blob/96da0a9beaa92597bd4abdd74dce7c3098fbb86d/src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt#L106-L115).

ccjernigan:
This was implemented by Honza.

Issue #203, opened 2021-01-16
WIP: Correct param file download behavior when storage is low


gmale:
This ticket is a place holder to fix any issues that surface during testing #202, which is only concerned with laying the groundwork of unit tests.

This issue falls under [the parent issue for param files](https://github.com/zcash/zcash-android-wallet-sdk/issues/201). It is not yet ready for work.

Pending tasks:

- Handle error gracefully
  - create (or leverage) a typed exception for storage issues
  - assign an error number
  - craft a user-facing error message
- [This logic almost certainly needs to change](https://github.com/zcash/zcash-android-wallet-sdk/blob/2415bac043e8fd731f307699b312a1774fcc0d45/src/main/java/cash/z/ecc/android/sdk/transaction/WalletTransactionEncoder.kt#L175-L189). The response is probably successful but the file write fails.

Issue #202, opened 2021-01-16
Test: Param download behavior when storage is low


gmale:
**As a** library developer, **I'd like to** test what happens when the library attempts to download parameter files on a device that does not have enough storage space **so that** I can confirm that the library exposes this problem clearly and as early as possible.

This issue falls under [the parent issue for param files](https://github.com/zcash/zcash-android-wallet-sdk/issues/201).

### Expected behavior

Note: the param files are expected to exist in the `params` folder of the cache directory:
```kotlin
"${context.cacheDir.absolutePath}/params"
```

## Test Case 1a:
**Given** a device where neither param file exists
**When** the validate function is called
**Then** it returns false

## Test Case 1b:
**Given** a device where the `Spend` params file exists in the expected directory but the `Output` params file does not
**When** the validate function is called
**Then** it returns false

## Test Case 1c:
**Given** a device where the `Output` params file exists in the expected directory but the `Spend` params file does not
**When** the validate function is called
**Then** it returns false

## Test Case 2:
**Given** a device with insufficient storage for both files
**When** the parameter files are downloaded
**Then** it fails with a descriptive error (such as "insufficient storage")

## Test Case 3:
**Given** a device with sufficient storage for only one of the two param files
**When** the parameter files are downloaded
**Then** it fails with a descriptive error (such as "insufficient storage")
&nbsp;&nbsp;  **And** the smaller of the two files has been successfully downloaded

### Criteria
- Repeatable tests exist that confirm the expected behavior
- It is okay for these tests to FAIL right now. **Correcting failures is out of scope for this ticket.**
- Ideally, this test can run on any device, regardless of whether the device has low storage space
  - One way to achieve that might be to mock a call that checks the available disk storage, beforehand

gmale:
Note: this test is a decent example to follow. It's the most recent test written in the SDK:

https://github.com/zcash/zcash-android-wallet-sdk/blob/master/src/androidTest/java/cash/z/ecc/android/sdk/integration/service/ChangeServiceTest.kt

herou:
@gmale assign it to me when you can.

gmale:
to create a context:

```kotlin
val context: Context = InstrumentationRegistry.getInstrumentation().context
```

herou:
@gmale You have it: #211

herou:
@gmale This needs to be closed.

PR #206, opened 2021-01-17
Create SaplingParamTool and add a function for clearing the param files from cache

herou:


src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
```diff
@@ -0,0 +1,134 @@
...
+         * necessary for the wallet to scan blocks.
+         *
+         * @return an http client suitable for downloading params data.
+         */
+        private fun createHttpClient(): OkHttpClient {
+            //TODO: add logging and timeouts
+            return OkHttpClient()
+        }
+    }
+}
```
gmale, line 134: minor: needs newline

gmale:
After a quick first pass, this looks good to me. Just need to get it compiling and test it out.

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
herou: Which number line needs a new line?

herou:
@gmale The only thing missing is to write unit tests for both methods. I am not sure what kind of parameter the methods needs.

src/androidTest/java/cash/z/ecc/android/sdk/tool/SaplingParamToolTest.kt
```diff
@@ -0,0 +1,19 @@
...
+import junit.framework.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class SaplingParamToolTest {
+
+    @Test
+    fun validateParamTest(){
+        val pathDir = "pathParamsDir"
```
gmale, line 13: This is fine for now, we can flesh this test out more as part of https://github.com/zcash/zcash-android-wallet-sdk/issues/202

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
```diff
@@ -0,0 +1,135 @@
...
+                        twig("directory exists!")
+                    } else {
+                        twig("directory did not exist attempting to make it")
+                        file.parentFile.mkdirs()
+                    }
+                    Okio.buffer(Okio.sink(file)).use {
+                        twig("writing to $file")
+                        it.writeAll(response.body().source())
+                    }
+                    twig("fetch succeeded, done writing $paramFileName")
```
gmale, line 78: Directly copying these functions over is exactly the right thing to do for this github issue (which is only concerned with creating the SaplingParamTool file). 

We can improve the logic here as part of https://github.com/zcash/zcash-android-wallet-sdk/issues/203 by wrapping this entire part of the if statement in a `try` block, then updating the failureMessage in the `catch` block.

However, lets save that for a separate commit as part of working on #203. For now, we can just take it one step at a time, as you've already done.

Nice work.

build.gradle
```diff
@@ -54,7 +54,7 @@ apply plugin: 'org.jetbrains.dokka'
 apply plugin: 'com.google.protobuf'
 apply plugin: 'com.github.ben-manes.versions'
 apply plugin: 'com.github.dcendents.android-maven'
-apply plugin: 'com.getkeepsafe.dexcount'
+//apply plugin: 'com.getkeepsafe.dexcount'
```
gmale, line 57: I don't think we should commit these changes as part of this PR.

I only commented them out because I was in a hurry and the plugin was having a problem and it's not necessary for building. We can correct or remove this plugin as part of a separate issue.

For now, we can leave these as local changes and reset this `build.gradle` file to keep it out of the PR.

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
```diff
@@ -0,0 +1,135 @@
...
+            ).forEach { paramFileName ->
+                val file = File(destinationDir, paramFileName)
+                if(file.exists()){
+                    twig("Files does  exist!")
+                    return true
+                }
+            }
+            twig("Error: File does not exist!")
+            return false
+        }
```
gmale, line 119: It appears that this will return true if one file exists. We probably want to change this so that it only returns true when both files exist. The simplest way to do that would probably be to elevate the return with something like this:

```suggestion
    fun validate(destinationDir: String): Boolean {
        return arrayOf(
            ZcashSdk.SPEND_PARAM_FILE_NAME,
            ZcashSdk.OUTPUT_PARAM_FILE_NAME
        ).all { paramFileName ->
            File(destinationDir, paramFileName).exists()
        }.also {
            twig("Param files${if(!it) "did not"} both exist!")
        }
    }
```

gmale (changes requested):
Did a closer review. Looks good. 

Minor changes requested. Most other things can be saved for other open issues.

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
```diff
@@ -0,0 +1,135 @@
...
+                    if(file.deleteRecursively()){
+                        twig("Files deleted successfully")
+                    }else{
+                        twig("Error: Files not able to be deleted!")
+                    }
+                }
+            }
+        }
+
+        fun validate(destinationDir: String): Boolean{
```
pacu, line 106: if no further validations will be added I would consider renaming the function to express clear intent of what's being validated (existence, for now)

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
```diff
@@ -0,0 +1,135 @@
...
+        //
+        /**
+         * Http client is only used for downloading sapling spend and output params data, which are
+         * necessary for the wallet to scan blocks.
+         *
+         * @return an http client suitable for downloading params data.
+         */
+        private fun createHttpClient(): OkHttpClient {
+            //TODO: add logging and timeouts
+            return OkHttpClient()
```
pacu, line 132: Does this enforce HTTS by default? It's 2021 so I assume it does.

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
pacu: Can by any chance the user manually delete one of the files? What's the action to take in that case? if no action is being taken, quick fix could be 
```

if ! validate(path) {
  clear(path)
  fetchParams(path)
}
```

pacu (changes requested):

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
gmale: Good point. Additional validations will be added in https://github.com/zcash/zcash-android-wallet-sdk/issues/205

build.gradle
herou: Reverted to initial grandle state

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
herou: > It appears that this will return true if one file exists. We probably want to change this so that it only returns true when both files exist. The simplest way to do that would probably be to elevate the return with something like this:

Committed your suggestions, made a few modifications because the code could not compile.

build.gradle
gmale: It appears that you may have reverted a bit too far because instead of having no changes to the `build.gradle` file, this PR now has [changes](https://github.com/zcash/zcash-android-wallet-sdk/pull/206/commits/023d1fb079e9a719c765a1c1a549cbad4ed36ee1#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7) to undo the [two most recent commits to that file](https://github.com/zcash/zcash-android-wallet-sdk/commit/9bf087bed801cec75421ba22727f05944081e89e#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7).

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
gmale: sorry for the compile issue. I was typing the code on github but didn't test it. Looks like it just needed the `else` clause. Adding `else ""` would fix that:

```kotlin
.also {
    println("Param files${if(!it) "did not" else ""} both exist!")
}
```

The [code changes here](https://github.com/zcash/zcash-android-wallet-sdk/pull/206/commits/023d1fb079e9a719c765a1c1a549cbad4ed36ee1#diff-ade3efa42bf3ca434f2d51dedb711f797fd77ba76dd65d44968e0f52079878e3R107-R113) would result in two log lines, when a file is missing:

> Param files
> did not

Since we probably want one log line, instead of two for this, I would suggest either adding the `else ""` code from above or maybe simplifying the existing code to the following:

```kotlin
            }.also {
                if(it)
                    twig("Both param files exist!")
                else
                    twig("Both param files did not exist!")
            }
```

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
```diff
@@ -0,0 +1,135 @@
...
+         *
+         * @param destinationDir the directory where the params should be stored.
+         */
+        suspend fun ensureParams(destinationDir: String) {
+            var hadError = false
+            arrayOf(
+                ZcashSdk.SPEND_PARAM_FILE_NAME,
+                ZcashSdk.OUTPUT_PARAM_FILE_NAME
+            ).forEach { paramFileName ->
+                if (!File(destinationDir, paramFileName).exists()) {
```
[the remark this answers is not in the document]
gmale: Good question @pacu . Previously, the directory we were passing to these functions was controlled by the SDK. Now that this tool exposes that behavior we should do something to ensure that the developer is passing the correct directory (rather than putting param files in an area that is not private to the app, which is a [potential security issue](https://github.com/zcash/zcash-android-wallet-sdk/issues/30))

We should add some protections that ensure the developer is using the right directory but I think that maybe falls outside the scope of this PR. I'd like to punt and give it some thought and maybe run it by Taylor and then immediately apply a change after this is merged.

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
```diff
@@ -0,0 +1,135 @@
...
+                    twig("ERROR: $paramFileName not found at location: $destinationDir")
+                    hadError = true
+                }
+            }
+            if (hadError) {
```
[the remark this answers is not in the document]
gmale: good point. my [previous comment](https://github.com/zcash/zcash-android-wallet-sdk/pull/206/files/d76e9b4dd13d9df5a623497d275529fa8b6b05af#r561954946) would address this. We basically need to validate the input parameter for these functions.

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
gmale: I believe it will fetch any URL that it is provided. We could configure it to be more restrictive or just make sure this tool validates what's downloaded. Librustzcash already verifies the checksum of the params before using them. The URL of the params is controlled by the SDK. So I think we're probably okay to leave this as is.

src/main/java/cash/z/ecc/android/sdk/tool/SaplingParamTool.kt
gmale: Resolved by https://github.com/zcash/zcash-android-wallet-sdk/pull/206/commits/ccd1464bc0fc6b4f2bdc432e9eb224717a1f31f1

Issue #185, opened 2020-11-30
Add KLint to the project


gmale:
Pinterest has a [great library](https://github.com/pinterest/ktlint) for linting the code and enforcing code style. It would be good to add this to the SDK.

nighthawk24:
I can take care of this.

gmale:
Update: chatted w/ @nighthawk24 and he mentioned work on this is under way! There might be merge conflicts with changes that are in flight but we'll cross that bridge when we get there.

nighthawk24:
Good to close wrt https://github.com/zcash/zcash-android-wallet-sdk/pull/207

PR #207, opened 2021-01-19
#185 Add ktlint

nighthawk24:
Fix formatting per ktlint rules.
Upgrade dependencies & gradle.

Review request to @gmale 
Please check compilation and tests before approving.


build.gradle
```diff
@@ -295,5 +295,14 @@ taskTree{
 
 defaultTasks 'ciBuild'
 
-preBuild.dependsOn includeDirBugFix
+apply from: "$rootDir/ktlint.gradle"
```
gmale, line 298: We should maybe move this to the [`apply` section](https://github.com/zcash/zcash-android-wallet-sdk/blob/31ab31f1c040ece25e7175555c6f1c46e9dd1e01/build.gradle#L45-L62) of the file

build.gradle
```diff
@@ -295,5 +295,14 @@ taskTree{
...
+
+afterEvaluate {
```
gmale, line 300: do we need `afterEvaluate` here? Isn't the graph of tasks set up during the configuration phase anyway? IIRC, `dependsOn` is evaluated lazily so it's not necessary to do this in the `afterEvaluate` phase.

Also, since the `preBuild` task is going to be added by the plugins that are applied, we don't need to check for its existence because all the `dependsOn` declarations are not evaluated until the task graph is fully defined.

BUT

would it make sense to ignore those questions completely and just use one of the gradle plugins for ktlint instead? Such as:
https://github.com/jeremymailen/kotlinter-gradle
https://github.com/jlleitschuh/ktlint-gradle

ktlint.gradle
```diff
@@ -0,0 +1,25 @@
+repositories {
```
gmale, line 1: we probably don't need to re-add repositories here. The parent gradle file will provide jcenter.

BUT

if we go with the ktlint gradle plugin, this entire file could be omitted.

.editorconfig
```diff
@@ -0,0 +1,14 @@
+# Comma-separated list of rules to disable (Since 0.34.0)
+# Note that rules in any ruleset other than the standard ruleset will need to be prefixed
+# by the ruleset identifier.
+disabled_rules=import-ordering,no-wildcard-imports,unused-import
```
gmale, line 4: Unused import seems useful. At least as a warning.

Import ordering would also probably be nice, going forward. But then it might be weird having all new files adhere to the ordering while old files do not.

Perhaps we can get this PR merged and then decide on an import ordering and then apply that to all files as a separate PR?

gmale (commented):
This is great. There are a lot of changes here and my review was taking a while so I just split out all the things related to setup. I'll continue all the format review (and testing) separately.

Also, as a next step, it would be nice to setup automation on this project so PRs automatically are checked to confirm whether they adhere to expected formatting. Of course, that's outside of the scope of this PR.

Thanks @nighthawk24 for doing all this. It has been in the backlog for a VERY LONG TIME!

build.gradle
nighthawk24: > would it make sense to ignore those questions completely and just use one of the gradle plugins for ktlint instead? Such as:

Not a fan of introducing 3rd party plugins for a plugin, they often are left unmaintained and deprecated.

I'll move the preBuild out.

ktlint.gradle
nighthawk24: Fixed, removed the repository declaration here.

gmale:
Latest changes look good. I'm shipping an internal build tomorrow so I'm going to hold this until all those moving parts are laying flat and then merge this just so that I'm sure I can get the build done without odd conflicts. I don't anticipate any but I just want to play it safe, given that this touches 69 files.

PR #210, opened 2021-01-22
Added responsible disclosure document describing bilateral vuln agree…

zebambam:
…ments.

### DIFF
--- a/responsible_disclosure.md
+++ b/responsible_disclosure.md
@@ -1,0 +1,64 @@
+This page is copyright The Electric Coin Company, 2021. It is posted in order to conform to this standard: https://github.com/RD-Crypto-Spec/Responsible-Disclosure/tree/d47a5a3dafa5942c8849a93441745fdd186731e6
+
+# Security Disclosures
+## Receiving Disclosures
+
+The Electric Coin Company is committed to working with researchers who submit security vulnerability notifications to us to resolve those issues on an appropriate timeline and perform a coordinated release, giving credit to the reporter if they would like.
+
+Please submit issues to security@z.cash, using the following PGP key:
+
+```
+-----BEGIN PGP PUBLIC KEY BLOCK-----
+
+mQENBFaegcoBCAC+G82ZBYYm1GoVn4dKa0WiLYD/Q+BuU89PS1X7A4eOOy8g9yS4
+wJKMzB0AxFsH/t85P7pPZwHw3i2gmiJKeIqEGhEBL08D3id2u6ZyCnwDuWs0i6My
+MXWTwK5shvE61ZI/KPbjemoOG6MPF5QdrouNqei2Vk+4RjbRCyyS0A59GQi2dNZX
+BMwTnHnUZ5qi6T0RFelqJ3dE5Nc/UwJPdAcg71c3b3dMOHjaDBMPB6+fTLBeidV6
+5B72nGO3eIYkMUNj+qCQmM/esRkmGmlDH/9WGMBOKCq7Yw3LyEoPOi5cba1m8SN2
+xFlNzkUGrlVrwZMF+1UdjvN7BGDypA3Dr/STABEBAAG0JVpjYXNoIFNlY3VyaXR5
+IFRlYW0gPHNlY3VyaXR5QHouY2FzaD6JATcEEwEIACEFAlaegcoCGwMFCwkIBwIG
+FQgJCgsCBBYCAwECHgECF4AACgkQiPuLhti1poxlRAf/bZ6fhUby5bAbViAO4TzQ
+yfbD0ksGeF8MHicPz7HqOYuXAE9GrKnVAOFptwRo94O+iRC5aXhW8OAP+38IWorv
+gsAuag7Y8k0nlfNdrJRqqJpjyxtiuv+cd2o5dre8E9PVNE5IPv9qEJA4Zag3snmC
+a+O4HAqKeXYddunFq2drLkTRlOwuFkGXJzwi3VSNVYCuuGyezFDuaD45ltmsXgid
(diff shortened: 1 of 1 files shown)


responsible_disclosure.md
```diff
@@ -0,0 +1,64 @@
...
+D0nht5ifwLjQxM4olvYHUwtT0wPhniH69ghFo8LQiMgncjaukDzbgANiuj07QYy/
+DlzhQUsp1qZvqZnVKqUJ3lFb86b6zoqoRNiUnvP9JB2v3kLG0T39UlcXUFnZJ/4H
+CDHCrwSovQRMHtoOWZijBNobO2y1d0FkUpzNlNw44Ssw0Vo=
+=6GYS
+-----END PGP PUBLIC KEY BLOCK-----
+```
+
+## Sending Disclosures
+
+In the case where we become aware of security issues affecting other projects that has never affected Zcash, our intention is to inform those projects of security issues on a best effort basis.
```
daira, line 44:
```suggestion
In the case where we become aware of security issues affecting other projects that have never affected Zcash, our intention is to inform those projects of security issues on a best effort basis.
```

responsible_disclosure.md
```diff
@@ -0,0 +1,64 @@
...
+We have set up agreements with the following neighboring projects to share vulnerability information, subject to the deviations described in the next section.
+
+Specifically, we have agreed to engage in responsible disclosures for security issues affecting this repository and others we maintain for the same platform with the following contacts:
+
+- Unstoppable wallet via PGP email to security.hs@protonmail.com
+- Nighthawk wallet via their disclosure policy @ https://github.com/nighthawk-apps/nighthawk-wallet-android/blob/master/README.md 
+
+## Deviations from the Standard
+
+Zcash is a technology that provides strong privacy. Notes are encrypted to their destination, and then the monetary base is kept via zero-knowledge proofs intended to only be creatable by the real holder of Zcash. If this fails, and a counterfeiting bug results, that counterfeiting bug might be exploited without any way for blockchain analyzers to identify the perpetrator or which data in the blockchain has been used to exploit the bug. Rollbacks before that point, such as have been executed in some other projects in such cases, are therefore impossible.
```
daira, line 59:
```suggestion
Zcash is a technology that provides strong privacy. Notes are encrypted to their destination, and then the monetary base is kept via zero-knowledge proofs intended to only be creatable by the real holder of Zcash. If this fails, and a counterfeiting bug results, that counterfeiting bug might be exploited without any way for blockchain analyzers to identify the perpetrator or which data in the blockchain has been used to exploit the bug. Rollbacks to before that point, such as have been executed in some other projects in such cases, are therefore impossible.
```

responsible_disclosure.md
```diff
@@ -0,0 +1,64 @@
...
+
+The standard describes reporters of vulnerabilities including full details of an issue, in order to reproduce it. This is necessary for instance in the case of an external researcher both demonstrating and proving that there really is a security issue, and that security issue really has the impact that they say it has - allowing the development team to accurately prioritize and resolve the issue.
```
daira, line 61:
```suggestion
The standard describes reporters of vulnerabilities including full details of an issue, in order to reproduce it. This is necessary for instance in the case of an external researcher both demonstrating and proving that there really is a security issue, and that the security issue really has the impact that they say it has - allowing the development team to accurately prioritize and resolve the issue.
```

daira (approved):
ACK with non-blocking grammar suggestions.

responsible_disclosure.md
gmale: Also, probably remove the comment after "destination" in:
> their destination, and then

responsible_disclosure.md
```diff
@@ -0,0 +1,64 @@
+This page is copyright The Electric Coin Company, 2021. It is posted in order to conform to this standard: https://github.com/RD-Crypto-Spec/Responsible-Disclosure/tree/d47a5a3dafa5942c8849a93441745fdd186731e6
```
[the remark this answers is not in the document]
zebambam: It is deliberate that we commit to a particular er.. commit. :) Otherwise it might change to something that we didn't agree to.

responsible_disclosure.md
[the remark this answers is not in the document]
zebambam: Ah shoot I'm no good at using this interface for multiple suggestions :) looks like my first commit invalidated this one, sorry about that.

responsible_disclosure.md
gmale: > The standard describes reporters of vulnerabilities including full details of an issue, in order to reproduce it. 

It's probably just me but, as worded, this sentence is unclear. First, what standard, exactly? To avoid ambiguity, should that be linked? Second, I had to read the "describes reporters of vulnerabilities including" part a few times before I understood what I think it is trying to say. Perhaps, moving things around would help?

> In order to reproduce vulnerabilities, [the standard](https://github.com/RD-Crypto-Spec/Responsible-Disclosure/tree/d47a5a3dafa5942c8849a93441745fdd186731e6) compels reporters to include full details.

gmale: > This is necessary for instance in the case of an external researcher both demonstrating and proving that there really is a security issue, and that the security issue really has the impact that they say it has - allowing the development team to accurately prioritize and resolve the issue.

Similarly, this one sentence was fairly difficult for me to parse. It seems that it might flow better as multiple sentences with fewer interrupting phrases. Perhaps something more simple, along the lines of:

> These details allow the development team to accurately prioritize and resolve issues. For example, an external researcher can provide sufficient detail about a security issue to both demonstrate that it exists and prove its impact.

gmale (approved):
A few more suggestions that do not block a merge.

zebambam:
We can fix this centrally and I'll push out changes to the repos once we agree on wording.

gmale:
@zebambam just following up on this. Have there been any changes to the wording in the central document, that need to be updated here?

PR #211, opened 2021-01-23
Unit test for param files

herou:
I wrote unit tests required from this ticket:  #202

### DIFF
--- a/src/androidTest/java/cash/z/ecc/android/sdk/tool/SaplingParamToolTest.kt
+++ b/src/androidTest/java/cash/z/ecc/android/sdk/tool/SaplingParamToolTest.kt
@@ -28,7 +28,7 @@ class SaplingParamToolTest {
 
 
     @Test
-    suspend fun testFilesExists(){
+    fun testFilesExists() = runBlocking {
         // Given
         SaplingParamTool.fetchParams(cacheDir)
 
@@ -40,7 +40,7 @@ class SaplingParamToolTest {
     }
 
     @Test
-    suspend fun testOnlySpendFileExits(){
+    fun testOnlySpendFileExits() = runBlocking {
         // Given
         SaplingParamTool.fetchParams(cacheDir)
         File("$cacheDir/${ZcashSdk.OUTPUT_PARAM_FILE_NAME}").delete()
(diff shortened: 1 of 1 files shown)


gmale:
This looks like a good start :tada: 

From here, you probably want a setup function that runs before each unit test, that clears the param directory. Something along the lines of:

```kotlin
    @Before
    fun setup() {
        // clear the param files
        SaplingParamTool.clear(cacheDir)
    }
```
Now, each unit test is starting with an empty cache and is independent of other tests. Next, you would want to setup each test scenario:

```kotlin
    @Test
    fun testFilesExist(){
        // Given
        SaplingParamTool.fetchParams(cacheDir)
        
        // When
        val result = SaplingParamTool.validate(cacheDir)

        // Then
        Assert.assertFalse(result)
    }
 ```
 
If the `fetchParams` function works, then the previous test will succeed. So we can consider it tested. Meaning, we can build on it, as if it works, for the rest of the tests, without having to mock too much:

```kotlin
    @Test
    fun testOnlyOutputOFileExists(){
        // Given
        SaplingParamTool.fetchParams(cacheDir)
        File("$cacheDir/${ZcashSdk.SPEND_PARAM_FILE_NAME}").delete()
                
        // When
        val result = SaplingParamTool.validate(cacheDir)

        // Then
        Assert.assertFalse("Validation should fail when the spend params are missing", result)
    }
 ```

File: src/androidTest/java/cash/z/ecc/android/sdk/tool/SaplingParamToolTest.kt [hunk 1]
```diff
@@ -1,11 +1,56 @@
...
+
+    @Test
+    fun testFilesExist(){
+        val result = SaplingParamTool.validate(cacheDir)
+
+        Assert.assertFalse(result)
+    }
+
+    @Test
+    fun testOnlySpendFileExit(){
```
gmale, line 29: minor typo `Exit` should be `Exists`

src/androidTest/java/cash/z/ecc/android/sdk/tool/SaplingParamToolTest.kt
```diff
@@ -1,11 +1,56 @@
...
+        val result = SaplingParamTool.validate(cacheDir)
+
+        Assert.assertFalse(result)
+    }
+
+    @Test
+    fun testOnlyOutputOFileExit(){
```
gmale, line 36: minor typo `Exit` should be `Exists`

gmale (changes requested):

herou:
@gmale I applied your changes above.

gmale:
Discussed over telegram the `suspend` function issues and all tests pass. Merging. 

Nice work.

From here, the next steps are on my end.

PR #218, opened 2021-04-01
Feature/autoshielding poc

gmale:


src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
```diff
@@ -399,9 +398,13 @@ class CompactBlockProcessor(
         }
     }
 
+    internal suspend fun refreshUtxos(tAddress: String, startHeight: Int): Int = withContext(IO) {
```
pacu, line 401: This API doesn't work well with multiple account support.

src/main/java/cash/z/ecc/android/sdk/transaction/TransactionRepository.kt
```diff
@@ -76,6 +77,8 @@ interface TransactionRepository {
 
     suspend fun count(): Int
 
+    suspend fun getAccount(accountId: Int): UnifiedAddressAccount?
```
pacu, line 80: This feels out of place. I'd add the AccountRepository to the CompactBlockProcessor or SDKSynchronizer directly rather than adding this function

gmale: Good callout this certainly needs to be refactored. I've spun that out into a separate github issue: https://github.com/zcash/zcash-android-wallet-sdk/issues/220

src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
gmale: Very good point. Multiple account support is broken in a few places so I've spun this out into a separate github issue so that we don't lose the feedback:
https://github.com/zcash/zcash-android-wallet-sdk/issues/221

PR #234, opened 2021-08-10
Fix: Corrected logic when calculating birthdates

gmale:
When wallets have no received notes, finding the last transaction will fail so we should search in a way that returns null rather than an exception. 

_Note that the `receivedTransactions` flow is never null so the call to `first()` is fine but the call to `last()` needed to change._


gmale:
Merging: minor change.

Issue #261, opened 2021-08-31
Add a property for NDK version


ccjernigan:
**What is your feature request?**
The NDK version should be refactored somewhere so that it can be shared between the sdk and the demo application.

This is somewhat different from #256 because the NDK isn't exactly like a Gradle dependency.

When implementing this change, also add a comment above the NDK version to reminder authors that the version has to be manually updated on Bitrise.  (Ideally, this would mean moving the Bitrise YML config into the repo instead of hosted on Bitrise).

**How would this feature help you?**
This will ensure the correct NDK version is used for SDK and demo app builds, as well as on the CI server.

Issue #258, opened 2021-08-30
Replace `kotlin-android-extensions`


ccjernigan:
**What is your feature request?**
The demo application is using `kotlin-android-extensions` which is deprecated.  This should be replaced by a combination of `kotlin-parcelize` and `buildFeatures { viewBinding true }` depending on which functionality is actually being used by the demo application.

PR #263, opened 2021-08-31
CI Integrations

### LINKED ISSUE #255: Configure SDK as Its Own Gradle Module
**What is your feature request?**
Consider moving the SDK to its own Gradle module, rather than being combined with the top level Gradle module.

**How would this feature help you?**
This will make it easier to split the SDK into multiple additional modules in the future.  Possible benefits of this could include:
 - Enables bitrise to see the SDK in the repo, as bitrise requires a module name (it cannot be blank)
 - Splitting the SDK internally into layers or components, which could make ongoing SDK development easier.
 - Hiding internal implementation details from consumers of the SDK. Currently non-public APIs are accessible to SDK clients.  If those were moved into a separate module, such as `sdkInternals` and the ultimate sdk module depends on it as an `implementation` instead of an `api` dependency, those internals don't get exported into the public API
 - Managing SDK size, by allowing consumers to only choose SDK functionality they need similar to how Android X is divided into many different components
 - Extension modules, similar to how Android X offers ktx extensions for various Android X features
 - Division of project wide Gradle configuration under the root build.gradle versus module-specific Gradle configuration makes the build scripts overall simpler

pacu:
Fixes #255 #258 #261


pacu:
approved by @ccjernigan over private convo

Issue #270, opened 2021-09-10
SDK Fails to fallback to checkpoint on testnet build


pacu:
**What is the bug?**
![image](https://user-images.githubusercontent.com/484008/132909667-846c9ee5-ee79-4aab-abae-86383d976900.png)
when using 1.3.0-beta.17, the testnet app fails to fall back to the latest local checkpoint.
Although beta18 fixes the issue because the checkpoint actually exists in the SDK's assets, this clearly indicates some flaw in the logic of the checkpoint selection 

**Solution**
Reproduce the issue, find out whether there are changes needed  or if this is just a data packaging error.

nighthawk24:
Fixed in https://github.com/zcash/zcash-android-wallet-sdk/pull/250

Although it would help in creating another checkpoint for both `main-net` & `test-net` variants.

PR #272, opened 2021-09-11
#270 Fixes for bad JSON files

ccjernigan:
This PR is primarily a set of fixes for #270 by making the following changes:
 - Add unit tests for validating JSON assets, reducing likelihood of bad assets being added to the SDK
 - Fixed several issues with the existing testnet JSON files which were picked up by the new tests
 - New fallback behavior, allowing a deployed SDK to continue if a bad JSON file is introduced.  This behavior change is behind a boolean flag currently, which can be removed once we confirm this is the best behavior
 - Fixed a resource leak in the loading of JSON assets (the input streams were not being closed)

In testing this PR, I was also trying to publish to maven local to pull into the app and ran into a few issues with publishing locally.  So I've also added those fixes as part of this work.


pacu (approved):

sdk-lib/src/main/java/cash/z/ecc/android/sdk/tool/WalletBirthdayTool.kt
```diff
@@ -76,44 +97,75 @@ class WalletBirthdayTool(appContext: Context) {
...
+                    }
+                } catch (t: Throwable) {
+                    val exception = BirthdayException.MalformattedBirthdayFilesException(
+                        directory,
+                        treefile
+                    )
+                    lastException = exception
+
+                    if (IS_FALLBACK_ON_FAILURE) {
+                        // TODO: If we ever add crash analytics hooks, this would be something to report
```
pacu, line 160: The SDK should report exceptions and never purposely crash, clients will handle them as they see fit. Some would choose to use crash reporting, others won't.

ccjernigan: Agreed.  Currently we don't have a mechanism for injecting crash analytics from the consumer of the SDK, so that's probably a new feature to consider for the SDK itself.

Issue #260, opened 2021-08-31
Remove Dependencies on jcenter


ccjernigan:
**What is your feature request?**
Remove dependency on jcenter, since jcenter is deprecated.  The demo-app appears to depend on jcenter for one of our own artifacts.

This is non-urgent given that jcenter will continue hosting artifacts but eventually this migration will need to happen as new artifact versions can't be published to jcenter in the future.

pacu:
I think this can already be worked on. 
Are there any changes needed on the reference wallet to support this or is maven central already adopted ?

ccjernigan:
The packages that appear to be in jcenter are:
 - cash.z.ecc.android:kotlin-bip39:1.0.1
 - cash.z.ecc.android:zcash-android-sdk-testnet:1.2.1-beta02

ccjernigan:
Do we have updated versions of those deployed to maven central or jitpack?

pacu:
> cash.z.ecc.android:kotlin-bip39:1.0.1 
This one is the BIP 39 library we developed. It's probably not on maven central. 

> cash.z.ecc.android:zcash-android-sdk-testnet:1.2.1-beta02 
I don't think this should be ever needed the sdk should support both mainnet and testnet under the same artifact

ccjernigan:
I found there's a newer version of bip39 on maven central, so that'll be easy.

Let me see what happens on removal of the testate artifact.

ccjernigan:
The demo application is using a very old version of the SDK, and there are a number of public API changes in the SDK which for which the demo application needs to be updated.  I push some commits to a branch [here](https://github.com/ccjernigan/zcash-android-wallet-sdk/tree/260-jcenter), although the build is broken with be following issues that need to be resolved.

I haven't attempted fixing any of these yet, so I'm just providing an update on where this is currently.  I'll try to do a first pass at resolving these compile issues, although I might need some guidance if the new public APIs aren't immediately clear.

GetAddressFragment.kt
Unresolved reference: deriveViewingKeys
None of the following functions can be called with the arguments supplied:
public open fun deriveShieldedAddress(seed: ByteArray, network: ZcashNetwork, accountIndex: Int = ...): String defined in cash.z.ecc.android.sdk.tool.DerivationTool.Companion
public open fun deriveShieldedAddress(viewingKey: String, network: ZcashNetwork): String defined in cash.z.ecc.android.sdk.tool.DerivationTool.Companion
No value passed for parameter 'network'
[...]

PR #275, opened 2021-09-12
Fix demo app crashes; update dependencies

### LINKED ISSUE #265: Demo App Crashes on newer versions of Android
STEPS TO REPRODUCE
1. Build and run the demo app on an Android 12 device
2. Choose List Transactions from the hamburger menu

RESULTS
Actual:
```
2021-09-01 15:03:28.015 9588-9952/cash.z.ecc.android.sdk.demoapp.mainnet E/AndroidRuntime: FATAL EXCEPTION: grpc-okhttp-0
    Process: cash.z.ecc.android.sdk.demoapp.mainnet, PID: 9588
    java.lang.AssertionError: Method getAlpnSelectedProtocol not supported for object SSL socket over Socket[address=mainnet.lightwalletd.com/140.238.134.157,port=9067,localPort=48698]
        at io.grpc.okhttp.internal.OptionalMethod.invoke(OptionalMethod.java:114)
        at io.grpc.okhttp.internal.OptionalMethod.invokeWithoutCheckedException(OptionalMethod.java:135)
        at io.grpc.okhttp.OkHttpProtocolNegotiator$AndroidNegotiator.getSelectedProtocol(OkHttpProtocolNegotiator.java:183)
        at io.grpc.okhttp.OkHttpProtocolNegotiator$AndroidNegotiator.negotiate(OkHttpProtocolNegotiator.java:145)
        at io.grpc.okhttp.OkHttpTlsUpgrader.upgrade(OkHttpTlsUpgrader.java:63)
        at io.grpc.okhttp.OkHttpClientTransport$4.run(OkHttpClientTransport.java:568)
        at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:923)
```
Expected:
Demo application shouldn't be crashing.

NOTES
Appears to be related to Android version, as the issue does not reproduce on an Android 7 device that I tested.  There seems to be a related issue here: 
https://github.com/grpc/grpc-java/issues/7519

Most likely, the issue is because the demo-app isn't currently relying on the SDK in the repo but instead uses the version downloaded from Maven and the version it depends on is currently very old (thus using an old version of GRPC).

I think the ultimate fix is to default the demo app to use the locally compiled version of the SDK (with an optional gradle property to switch to mavenLocal for testing).  We already have the [other project](https://github.com/zcash/zcash-android-wallet) which should be demonstrating integration with Maven Central.

ccjernigan:
This PR fixes #265 and #260 by updating dependencies and having the demo-app depend directly on the sdk as a gradle module, rather than depending on the published version.  All of the changes should just be to the demo app, not to the SDK itself.


demo-app/src/zcashtestnet/res/values/strings.xml
```diff
@@ -1,4 +1,5 @@
 <resources>
     <string name="app_name">Testnet Demo</string>
     <string name="nav_header_title">Android SDK Demo : TESTNET</string>
+    <string name="network_name">Testnet</string>
```
pacu, line 4: Out of curiosity, how does the user select mainnet or testnet? is it a build flavor / target or a dynamic setting?

ccjernigan: The build flavor selects it automatically, as there's a resource file under both `demo-app/src/zcashtestnet/res/values/strings.xml` and `demo-app/src/zcashmainnet/res/values/strings.xml`

pacu (approved):

PR #299, opened 2021-10-04
[#293] Refactor non-public APIs to internal package

ccjernigan:
This implements phase 1 of the refactoring of the SDK to make internal APIs better isolated from the public API.

This is primarily moving files around.  The few "new" files are the following:
 - Sma was inside another file and was moved to the top level
 - Metrics was renamed to BatchMetrics to match class name with file name

To test these changes, I've deployed a mavenLocal version of the SDK and verified it does not break compile compatibility with the existing wallet app.  These tests are documented in this commit under docs/tests/


sdk-lib/src/main/java/cash/z/ecc/android/sdk/ext/BlockExt.kt
```diff
@@ -7,14 +7,8 @@ fun ByteArray.toHex(): String {
     return sb.toString()
 }
 
-fun ByteArray.toHexReversed(): String {
```
pacu, line 10: Why is this being removed? wasn't it being used?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/BlockExt.kt
```diff
@@ -0,0 +1,9 @@
+package cash.z.ecc.android.sdk.internal.ext
```
pacu, line 1: disregard last comment

PR #300, opened 2021-10-04
Changes to make the saplingActivation configurable

pacu:
This makes the network parameter extend an interface and be configurable for testing

### DIFF
--- a/darkside-test-lib/src/androidTest/java/cash/z/ecc/android/sdk/darkside/test/DarksideTestCoordinator.kt
+++ b/darkside-test-lib/src/androidTest/java/cash/z/ecc/android/sdk/darkside/test/DarksideTestCoordinator.kt
@@ -4,6 +4,7 @@ import cash.z.ecc.android.sdk.SdkSynchronizer
 import cash.z.ecc.android.sdk.Synchronizer
 import cash.z.ecc.android.sdk.darkside.fixture.FixtureTransaction
 import cash.z.ecc.android.sdk.ext.twig
+import cash.z.ecc.android.sdk.type.NetworkType
 import cash.z.ecc.android.sdk.type.ZcashNetwork
 import io.grpc.StatusRuntimeException
 import kotlinx.coroutines.ExperimentalCoroutinesApi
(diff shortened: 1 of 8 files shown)


sdk-lib/src/main/java/cash/z/ecc/android/sdk/type/WalletTypes.kt
```diff
@@ -73,7 +73,24 @@ interface UnifiedAddress {
...
+    val networkName: String
+    val saplingActivationHeight: Int
+    val defaultHost: String
+    val defaultPort: Int
+
+    // Required to enable extension functions
+    companion object
+}
+
+enum class NetworkType(
```
ccjernigan, line 87: Note this is a public API change, so we'll need to update the changelog and migration readme for this.  It shouldn't have a huge impact, but it will probably break compile compatibility in our own wallet app.

This seems like a worthwhile API change, I just wanted to call it out explicitly.

darkside-test-lib/src/androidTest/java/cash/z/ecc/android/sdk/darkside/test/DarksideTestCoordinator.kt
```diff
@@ -301,27 +297,44 @@ class DarksideTestCoordinator(val wallet: TestWallet = newDarksideTestWallet())
...
         fun newDarksideTestWallet(
-            alias: String = "DarksideTestCoordinator",
-            seedPhrase: String = DEFAULT_SEED_PHRASE,
-            startHeight: Int = DEFAULT_START_HEIGHT,
-            host: String = COMPUTER_LOCALHOST,
-            network: ZcashNetwork = ZcashNetwork.Mainnet,
-            port: Int = network.defaultPort,
+                alias: String = "DarksideTestCoordinator",
+                seedPhrase: String = DEFAULT_SEED_PHRASE,
+                startHeight: Int = DarksideNetwork.DEFAULT_START_HEIGHT,
```
ccjernigan: In the future, will we expect test cases to pass in a start height that's different from the default?  I'm basically asking whether we need this method parameter or not.

ccjernigan (commented):
This is a nice refactor and gives some good additional flexibility for testing!  I pushed a few commits to this branch, fixing some compile failures from the refactoring of Network type.

I also fixed the Android emulator detection, so that should work now.

There are a few open questions from me inline.

darkside-test-lib/src/androidTest/java/cash/z/ecc/android/sdk/darkside/test/DarksideTestCoordinator.kt
pacu: We actually should, because the correct usage of this tool involves being able to use a different set of activation heights. We haven't due to priorities/time constraints but the support for this parameter seems adecquate.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/type/WalletTypes.kt
pacu: Thanks for bringing this up. I'll put together that documentation.

PR #315, opened 2021-11-17
[#313] Gradle 7.3

ccjernigan:
In addition to updating the version of Gradle, I've added the sha256 to enable verification of the Gradle download


ccjernigan:
We should wait until PR #320 is merged and we can rebase, to verify that this change doesn't break the CI build.

pacu (approved):

PR #320, opened 2021-11-18
Fix ktlint formatting issues

ccjernigan:
This is the result of updating the ktlint version and running the gradle task ktlintFormat


pacu (approved):
good job mr linter

PR #329, opened 2021-11-22
Did `cargo update` to fix compilation error

pacu:
Same as iOS SDK. 

In the process of publishing the SDK I got this cargo build error

````
use of unstable library feature 'int_roundings'
   --> /Users/pacu/.cargo/registry/src/github.com-1ecc6299db9ec823/num-bigint-0.3.2/src/biguint.rs:393:45
    |
393 |                 let root_scale = extra_bits.div_ceil(&n64);
````

which was fixed by running `cargo update` in side of the `sdk-lib` 

Somehow this error is not reproducible in every build. @defuse was able to build the beta19 branch without problems. could you guys please review this?


ccjernigan (approved):
I am not able to reproduce the original build failure.  I did try building this branch locally, and did not run into any issues after the dependency update.

This change looks reasonable to me, although I did notice three dependencies went backwards in their version number with this change.  Those dependencies are: `memchr`, `regex`, and `aho-corasick`

pacu:
There's code smell here because this build is not either failing for you  @ccjernigan and did not fail on @defuse's side either.

ccjernigan:
> There's code smell here because this build is not either failing for you @ccjernigan and did not fail on @defuse's side either.

Agreed, because it doesn't fail on the Continuous Integration server either.

pacu:
I isolated the error to `num-bigint` 0.3.2 and bumped to 0.3.3

pacu:
I figured it out. it's the nightly toolchain I had set up because of the M1 compatibility on the iOS SDK that broke the bigint thing.

PR #334, opened 2021-11-29
Improve compatibility on Apple Silicon

ccjernigan:
This PR updates the Room dependency, to resolve a build failure on Apple Silicon.

I've also made a fix to ktlint's task, forcing it to use Java 11.  Some related documentation is updated, because getting the correct JDK on Apple Silicon can be a bit difficult due to compatibility issues between Java and Kotlin.


gradle.properties
```diff
@@ -79,7 +79,7 @@ ANDROIDX_LIFECYCLE_VERSION=2.4.0
 ANDROIDX_MULTIDEX_VERSION=2.0.1
 ANDROIDX_NAVIGATION_VERSION=2.3.5
 ANDROIDX_PAGING_VERSION=2.1.2
-ANDROIDX_ROOM_VERSION=2.3.0
+ANDROIDX_ROOM_VERSION=2.4.0-beta02
```
pacu, line 82: does this update have to do with the m1 stuff?

pacu (approved):

gradle.properties
ccjernigan: Yes, it has a compile-time dependency that breaks during builds.  The beta includes this fix https://issuetracker.google.com/issues/174695268

PR #360, opened 2022-01-11
GitHub Actions

ccjernigan:
The primary purpose of this is to set the stage for automated testing and deployment of snapshot releases to maven central, along with manual deployment of releases.  Most of the documentation on this process is in the commits.  The deployment is automatically disabled until the necessary credentials are added.

Addition of the detekt baseline allows detekt static analysis to also be run on each build.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,87 @@
+name: 'Setup Java, Rust, and Dependency Cache'
+description: "Configures the build environment and caches Gradle, dependencies, and build outputs."
+runs:
+  using: "composite"
+  steps:
+    - name: Set Env
+      shell: bash
+      run: |
+        echo "home=${HOME}" >> "$GITHUB_ENV"
+    - name: Set up Java
+      uses: actions/setup-java@v2
+      with:
+        distribution: 'zulu'
+        java-version: 17
+    - name: Setup Rust
+      shell: bash
+      run: |
+        rustup target add armv7-linux-androideabi aarch64-linux-android i686-linux-android x86_64-linux-android
+    - name: Disable Gradle Daemon
+      shell: bash
+      run: |
+        mkdir ~/.gradle
+
+        echo "org.gradle.daemon=false" >> ~/.gradle/gradle.properties
+    - name: Gradle Wrapper Cache
+      id: gradle-wrapper-cache
+      uses: actions/cache@v2.1.7
+      with:
+        path: ~/.gradle/wrapper
+        key: ${{ runner.os }}-gradle-wrapper-${{ hashFiles(format('{0}{1}', github.workspace, '/gradle/wrapper/gradle-wrapper.properties')) }}
+    - name: Gradle Dependency Cache
+      id: gradle-dependency-cache
+      uses: actions/cache@v2.1.7
+      with:
+        path: ~/.gradle/caches/modules-2
+        key: ${{ runner.os }}-gradle-deps-${{ hashFiles(format('{0}{1}', github.workspace, '/gradle.properties')) }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-deps
+    # This tries to fall back to the build cache from the main branch, while ensuring that
+    # main branch builds repopulate the cache each time.
+    - name: Gradle Build Cache Main
+      id: gradle-build-cache-main
+      if: github.event.pull_request.head.sha == ''
+      uses: actions/cache@v2.1.7
+      with:
+        path: |
+          ~/.gradle/caches/build-cache-1
+          ~/.gradle/caches/transforms-3
+        key: ${{ runner.os }}-gradle-build-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-build
+    - name: Gradle Build Cache Pull Request
+      id: gradle-build-cache-pr
+      if: github.event.pull_request.head.sha != ''
+      uses: actions/cache@v2.1.7
+      with:
+        path: |
+          ~/.gradle/caches/build-cache-1
+          ~/.gradle/caches/transforms-3
+        key: ${{ runner.os }}-gradle-build-${{ github.event.pull_request.base.sha }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-build
+    - name: Rust Cache
+      id: rust-cache
+      uses: actions/cache@v2.1.7
+      with:
+        path: |
+          sdk-lib/target
+          ~/.cargo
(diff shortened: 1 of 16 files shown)


pacu:
adding @defuse and @mdr0id  from DevSecOps for review

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,85 @@
...
+      id: gradle-build-cache-main
+      if: github.event.pull_request.head.sha == ''
+      uses: actions/cache@v2.1.7
+      with:
+        path: |
+          ~/.gradle/caches/build-cache-1
+          ~/.gradle/caches/transforms-3
+        key: ${{ runner.os }}-gradle-build-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-build
```
defuse, line 51: We need to be careful that builds on the main branch don't use caches produced by potentially-malicious PRs, otherwise a malicious PR could insert malware into the cache that comes alive when a main branch build happens and steals the secrets.

I think [cache access restrictions](https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) stop this from happening:

> Access restrictions provide cache isolation and security by creating a logical boundary between different branches. For example, a cache created for the branch feature-a (with the base main) would not be accessible to a pull request for the branch feature-b (with the base main).

This guarantees that PRs can't interfere with each other. It doesn't exactly say that main branch builds can't access PR branch caches but I think that's what it means.

Does that sound right? (It might be a good idea to mention this in the documentation.)

defuse (approved):
utACK modulo my question about caching.

.github/actions/setup/action.yml
ccjernigan: There's a hierarchy for cache restores.  My understanding is that branches can restore from the main branch but the main branch cannot restore from other branches, however the docs could probably more more explicit about this.  This what I read:  https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

mdr0id:
utACK; couple minor questions

1. What branch protection rules will be used with the above CI workflow and which status checks will be required? 

2. Will the test_robo_demo_app use different machines than the shared runners or will they all use shared runners for near term wallet needs?

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,300 @@
...
+      - name: Collect Artifacts
+        timeout-minutes: 1
+        env:
+          ARTIFACTS_DIR_PATH: ${{ format('{0}/artifacts', env.home) }}
+          BINARIES_ZIP_PATH: ${{ format('{0}/artifacts/binaries.zip', env.home) }}
+          MAPPINGS_ZIP_PATH: ${{ format('{0}/artifacts/mappings.zip', env.home) }}
+        run: |
+          mkdir ${ARTIFACTS_DIR_PATH}
+          zip -r ${BINARIES_ZIP_PATH} . -i *app/build/outputs/apk/*/release/*.apk *app/build/outputs/bundle/*/release/*.aab
+          zip -r ${MAPPINGS_ZIP_PATH} . -i *app/build/outputs/mapping/*/mapping.txt
```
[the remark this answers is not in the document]
ccjernigan: There are two things we could do to enhance this further:
 - Add a readme to the zip file so that someone downloading it would have a reminder visible to them
 - Our pull request checklist also covers this, although the checklist hasn't been added to this repo yet.  (Example is in the [Secant repo](https://github.com/zcash/secant-android-wallet/blob/main/.github/pull_request_template.md))

ccjernigan:
> utACK; couple minor questions
> 
> 1. What branch protection rules will be used with the above CI workflow and which status checks will be required?

Suggestions: 
 - Require at least 1 review approval
 - Require the static analysis steps to succeed
 - Require the robo test to succeed (once we add the API key)
 - Require the test_android_modules to succeed (once we add the API key and once we fix the tests, as some of the tests are sort of broken right now)

Enhancements:
 - We can add a codeowners configuration, requiring additional approval of changes to the CI scripts

> 2. Will the test_robo_demo_app use different machines than the shared runners or will they all use shared runners for near term wallet needs?

There are two systems involved:
 - GitHub Actions runners for running the Gradle tasks; these should be pretty safe.  Each action step uses an ephemeral runner.  The cache is set up before this step, so this step shouldn't add anything back into the cache.  So it should be mostly independent from the other build steps
 - Firebase Test Lab provisions a temporary virtual device for us to use.  Its run in Google's cloud and should be very isolated.  We're relying on Google to manage setup and teardown in a secure way.  There's supposed to be no data carried over from one run to the next.  This should be significantly more secure than running the Android emulator locally.  (Also faster too, since the emulator run inside GitHub Actions won't support nested virtualization and therefore will be slow emulation of ARM instructions)

ccjernigan:
Also some settings in GitHub Actions should probably be updated from their defaults:
 - Enable Allow select actions: Actions created by Github and Marketplace Actions by verified creators (allows the Google auth action)
 - Disable "run workflows from fork pull requests"
 - Change workflow permissions: Read repository contents permission.

It looks something like this

image

pacu (approved):

mdr0id:
> > utACK; couple minor questions
> > 
> > 1. What branch protection rules will be used with the above CI workflow and which status checks will be required?
> 
> Suggestions:
> 
>     * Require at least 1 review approval
> 
>     * Require the static analysis steps to succeed
> 
>     * Require the robo test to succeed (once we add the API key)
> 
>     * Require the test_android_modules to succeed (once we add the API key and once we fix the tests, as some of the tests are sort of broken right now)
> 
> 
> Enhancements:
> 
>     * We can add a codeowners configuration, requiring additional approval of changes to the CI scripts
> 
> 
> > 2. Will the test_robo_demo_app use different machines than the shared runners or will they all use shared runners for near term wallet needs?
> 
> There are two systems involved:
> 
>     * GitHub Actions runners for running the Gradle tasks; these should be pretty safe.  Each action step uses an ephemeral runner.  The cache is set up before this step, so this step shouldn't add anything back into the cache.  So it should be mostly independent from the other build steps
> 
>     * Firebase Test Lab provisions a temporary virtual device for us to use.  Its run in Google's cloud and should be very isolated.  We're relying on Google to manage setup and teardown in a secure way.  There's supposed to be no data carried over from one run to the next.  This should be significantly more secure than running the Android emulator locally.  (Also faster too, since the emulator run inside GitHub Actions won't support nested virtualization and therefore will be slow emulation of ARM instructions)

Sounds reasonable. I have updated Github setting but have not changed the required status checks yet. Please let me know when you want to update these of if they block

Issue #368, opened 2022-01-16
Enhance error API


ccjernigan:
**What is your feature request?**
Currently, the SDK emits certain errors but they don't always contain sufficient information for clients to know how to handle them (#344 is an example). Although over time the goal is to eliminate bugs to ensure the SDK is as robust as possible, errors can and will still occur at some frequency.

This will depend on enhancements to librustzcash to better convey errors to the SDK, so that the SDK can convey that to apps consuming the SDK.

I can think of three possible approaches:
 - Convey low level error codes (`SQLITE_FULL`, `SQLITE_CORRUPT`, etc).  Pro: Consistent representation of errors throughout.  Con: Exposes implementation details (e.g. that the SDK uses SQLite).  Future implementation changes would therefore require API changes.
 - Convey high level error codes (`DISK_FULL`, `DATA_CORRUPT`).  Pro: Less tightly coupled to implementation details.  Con: Can make debugging slightly more difficult.  
 - Convey solution codes (`FREE_UP_DISK_SPACE`, `WIPE_AND_START_OVER`).  Pro: Explicitly conveys what clients should do.  Con: Can make debugging much more difficult.

A hybrid approach is also possible.  For example, high level error codes or solution codes could contain an extra field of debug information (e.g. a stacktrace) which isn't really the API consumers would use but would help with debugging in the field.

Before deciding on the approach, we should try to enumerate the type of errors that we expect as well as the type of repair options.  That will help guide the design of this API.

**How would this feature help you?**
Clients of the SDK can present users with options to repair the issue, e.g. as described in https://github.com/zcash/secant-android-wallet/issues/175

Issue #344, opened 2022-01-03
ECC wallet fails to scan blocks: "database disk image is malformed"


pacu:
![image.png](https://images.zenhubusercontent.com/5dc33e54c84ceb0001835974/83132892-f698-4cad-894b-ffaa8b6413e6)

**cause:**
 java.lang.RuntimeException: Rust error while scanning block batch: database disk image is malformed

**description**
Error: non-fatal reported error due to: cash.z.ecc.android.sdk.exception.CompactBlockProcessorException$FailedScan: Error while scanning blocks. This most likely means a block was missed or a reorg was mishandled. See logs for details.

**More bugsnag details**
````
exceptionString
cash.z.ecc.android.sdk.exception.CompactBlockProcessorException$FailedScan: Error while scanning blocks. This most likely means a block was missed or a reorg was mishandled. See logs for details.

isError
true

isFatal
false

key
error.nonfatal.reported

message
Error while scanning blocks. This most likely means a block was missed or a reorg was mishandled. See logs for details.

stacktrace.0
cash.z.ecc.android.sdk.exception.CompactBlockProcessorException$FailedScan: Error while scanning blocks. This most likely means a block was missed or a reorg was mishandled. See logs for details.
	at cash.z.ecc.android.sdk.block.CompactBlockProcessor

stacktrace.1
$scanNewBlocks$2.invokeSuspend(CompactBlockProcessor.kt:539)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.scheduling

stacktrace.2
.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:67

stacktrace.3
8)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
Caused by: java.lang.RuntimeException: Rust error while scanning block batch: database disk image is malformed
	at cash.z.ecc.android.sdk.jni.RustBackend.sc

stacktrace.4
anBlockBatch(Native Method)
[...]

Issue #308, opened 2021-10-21
Remove `runBlocking` calls from demo-app


ccjernigan:
During implementation of #269, a number of calls in the demo-app were wrapped in `runBlocking`.  This doesn't harm anything, and performance is the same as before.  However it doesn't set a good precedent in terms of others looking at the code or copying it.

So this is a followup issue to come back and address those instances.  A quick search for `runBlocking` and `newBlocking` should turn them all up.

ccjernigan:
This is being closed because the old demo app is scheduled to be deleted.

Issue #242, opened 2021-08-20
Incorporate the Paging3 library


gmale:
As a wallet developer, I'd like paging to be easy to use in the SDK so that I can display transactions efficiently in the wallet.

Background:
- Paging is currently broken! Therefore this bug is labeled as high-severity.
    - It relies on a [class that I had been told would be moved into Android](https://github.com/zcash/zcash-android-wallet-sdk/tree/296782286b473c309764e6f26ac0bf6711880302/src/main/java/cash/z/ecc/android/sdk/ext/android) but there's a bug
    - We are [passing in 1000 as the "page size"](https://github.com/zcash/zcash-android-wallet-sdk/blob/4586e07cac5d431d0c6842a62d69a0de5036547c/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt#L789) to mask the fact that paging is broken!!!
    - We should be passing something more reasonable like 10 (the number that can fit on one or two screens of the app)
- I was [contemplating writing a component myself ](https://androidstudygroup.slack.com/archives/C5ELAPJC9/p1572878898076000) until I found out that [Chris Banes created basically the same thing](https://twitter.com/chrisbanes/status/1160858967532163072?s=20) so I switched to his implementation of a paging datasource that was compatible with flows
- At the time the maintainers of the Paging library said they were working on a complete overhaul:
    > we're working on a Coroutine-first overhaul of the Paging library, with the focus and benefits more on moving the loading APIs (DataSource) to coroutines. We're planning to natively support flow.

Criteria:
- leverages the new Paging 3 library
- paging is easy and intuitive
- paging is no longer broken

ccjernigan:
If we implement paging, it will be done as paging primitives on the public API rather than using the Paging library.  We don't want our public API to be tightly coupled with Google.

PR #369, opened 2022-01-17
[#344] Fix threading model to reduce SQLite failures

### LINKED ISSUE #310: Add Empty Companion Objects for Synchronizer, Config, WalletBirthday
**What is your feature request?**
Consider adding empty Companion objects for several of the important public classes or interfaces in the SDK, such as Synchronizer, Config, and WalletBirthday.

**How would this feature help you?**
As an integrator with the SDK, there are times when I'd like to write Companion extension functions to simplify initialization.  By adding empty companion objects, this becomes possible.

If this is implemented, it will be important to add a comment to each empty companion that it shouldn't be deleted, because it would become part of the public API. Sometimes seeing a line that just reads `companion object` can make someone reading the line think it is useless.

### LINKED ISSUE #355: Blocking calls in `PagedTransactionRepository`
**What is the bug?**
During setup of the `PagedTransactionRepository`, disk IO is performed on the main thread.

**Additional context**
```
StrictMode policy violation; ~duration=29 ms: android.os.strictmode.DiskReadViolation
	at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1659)
	at android.database.sqlite.SQLiteConnection.applyBlockGuardPolicy(SQLiteConnection.java:1199)
	at android.database.sqlite.SQLiteConnection.executeForLong(SQLiteConnection.java:749)
	at android.database.sqlite.SQLiteConnection.setAutoCheckpointInterval(SQLiteConnection.java:303)
	at android.database.sqlite.SQLiteConnection.open(SQLiteConnection.java:262)
	at android.database.sqlite.SQLiteConnection.open(SQLiteConnection.java:205)
	at android.database.sqlite.SQLiteConnectionPool.openConnectionLocked(SQLiteConnectionPool.java:505)
	at android.database.sqlite.SQLiteConnectionPool.open(SQLiteConnectionPool.java:206)
	at android.database.sqlite.SQLiteConnectionPool.open(SQLiteConnectionPool.java:198)
	at android.database.sqlite.SQLiteDatabase.openInner(SQLiteDatabase.java:919)
	at android.database.sqlite.SQLiteDatabase.open(SQLiteDatabase.java:899)
	at android.database.sqlite.SQLiteDatabase.openDatabase(SQLiteDatabase.java:763)
	at android.database.sqlite.SQLiteDatabase.openDatabase(SQLiteDatabase.java:752)
	at android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:373)
	at android.database.sqlite.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:316)
	at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper$OpenHelper.getWritableSupportDatabase(FrameworkSQLiteOpenHelper.java:151)
	at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper.getWritableDatabase(FrameworkSQLiteOpenHelper.java:112)
	at cash.z.ecc.android.sdk.internal.transaction.PagedTransactionRepository.buildDatabase(PagedTransactionRepository.kt:180)
	at cash.z.ecc.android.sdk.internal.transaction.PagedTransactionRepository.prepare(PagedTransactionRepository.kt:106)
	at cash.z.ecc.android.sdk.internal.transaction.PagedTransactionRepository$prepare$1.invokeSuspend(Unknown Source:14)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at android.os.Handler.handleCallback(Handler.java:938)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loopOnce(Looper.java:201)
	at android.os.Looper.loop(Looper.java:288)
	at android.app.ActivityThread.main(ActivityThread.java:7839)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
```

**Solution**
The prepare method is already a `suspend` function, so this should not require an API change.  The buildDatabase method appears to be the culprit.

ccjernigan:
While there are a lot of changes, the most significant are:
 - Update the main entrypoint of the SDK to use a suspending companion function and eliminate lazy initialization.  This also resolves #310
 - Changing Dao methods to `suspend`, allowing Room to implicitly manage the threading model under the hood
 - Explicitly set the executor thread for the `CompactBlockDb` and `DerivedDataDb`, which is the same thread used when calling into the Rust native library.  Room will then use this thread for `suspend` fun calls into it
 - With the above set, remove unnecessary hops to Dispatchers.IO which is a thread pool
 - Eliminate lazy initialization in `PagedTransactionRepository` to make it easier to reason about the threading model and resolve #355 where the database was being touched from the UI thread

Note that `PendingTransactionsDb` is not opted into this single-threaded approach, because that database is not shared with the Rust code

A few notes on reviewing:
 - The new suspending interface for Synchronizer initialization required using `newBlocking` in the demo app.  This doesn't degrade performance from where it was before, but also isn't an ideal construct to model for others using the SDK.  Eventually we should come back and fix those instances.  This is effectively captured in the existing issue #308
 - There were a few compile issues with the tests from before (Okio's interface changed).  These were fixed as well
 - I filed a followup issue #368 so that we can improve error handling going forward
 - I filed followup issue #370 so that we can remove large swaths of multithreaded dispatch from the SDK

For testing, I used the demo app and allowed the transactions to sync.  A screenshot of those most recent transactions is also attached

post-changes
.


ccjernigan:
@pacu 

Do any of these Rust backend methods touch the database or have multithreading concerns? `isValidShieldedAddr`, `isValidTransparentAddr`, `getBranchIdForHeight`

I ask because they're defined in RustBackendWelding as non-suspend functions.  They delegate to native calls which don't take in `dbDataPath` so I'm assuming they don't touch anything on disk but still wanted to confirm.

ccjernigan:
There's also now a [branch](https://github.com/zcash/zcash-android-wallet/tree/included-build) in the ECC wallet app that uses an included build to test these changes.  It is downloading synchronizing now, so I'll report back when it completes.

ccjernigan:
The ECC wallet running this version of the SDK also syncs and displays transactions succesfully

ecc wallet

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -247,7 +246,7 @@ class SdkSynchronizer internal constructor(
     override val latestBirthdayHeight: Int get() = processor.birthdayHeight
 
     override suspend fun prepare(): Synchronizer = apply {
-        storage.prepare()
+        // Do nothing; this could likely be removed
```
pacu, line 249: the idea behind having a prepare() method was to make init() lightweight.  How would a suspending initializer impact SDK clients in general?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -736,71 +748,78 @@ class SdkSynchronizer internal constructor(
...
+        // alternatively, this entire object graph can be supplied by Dagger
+        // This builder just makes that easier.
+        return SdkSynchronizer(
+            repository,
+            txManager,
+            processor
+        )
+    }
+
+    // TODO: fix this pagesize bug, small pages should not crash the app. It crashes with: Uncaught Exception: android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views. and is probably related to FlowPagedList
```
pacu, line 781: do we have a ticket for this issue?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/block/CompactBlockDbStore.kt
```diff
@@ -5,58 +5,62 @@ import androidx.room.Room
...
  */
-class CompactBlockDbStore(
-    appContext: Context,
-    val dbPath: String
-) : CompactBlockStore {
-
-    private val cacheDao: CompactBlockDao
+class CompactBlockDbStore private constructor(
+    private val cacheDao: CompactBlockDao,
     private val cacheDb: CompactBlockDb
```
pacu, line 21: why would you need a cacheDao in the constructor if it can be constructed from the cacheDb? is it for testing purposes?

Would it be possible to mistakenly pass a Dao and a CompactBlockDb that don't use the same database? What would happen in that case?

can this be done in an extension that allows the user to inject the  interface?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/block/CompactBlockDbStore.kt
```diff
@@ -5,58 +5,62 @@ import androidx.room.Room
...
+    companion object {
+        /**
+         * @param appContext the application context. This is used for creating the database.
+         * @property dbPath the absolute path to the database.
+         */
+        fun new(appContext: Context, dbPath: String): CompactBlockDbStore {
+            val cacheDb = createCompactBlockCacheDb(appContext.applicationContext, dbPath)
+            val cacheDao = cacheDb.compactBlockDao()
+
+            return CompactBlockDbStore(cacheDao, cacheDb)
```
pacu, line 50: See previous comment above. Could someone create an instance of this object that refers to different databases?

pacu (commented):
I left comments, that may or may not require changes

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
ccjernigan: Yes.  This is a pre-existing todo comment; it just got moved around with this commit.  The issue is here https://github.com/zcash/zcash-android-wallet-sdk/issues/242

(Although that issue describes the problem, it may not be the implementation we want.)

I can add the issue number to this todo so that its more explicit.

gradle.properties
```diff
@@ -53,7 +53,7 @@ IS_MINIFY_APP_ENABLED=true
 ZCASH_DEBUG_KEYSTORE_PATH=
 
 # Versions
-ANDROID_MIN_SDK_VERSION=16
+ANDROID_MIN_SDK_VERSION=19
```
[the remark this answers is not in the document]
ccjernigan: This would drop support for Android that are over 10 years old, which represents less than 1% of all Android devices.  [Google Play Services requires API 19](https://android-developers.googleblog.com/2021/07/google-play-services-discontinuing-jelly-bean.html), which is a gating factor for apps consuming our SDK distributed through Google Play.

The context is that this line is a fix for regression introduced recently.  GRPC now has a minimum of API 19.  Our options are either:
 - Roll back to an older version of GRPC
 - Keep newer GRPC, and bump our SDK to API 19

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/block/CompactBlockDbStore.kt
ccjernigan: Nice suggestion.  I'll make a change that simplifies this.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
ccjernigan: Our SDK initialization is already kind of heavyweight.  The `Initializer` object needs to load the Rust native libraries (which performs disk IO), so I converted `Initializer.new()` to a suspending function a while back.  An `Initializer` is required to construct the `Synchronizer`.

Since clients already need to make a suspending API call (for the `Initializer`), making `SdkSynchronizer` also use a suspending API call seems like a reasonable tradeoff which simplifies both the public API of the SDK and the implementation.  It makes it much easier to reason about thread-safety, since we have so much less mutable state.

Thoughts?

pacu:
> 



> Do any of these Rust backend methods touch the database or have multithreading concerns? `isValidShieldedAddr`, `isValidTransparentAddr`, `getBranchIdForHeight`

No, they don't. If you check the rust lib.rs files you'll see that these methods just call the rust layer. I don't think they have any multi-threading concerns either

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
pacu: excellent!

pacu (approved):

Issue #526, opened 2022-05-19
Migrate to `zcash_primitives 0.6`


str4d:

str4d:
Closed by #530.

Issue #525, opened 2022-05-19
Change to generating and using UnifiedFullViewingKeys


str4d:
This requires at least the following changes (I'll add more if I find them):
- Delete `UnifiedViewingKey` (which is a Sapling ExtFVK + transparent public key)
- Add `UnifiedFullViewingKey` (being a Sapling ExtFVK +transparent extended public key)
- Change all APIs that generate or use transparent spending keys to instead generate or use transparent extended spending keys.
- Fix Rust APIs that claim to accept transparent extended public keys to actually do so (instead of transparent public keys).

This issue only covers support for new wallets. Migrating existing wallets is covered by #529.

str4d:
We can't entirely switch to UFVKs because support for parsing and serializing them requires #526. So what I want to do in this issue is a "prefactor", where I make as many changes as we feasibly can to migrate towards UAs, but using some private and VERY MUCH NOT PRODUCTION encoding of the individual UFVK components. Then in #526 we can migrate to the proper UFVK encoding, in what should otherwise be a relatively minimal change.

str4d:
Closed by #530.

PR #530, opened 2022-05-20
Replace UnifiedViewingKey with UnifiedFullViewingKey

str4d:
UnifiedViewingKey had a bug that made it incompatible with ZIP 316.

For compatibility with the current `zcash/librustzcash` revision we use
a temporary fake UFVK encoding that stores sufficient information to use
the current APIs, and a superset of the actual ZIP 316 FVK information.

Depends on https://github.com/zcash/librustzcash/pull/570.

Closes zcash/zcash-android-wallet-sdk#525.
Closes zcash/zcash-android-wallet-sdk#526.

# Author

- [ ] Self-review: Did you review your own code in GitHub's web interface? _Code often looks different when reviewing the diff in a browser, making it easier to spot potential bugs._

# Reviewer


sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackend.kt
```diff
@@ -56,18 +56,15 @@ class RustBackend private constructor() : RustBackendWelding {
         )
     }
 
-    override suspend fun initAccountsTable(vararg keys: UnifiedViewingKey): Boolean {
-        val extfvks = Array(keys.size) { "" }
-        val extpubs = Array(keys.size) { "" }
+    override suspend fun initAccountsTable(vararg keys: UnifiedFullViewingKey): Boolean {
+        val ufvks = Array(keys.size) { "" }
```
ccjernigan, line 60: A stylistic choice, but we can probably inline this now and drop the forEach below.  The lambda for the Array() constructor provides an index as `it` and the return value initializes that index.

`Array(keys.size) { keys[it].encoding }`

sdk-lib/src/main/java/cash/z/ecc/android/sdk/type/WalletTypes.kt
```diff
@@ -52,16 +52,16 @@ data class WalletBirthday(
...
+ * does not have spend authority. It can be used to derive a [UnifiedAddress] for the
+ * account.
+ *
+ * @param[encoding] The string encoding of the UFVK.
  */
-data class UnifiedViewingKey(
-    val extfvk: String = "",
-    val extpub: String = ""
+data class UnifiedFullViewingKey(
+    val encoding: String = ""
```
ccjernigan, line 64: This is a pre-existing issue.  I don’t think we ever want to allow `UnifiedFullViewingKey(“”)` 

Since we’re already making a breaking change here, can we drop the empty string as a default constructor parameter?

sdk-lib/Cargo.toml
```diff
@@ -12,6 +12,7 @@ edition = "2018"
 [dependencies]
 android_logger = "0.9"
 failure = "0.1"
+hdwallet = "=0.3.0"
```
ccjernigan, line 15: I noticed most of our dependencies are not locked with `=`. Is there a particular reason this one is?

ccjernigan (commented):
I am going to push commit to this branch later today which includes updates to the MIGRATIONS.md file.  Besides that needing to be updated, I think this looks good.

I did have a few very small comments/questions inline

sdk-lib/src/main/java/cash/z/ecc/android/sdk/type/WalletTypes.kt
ccjernigan: Also are there any invariants that you’d like to enforce here?  (Minimum length?)

ccjernigan: I notice a security issue too.  We should probably drop `data` so this becomes an ordinary class. Data provides a default equals, hash code, and toString implementation.  The `toString` is dangerous in terms of accidentally logging sensitive information. (Alternative is to override the toString to block its default return value).

sdk-lib/Cargo.toml
str4d: The upstream developer merged one of our PRs to bump the `secp256k1` version from 0.19 to 0.21 (a breaking change), and released it as 0.3.1 (which Rust SemVer treats as backwards-compatible). We've migrated to `secp256k1 0.21` in `zcash/librustzcash`, but as of this PR we are still on `secp256k1 0.19` and thus need to pin this temporarily.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/type/WalletTypes.kt
str4d: This is actually something I want to overhaul before we make the breaking release. I agree with making this an opaque class, so we can better control how it is instantiated, and alter its internals freely without breaking changes. I'm not sure I want to do that in this PR though, which is focused on the UFVK breaking API changes that need to affect the FFI. I'll open an issue for addressing this.

str4d: #531

sdk-lib/src/main/rust/utils.rs
```diff
@@ -77,3 +89,75 @@ where
...
+        .derive_public_key(KeyIndex::Normal(0))?
+        .derive_public_key(KeyIndex::Normal(u32::from_le_bytes(
+            index.0[..4].try_into().unwrap(),
+        )))?
+        .public_key;
+    Ok(derive_transparent_address_from_public_key(&pubkey))
+}
+
+/// This is temporary, and will be replaced by `zcash_address::unified::Ufvk`.
+pub(crate) fn fake_ufvk_encode(p2pkh: &ExtendedPubKey, sapling: &ExtendedFullViewingKey) -> String {
```
nuttycom, line 120: Since zcash_address doesn't have any dependency upon zcash_primitives, why not just pull that in now? The `zcash_address` crate didn't even exist at the time of `zcash_primitives = "0.5"` so there isn't any conflict.

str4d: I did try this, but none of the `zcash_primitives::zip32` refactors are present in this branch, so there's a mismatch between the types I have at my disposal, and the data that `zcash_address::unified` allows to be serialized. I figured this is only going to exist for a single PR, and then will disappear, so it's a helpful stepping stone.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackend.kt
str4d: I stuck close to the existing style because I'm not a Kotlin dev 😄 We should just clean this up across the board in a subsequent PR.

MIGRATIONS.md
```diff
@@ -1,7 +1,24 @@
...
+To support Network Upgrade 5, the way keys are generated has changed.
+
+For SDK clients that regenerate the keys from a mnemonic, migration might look like:
+ * Replace usage of `UnifiedViewingKey` with `UnifiedFullViewingKey`
+ * Replace `DerivationTool.deriveUnifiedViewingKeys` with `DerivationTool.deriveUnifiedFullViewingKeys`
+
+For SDK clients that store the key separately from the mnemonic, the migration might look like:
+ * Replace usage of `UnifiedViewingKey(extfvk: String, extpub: String)` with `UnifiedFullViewingKey(encoding: String)`
+ * Replace `DerivationTool.deriveUnifiedViewingKeys` with `DerivationTool.deriveUnifiedFullViewingKeys`
+ * Delete any previously persisted values for `UnifiedViewingKey(extfvk: String, extpub: String)`.  The field `extpub` no longer exists.  The field `extfvk` needs to be re-generated from the mnemonic, and it has been renamed to `encoding` in `UnifiedFullViewingKey(encoding: String)`
```
str4d, line 14: This migration step is incorrect; the field `extfvk` technically doesn't exist any more either. Its value happens to be contained within the `encoding` field as part of the UFVK, but that is an internal detail of UFVKs that isn't relevant to the SDK user.

ccjernigan: Feel free to propose alternative phrasing here or push another commit.

str4d:
Rebased on current `main`.

str4d:
[Force-pushed](https://github.com/zcash/zcash-android-wallet-sdk/compare/da6ef7016e760a0c2b21dff220e1ec481939b9db..58641b7d12b3a0f64792fd2aa949f20b676a1460) to address my comment on `MIGRATIONS.md`.

MIGRATIONS.md
str4d: Done.

pacu (commented):
no specific change. Please verify the changes on verified balance and total balance since regressions had been spotted here 
https://github.com/zcash/ZcashLightClientKit/issues/366

and here https://github.com/zcash-hackworks/zcash-light-client-ffi/issues/8

sdk-lib/src/main/rust/lib.rs
```diff
@@ -467,6 +452,9 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_isValidTran
             Some(addr) => match addr {
                 RecipientAddress::Shielded(_) => Ok(JNI_FALSE),
                 RecipientAddress::Transparent(_) => Ok(JNI_TRUE),
+                // TODO: Should this count? What are the intended semantics of this FFI method?
```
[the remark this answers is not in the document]
str4d: Hmm, so the goal is to verify that it _is_ a t-addr, or is this used to show what kind of address has been scanned? If that's the case, then UAs should return `false` in both of these and we should have a separate API for exposing UA metadata. Or... I can just ignore that and wait until Uniffi gives us much nicer types that make these APIs obvious.

sdk-lib/src/main/rust/utils.rs
str4d: I was able to jump all the way to proper UFVK encodings in this PR, so this is now just an internal detail of the branch.

sdk-lib/src/main/rust/lib.rs
```diff
@@ -138,29 +140,22 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_initAccount
             .map(|i| env.get_object_array_element(ufvks_arr, i))
             .map(|jstr| utils::java_string_to_rust(&env, jstr.unwrap().into()))
             .map(|ufvkstr| {
-                // TODO: replace with `zcash_address::unified::Ufvk`
-                utils::fake_ufvk_decode(&ufvkstr).ok_or_else(|| {
+                UnifiedFullViewingKey::decode(&network, &ufvkstr).map_err(|e| {
+                    if e.starts_with("UFVK is for network") {
```
nuttycom, line 144: This string matching is a bit ugly; can `UnifiedFullViewingKey::decode` return an enum for its error type instead?

sdk-lib/src/main/rust/lib.rs
```diff
@@ -264,16 +258,15 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_tool_DerivationTool_deriveU
...
-        .unwrap();
-        // TODO: replace this with `zcash_address::unified::Address`.
-        let address_str = utils::fake_ua_encode(&p2pkh, &sapling);
+        let account_id = AccountId::from(account_index);
+        let ufvk = UnifiedSpendingKey::from_seed(&network, &seed, account_id)
+            .map_err(|e| format_err!("error generating unified spending key from seed: {:?}", e))
+            .map(|usk| usk.to_unified_full_viewing_key())?;
+
+        let (ua, _) = ufvk
+            .find_address(DiversifierIndex::new())
```
nuttycom, line 267: I think that this should be `DiversifierIndex::zero()` or `DiversifierIndex::from(0)`; `new` doesn't seem like the right constructor for `DiversifierIndex`.

sdk-lib/src/main/rust/lib.rs
```diff
@@ -292,22 +285,20 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_tool_DerivationTool_deriveU
...
-        // Derive the default Sapling payment address (like older SDKs used).
-        let (di, sapling) = ufvk.1.default_address().unwrap();
-        // Derive the transparent address corresponding to the default Sapling diversifier
-        // index (matching ZIP 316).
-        let p2pkh = utils::p2pkh_addr(ufvk.0, di).unwrap();
-        // TODO: replace this with `zcash_address::unified::Address`.
-        let address_str = utils::fake_ua_encode(&p2pkh, &sapling);
+        // Derive the default Unified Address (containing the default Sapling payment
+        // address that older SDKs used).
+        let (ua, _) = ufvk.default_address();
```
nuttycom, line 300: This reminds me that we need to ensure that we handle search for the child 0 "diversified" transparent address correctly even when the Sapling default address is at a different diversifier index.

sdk-lib/src/main/rust/lib.rs
```diff
@@ -443,7 +426,9 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_isValidShie
 
         match RecipientAddress::decode(&network, &addr) {
             Some(addr) => match addr {
-                RecipientAddress::Shielded(_) => Ok(JNI_TRUE),
+                // TODO: What are the intended semantics of this FFI method?
+                // Unified Addresses always have a shielded component.
```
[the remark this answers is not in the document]
nuttycom: The SDKs will support sending to Sapling addresses in all cases, so this question just applies to receiving funds at bare Sapling addresses. I think that we still need to support this at the SDK level; users may need to share shielded addresses with people who haven't yet upgraded to a wallet with UA parsing support.

nuttycom (approved):
utACK with a minor comment about how the librustzcash API that this depends upon could be improved (hopefully before releasing this?)

sdk-lib/src/main/rust/lib.rs
str4d: Yeah, we can rework this API once we've finished working out how it is used. I was more focused on other parts of the stack to get us to this point.

sdk-lib/src/main/rust/lib.rs
str4d: I prefer the `orchard` `DiversifierIndex` type, but we're in the `zcash_*` crates so I used the one from `zcash_primitives`. We should probably figure out a way to reify the types.

nuttycom: Maybe `DiversifierIndex` should be in `zcash_address`?

sdk-lib/src/main/rust/lib.rs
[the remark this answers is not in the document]
str4d: @pacu and @ccjernigan confirmed to me that these APIs are used for _type_ checks (i.e. form validation), so my changes here were incorrect: I instead need to add an `isValidUnifiedAddress` FFI method. These will all be replaced by the `zcash_address` crate once we have Uniffi wrappers.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackend.kt
ccjernigan: I've resolved this with an additional commit.

ccjernigan (changes requested):
Besides a few small comments inline, I pushed some changes in separate commits to resolve build failures on CI (e.g. tests wouldn't compile, linting failed).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/db/entity/Sent.kt
```diff
@@ -26,6 +26,9 @@ data class Sent(
     @ColumnInfo(name = "tx")
     val transactionId: Long = 0,
 
+    @ColumnInfo(name = "output_pool")
+    val outputPool: Int = 0,
```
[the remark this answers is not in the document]
str4d: I'd planned to handle migrations in a separate subsequent PR.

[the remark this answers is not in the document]
pacu: Yes, we do. I had to do it in order for existing wallets to work

sdk-lib/src/main/java/cash/z/ecc/android/sdk/tool/DerivationTool.kt
```diff
@@ -87,17 +87,20 @@ class DerivationTool {
...
-        override suspend fun deriveTransparentSecretKey(seed: ByteArray, network: ZcashNetwork, account: Int, index: Int): String = withRustBackendLoaded {
-            deriveTransparentSecretKeyFromSeed(seed, account, index, networkId = network.id)
+        override suspend fun deriveTransparentAccountPrivateKey(seed: ByteArray, network: ZcashNetwork, account: Int): String = withRustBackendLoaded {
+            deriveTransparentAccountPrivKeyFromSeed(seed, account, networkId = network.id)
         }
 
-        fun validateUnifiedViewingKey(viewingKey: UnifiedViewingKey, networkId: Int = ZcashNetwork.Mainnet.id) {
-            // TODO
-        }
+        // fun validateUnifiedFullViewingKey(
```
pacu, line 98: This is needed for clients implementing receive-only Point-Of-Sale solutions

pacu (changes requested):

sdk-lib/src/main/java/cash/z/ecc/android/sdk/db/entity/Sent.kt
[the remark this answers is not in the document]
ccjernigan: We try to have the main branch be always releasable, e.g. we don't want to intentionally merge a commit that we know will break the main branch for consumers of the SDK.

We're trying to resolve a few different bugs concurrently right now, so I think we need to keep the main branch stable.

We are also about as close to continuous deployment as we can be.  The Secant app consumes snapshots of the SDK so as soon as we merge this PR the Secant app will be ready to consume it.

We should address this migration in some way before merging this, even if we do a followup with an improved migration.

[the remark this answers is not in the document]
pacu: I agree with carter. The migration follows the same pattern as existing ones.

nuttycom:
Oops, I fouled up that rebase, going to redo it.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/db/entity/Sent.kt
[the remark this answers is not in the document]
str4d: I somewhat disagree; my intention by splitting the PRs was to keep the process and review load manageable, as there was a _lot_ to do. But if reviewers are willing to review a noticeably more complex PR in once go, then we can combine them.

str4d:
[Rebased on `main`](https://github.com/zcash/zcash-android-wallet-sdk/compare/3218b18c91eafaf7c1b4c29b0990a6ff9843bf4f..ee8469d7948929fd8b5c5be054a48d3a24ffe41d) to fix merge conflicts due to the recent 1.8 release. I also cleaned up the commit history slightly, and re-targeted this PR to merge into the new `feature/zip-316-and-latest-upstream` branch, which we will use for staging purposes (so we don't break the `main` branch for existing wallets, which this PR alone will do).

sdk-lib/Cargo.toml
str4d: This has since been fixed; we now depend on `hdwallet >= 0.3.1`.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/db/entity/Sent.kt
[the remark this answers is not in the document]
str4d: In a meeting we decided that we'd merge this PR into a feature branch, so we could merge it without migrations (handling those in a separate PR), while waiting until the feature branch will not break `main` before merging it there.

str4d:
[Force-pushed](https://github.com/zcash/zcash-android-wallet-sdk/compare/ee8469d7948929fd8b5c5be054a48d3a24ffe41d..626f6cfa80a7b9cad1545e2e686621c82ca70831):
- Fixed TODOs related to my incomplete understanding of what `isValid{Shielded, Transparent}Address` were being used for; there is now a corresponding `isValidUnifiedAddress` and `AddressType.Unified`.
- Fixed the type signature of `Synchronizer.shieldFunds` (it compiled before because there's no actual type change, but one of the input variables was renamed in `SdkSynchronizer` but not here).
- Made a small `cargo fmt` change to the PR contents (the PR is not completely `cargo fmt` clean because the `main` branch is not; I've left that out of this PR).

sdk-lib/src/main/rust/lib.rs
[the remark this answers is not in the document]
str4d: This was clarified in a meeting: this API is for type checking (form validation). I've fixed it to treat `isValidShieldedAddress` as only checking for `RecipientAddress::Shielded`, with a new `isValidUnifiedAddress` handling `RecipientAddress::Unified`.

sdk-lib/src/main/rust/lib.rs
```diff
@@ -520,7 +508,7 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_getVerified
         let taddr = TransparentAddress::decode(&network, &addr).unwrap();
 
         let amount = (&db_data)
-            .get_target_and_anchor_heights()
+            .get_target_and_anchor_heights(ANCHOR_OFFSET)
```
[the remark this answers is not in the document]
str4d: This change should be a no-op. Previously `get_target_and_anchor_heights` internally used a fixed offset, and this PR just maintains that fixed offset.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/tool/DerivationTool.kt
str4d: It's never been implemented, so *shrug*

In any case, the PR now keeps this method in (and it continues to be a no-op).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/type/WalletTypes.kt
str4d: As we're now using a feature branch, we can ensure that we complete #531 before merging it.

CHANGELOG.md
```diff
@@ -12,6 +12,45 @@ Version 1.6.0-beta01
...
+  - This type had a bug where the `extpub` field actually was storing a plain transparent
+    public key, and not the extended public key as intended. This made it incompatible
+    with ZIP 316.
+- `cash.z.ecc.android.sdk.tool`:
+  - `DerivationTool.deriveTransparentAddressFromPrivateKey` (use
+    `DerivationTool.deriveTransparentAddressFromAccountPrivateKey` instead).
+  - `DerivationTool.deriveTransparentSecretKey` (use
+    `DerivationTool.deriveTransparentAccountPrivateKey` instead).
+  - `DerivationTool.deriveShieldedAddress`
+    - TODO: Do we still need to be able to derive Sapling shielded addresses for legacy
```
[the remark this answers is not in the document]
str4d: We can address this later on in the feature branch, once we've got a working integration with the Android apps.

str4d:
[Force-pushed](https://github.com/zcash/zcash-android-wallet-sdk/compare/626f6cfa80a7b9cad1545e2e686621c82ca70831..d7f779752e02be8df05c04874e0bc4e1c9cd647a) to undo some now-unnecessary compile warning fixes (that after rebasing on `main` began causing compile errors).

str4d:
[Force-pushed](https://github.com/zcash/zcash-android-wallet-sdk/compare/d7f779752e02be8df05c04874e0bc4e1c9cd647a..a86ba9855d7d548d641794bc4dab3110ad29089a) to squash one of the commits to clean up the PR history.

str4d:
[Force-pushed](https://github.com/zcash/zcash-android-wallet-sdk/compare/a86ba9855d7d548d641794bc4dab3110ad29089a..d1d406229880f0c12456ba87762454c13518cf91) to fix more Detekt warnings.

MIGRATIONS.md
```diff
@@ -1,7 +1,22 @@
...
+To support Network Upgrade 5, the way keys are generated has changed.
+
+For SDK clients that regenerate the keys from a mnemonic, migration might look like:
+ * Replace usage of `UnifiedViewingKey` with `UnifiedFullViewingKey`
+ * Replace `DerivationTool.deriveUnifiedViewingKeys` with `DerivationTool.deriveUnifiedFullViewingKeys`
+
+For SDK clients that store the key separately from the mnemonic, the migration might look like:
+ * Replace usage of `UnifiedViewingKey(extfvk: String, extpub: String)` with `UnifiedFullViewingKey(encoding: String)`
+ * Replace `DerivationTool.deriveUnifiedViewingKeys` with `DerivationTool.deriveUnifiedFullViewingKeys`
+ * Delete any previously persisted values for `UnifiedViewingKey(extfvk: String, extpub: String)`.
```
daira, line 15:
```suggestion
 * Delete any previously persisted values for `UnifiedViewingKey(extfvk: String, extpub: String)`,
   provided that they can be rederived from the mnemonic.
```

darkside-test-lib/src/androidTest/java/cash/z/ecc/android/sdk/darkside/test/DarksideTestCoordinator.kt
```diff
@@ -144,8 +144,8 @@ class DarksideTestCoordinator(val wallet: TestWallet) {
 
         fun validateHasBlock(height: BlockHeight) {
             runBlocking {
-                assertTrue((synchronizer as SdkSynchronizer).findBlockHashAsHex(height) != null)
-                assertTrue((synchronizer as SdkSynchronizer).findBlockHash(height)?.size ?: 0 > 0)
+                assertTrue((synchronizer).findBlockHashAsHex(height) != null)
```
daira, line 147: Non-blocking nit:
```suggestion
                assertTrue(synchronizer.findBlockHashAsHex(height) != null)
```
and similarly in other cases.

sdk-lib/Cargo.lock
```diff
@@ -1598,53 +1904,77 @@ dependencies = [
...
+ "chacha20",
+ "chacha20poly1305",
+ "rand_core",
+ "subtle",
+]
+
+[[package]]
+name = "zcash_note_encryption"
+version = "0.1.0"
+source = "git+https://github.com/zcash/librustzcash.git?rev=b52e949bd6ee1dbbc0324437440bed8b3c423f79#b52e949bd6ee1dbbc0324437440bed8b3c423f79"
```
daira, line 1930: This is using two different revisions of zcash_note_encryption.

sdk-lib/Cargo.toml
```diff
@@ -12,28 +12,26 @@ edition = "2018"
...
+# with NU5 updates yet.
 [patch.crates-io]
-zcash_client_backend = { git = 'https://github.com/zcash/librustzcash.git', rev='0a1ed9b8acf336bdc20ba49083c66f114ecf6877' }
-zcash_client_sqlite = { git = 'https://github.com/zcash/librustzcash.git', rev='0a1ed9b8acf336bdc20ba49083c66f114ecf6877' }
-zcash_primitives = { git = 'https://github.com/zcash/librustzcash.git', rev='0a1ed9b8acf336bdc20ba49083c66f114ecf6877' }
-zcash_proofs = { git = 'https://github.com/zcash/librustzcash.git', rev='0a1ed9b8acf336bdc20ba49083c66f114ecf6877' }
+zcash_client_backend = { git = 'https://github.com/zcash/librustzcash.git', rev='b52e949bd6ee1dbbc0324437440bed8b3c423f79' }
+zcash_client_sqlite = { git = 'https://github.com/zcash/librustzcash.git', rev='b52e949bd6ee1dbbc0324437440bed8b3c423f79' }
+zcash_primitives = { git = 'https://github.com/zcash/librustzcash.git', rev='b52e949bd6ee1dbbc0324437440bed8b3c423f79' }
+zcash_proofs = { git = 'https://github.com/zcash/librustzcash.git', rev='b52e949bd6ee1dbbc0324437440bed8b3c423f79' }
```
daira, line 34: Adding a patch for zcash_note_encryption here might fix the duplication, I think.

sdk-lib/src/main/rust/lib.rs
```diff
@@ -9,60 +9,61 @@ use std::str::FromStr;
...
-    zip32::ExtendedFullViewingKey,
+    zip32::{AccountId, DiversifierIndex, ExtendedFullViewingKey},
 };
 use zcash_proofs::prover::LocalTxProver;
 
 use crate::utils::exception::unwrap_exc_or;
 
 mod utils;
 
+const ANCHOR_OFFSET: u32 = 10;
```
daira, line 65: This is not what we want long-term (see draft ZIP 315), but it will do for now.

sdk-lib/src/main/rust/lib.rs
daira: Please file an issue so that we don't lose track.

sdk-lib/src/main/rust/lib.rs
```diff
@@ -226,40 +219,30 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_tool_DerivationTool_deriveU
...
+            .collect::<Result<_, _>>()?;
 
-        Ok(utils::rust_vec_to_java_2d(
+        Ok(utils::rust_vec_to_java(
             &env,
-            extfvks,
-            extpubs,
-            |env, extfvkstr| env.new_string(extfvkstr),
+            ufvks,
+            "java/lang/String",
```
daira, line 236: Abstractions leaking, leaking everywhere. (Don't mind me. Just grumbling at the state of software infrastructure.)

sdk-lib/src/main/rust/lib.rs
str4d: Yep, this is intentionally a no-op in this PR.

sdk-lib/src/main/rust/lib.rs
daira: Please file an issue.

sdk-lib/src/main/rust/lib.rs
```diff
@@ -444,7 +427,7 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_isValidShie
         match RecipientAddress::decode(&network, &addr) {
             Some(addr) => match addr {
                 RecipientAddress::Shielded(_) => Ok(JNI_TRUE),
-                RecipientAddress::Transparent(_) => Ok(JNI_FALSE),
+                RecipientAddress::Transparent(_) | RecipientAddress::Unified(_) => Ok(JNI_FALSE),
```
daira, line 430: Document in the API docs that this returns false for unified addresses.

sdk-lib/src/main/rust/lib.rs
```diff
@@ -529,12 +534,13 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_getVerified
...
-                    .get_unspent_transparent_utxos(&taddr, anchor - ANCHOR_OFFSET)
+                    .get_unspent_transparent_outputs(&taddr, anchor - ANCHOR_OFFSET)
                     .map_err(|e| format_err!("Error while fetching verified balance: {}", e))
             })?
             .iter()
-            .map(|utxo| utxo.value)
-            .sum::<Amount>();
+            .map(|utxo| utxo.txout.value)
+            .sum::<Option<Amount>>()
+            .ok_or_else(|| format_err!("Balance overflowed MAX_MONEY. Lucky you!"))?;
```
daira, line 543:
```suggestion
            .ok_or_else(|| format_err!("Balance overflowed MAX_MONEY"))?;
```

sdk-lib/src/main/rust/lib.rs
```diff
@@ -566,12 +572,13 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_getTotalTra
...
-                    .get_unspent_transparent_utxos(&taddr, anchor)
+                    .get_unspent_transparent_outputs(&taddr, anchor)
                     .map_err(|e| format_err!("Error while fetching verified balance: {}", e))
             })?
             .iter()
-            .map(|utxo| utxo.value)
-            .sum::<Amount>();
+            .map(|utxo| utxo.txout.value)
+            .sum::<Option<Amount>>()
+            .ok_or_else(|| format_err!("Balance overflowed MAX_MONEY. Lucky you!"))?;
```
daira, line 581:
```suggestion
            .ok_or_else(|| format_err!("Balance overflowed MAX_MONEY"))?;
```

sdk-lib/src/main/rust/lib.rs
```diff
@@ -943,17 +969,31 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_tool_DerivationTool_deriveT
...
 ) -> jstring {
     let res = panic::catch_unwind(|| {
         let network = parse_network(network_id as u32)?;
-        let tsk_wif = utils::java_string_to_rust(&env, secret_key);
-        let sk: SecretKey = (&Wif(tsk_wif)).try_into().expect("invalid private key WIF");
-        let taddr = derive_transparent_address_from_secret_key(&sk).encode(&network);
+        let index = if index >= 0 {
+            index as u32
+        } else {
+            return Err(format_err!("index argument must be positive"));
```
daira, line 984:
```suggestion
            return Err(format_err!("index argument must be nonnegative"));
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/db/entity/Account.kt
```diff
@@ -11,11 +11,11 @@ data class Account(
...
-    val extendedFullViewingKey: String = "",
+    @ColumnInfo(name = "ufvk")
+    val unifiedFullViewingKey: String? = "",
 
-    val address: String = "",
+    val address: String? = "",
 
     @ColumnInfo(name = "transparent_address")
-    val transparentAddress: String = ""
+    val transparentAddress: String? = ""
```
pacu, line 20: Do we still need this if we have the UFVK?

pacu (commented):
Looks good to me. 
Left a few unblocking comments.

Little bit jealous for this PR to be honest 😅

sdk-lib/src/main/rust/lib.rs
```diff
@@ -1062,7 +1109,7 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_createToAdd
```
daira, line 1109: On line 1078, "account argument must be nonnegative". Also look for other incorrect uses of "positive".

sdk-lib/src/main/rust/utils.rs
```diff
@@ -41,39 +51,63 @@ where
...
-{
-    let jempty = empty_element(env).expect("Couldn't create Java string!");
-    let outer = env
-        .new_object_array(
-            data1.len() as jsize,
-            "[Ljava/lang/String;",
-            *jni::objects::JObject::null(),
-        )
-        .expect("Couldn't create Java array of string arrays!");
+//pub(crate) fn rust_vec_to_java_2d<'a, T, V, F, G>(
```
daira, line 54: Just delete this. It can always be restored if needed.

daira (approved):
utACK modulo suggestions.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/db/entity/Account.kt
str4d: No, but that requires a separate database change (specifically https://github.com/zcash/librustzcash/issues/489).

pacu (approved):

sdk-lib/src/main/rust/lib.rs
nuttycom: https://github.com/zcash/librustzcash/issues/599

nuttycom (approved):
utACK for merge to the feature branch. I will address @daira's documentation comments and then merge.

nuttycom:
[Force-pushed](https://github.com/zcash/zcash-android-wallet-sdk/compare/d1d406229880f0c12456ba87762454c13518cf91..a3bb9cb7bb86ab3fc1f507f5513ea89c2b7fc173) to address review comments.

nuttycom (approved):
re-utACK a3bb9cb7bb86ab3fc1f507f5513ea89c2b7fc173

ccjernigan (approved):

nuttycom (approved):
re-utACK https://github.com/zcash/zcash-android-wallet-sdk/pull/530/commits/7152c7be431bacd5602030c4fc4c29a4d9a1cd82

Issue #474, opened 2022-04-08
Add `BlockHeight(Long)` class


ccjernigan:
## Is your feature request related to a problem? Please describe.
#471

## Describe the solution you'd like
uint32 exists in Kotlin, but not Java.

For Java interoperability, create a `BlockHeight(Long)` class with constructor constraints on a long's range of values allows us to represent a UInt32 safely and resolve the downcasting bug.  Further, this BlockHeight can be consumed by WalletBirthday and the initialization for the SDK.

```
// Not using value class for initial version due to name mangling and overloads
// https://kotlinlang.org/docs/inline-classes.html#calling-from-java-code
class BlockHeight internal constructor(val value: Long) {
   init {
       require(UINT_RANGE.contains(value)) { "Value $value is outside of allowed range $UINT_RANGE" }
   }
   companion object {
       private val UINT_RANGE = 0.toLong()..UInt.MAX_VALUE.toLong()
   }
}
```

Consider making the BlockHeight constructor internal.  Add a block height factory function to ZcashNetwork to ensure that BlockHeights are valid for a given network (e.g. above saplingActivationHeight). Example of new method below:

```
enum class ZcashNetwork(val id: Int, val networkName: String, val saplingActivationHeight: Int, val defaultHost: String, val defaultPort: Int) {
   Testnet(0, "testnet", 280_000, "testnet.lightwalletd.com", 9067),
   Mainnet(1, "mainnet", 419_200, "mainnet.lightwalletd.com", 9067);
  
   fun newBlockHeight(value: Long) {
       require(value >= saplingActivationHeight)
      
       BlockHeight(value)
   }
}
```

## API Migration Plan
1. Modify current APIs to use a Long instead of Int (this is breaking, but a minor transition).  Java clients shouldn't break here, but Kotlin clients will require casting to Long.
1. Identify APIs that need to consume the BlockHeight object:
  - Synchronizer.networkHeight
  - Synchronizer.latestHeight
  - Synchronizer.latestBirthdayHeight
  - Synchronizer.getNearestRewindHeight()
  - Synchronizer.rewindToNearestHeight()
  - Synchronizer.refreshUtxos(tAddr: String, sinceHeight: Int)
  - Initializer.Config.birthdayHeight
  - ZcashNetwork.saplingActivationHeight
  - WalletBirthday.blockHeight
1. Have the old APIs wrap the new APIs, so that the old APIs can be deleted in a future release

[...]

PR #594, opened 2022-07-05
[#474] Add BlockHeight object

ccjernigan:
This adds a `BlockHeight(Long)` object to improve type safety of the public APIs.  The preferred way to construct a BlockHeight is through a factory method `new` which also takes in the `ZcashNetwork` to ensure that the height is above the network's `saplingActivationHeight`.

As part of this refactor, `WalletBirthday` was renamed to `Checkpoint` and moved to be an internal API.  SDK clients now use `BlockHeight` to represent a birthday.

# Author

- [ ] Documentation: Did you update documentation as appropriate? (e.g [README.md](../blob/main/README.md), etc.)

# Reviewer


sdk-lib/src/main/java/cash/z/ecc/android/sdk/db/entity/Transactions.kt
```diff
@@ -81,8 +82,8 @@ data class PendingTransactionEntity(
     override val value: Long = -1,
     override val memo: ByteArray? = byteArrayOf(),
     override val accountIndex: Int,
-    override val minedHeight: Int = -1,
-    override val expiryHeight: Int = -1,
+    override val minedHeight: Long = -1,
```
ccjernigan, line 85: TODO: I missed refactoring these.

ccjernigan: Actually never mind, these are database objects so they have to map to SQLite columns

sdk-lib/src/test/java/cash/z/ecc/android/sdk/ext/ConversionsTest.kt
```diff
@@ -9,7 +10,7 @@ internal class ConversionsTest {
 
     @Test
     fun `default right padding is 6`() {
-        assertEquals(1.13.toZec(6), 113000000L.convertZatoshiToZec())
+        assertEquals(1.13.toZec(6), Zatoshi(113000000L).convertZatoshiToZec())
```
pacu, line 13: Is this from another PR?

ccjernigan: Yes I found a few stragglers from the prior PR for Zatoshi.  In this PR, I've tried to isolate them to a separate commit.

pacu: ok! np!. just saying!

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
```diff
@@ -193,25 +201,33 @@ class CompactBlockProcessor(
...
+                    }
+                    is BlockProcessingResult.Error -> {
+                        val napTime = calculatePollInterval(true)
+                        twig("Unable to process new blocks because they failed validation! Attempting to reconnect in ${napTime}ms")
+                        delay(napTime)
+                    }
+                    is BlockProcessingResult.Success -> {
+                        // Do nothing. We are done.
                     }
-                    consecutiveChainErrors.getAndIncrement()
```
pacu, line 214: I think that the consecutive errors count is not being updated as it was. Was this moved somewhere else?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
```diff
@@ -278,19 +299,36 @@ class CompactBlockProcessor(
...
                     networkBlockHeight = initialInfo.networkBlockHeight,
                     lastScannedHeight = initialInfo.lastScannedHeight,
                     lastDownloadedHeight = initialInfo.lastDownloadedHeight,
-                    lastScanRange = (initialInfo.lastScannedHeight + 1)..initialInfo.networkBlockHeight,
-                    lastDownloadRange = (
-                        max(
-                            initialInfo.lastDownloadedHeight,
-                            initialInfo.lastScannedHeight
-                        ) + 1
+                    lastScanRange = if (initialInfo.lastScannedHeight != null && initialInfo.networkBlockHeight != null) {
```
pacu, line 316: I wonder if these conditions are part of logic we might want to "document" in code by naming them under a computed variable or helper function.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
```diff
@@ -278,19 +299,36 @@ class CompactBlockProcessor(
...
+                        initialInfo.lastScannedHeight + 1..initialInfo.networkBlockHeight
+                    } else {
+                        null
+                    },
+                    lastDownloadRange = if (initialInfo.lastDownloadedHeight != null && initialInfo.lastScannedHeight != null && initialInfo.networkBlockHeight != null) {
```
pacu, line 321: I wonder if these conditions are part of logic we might want to "document" in code by naming them under a computed variable or helper function.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
```diff
@@ -306,35 +344,34 @@ class CompactBlockProcessor(
...
-            val success = scanNewBlocks(lastScanRange)
-            if (!success) throw CompactBlockProcessorException.FailedScan()
-            else {
-                setState(Scanned(lastScanRange))
+     */
+    private suspend fun validateAndScanNewBlocks(lastScanRange: ClosedRange<BlockHeight>?): BlockProcessingResult =
+        withContext(IO) {
+            setState(Validating)
+            val error = validateNewBlocks(lastScanRange)
+            if (error == BlockProcessingResult.Success) {
```
pacu, line 352: In the spirit of improving the code, guess we could find a better name for this `val error` since  `error = success` seems like a rhetorical question 😅

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
```diff
@@ -306,35 +344,34 @@ class CompactBlockProcessor(
...
+                val success = scanNewBlocks(lastScanRange)
+                if (!success) {
+                    throw CompactBlockProcessorException.FailedScan()
+                } else {
+                    setState(Scanned(lastScanRange))
+                }
             }
-            ERROR_CODE_NONE
-        } else {
+
```
pacu, line 364: I think there's a dangling vertical space here

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
```diff
@@ -426,30 +477,35 @@ class CompactBlockProcessor(
...
 
-    internal suspend fun processUtxoResult(result: List<Service.GetAddressUtxosReply>, tAddress: String, startHeight: Int): Int = withContext(IO) {
+    internal suspend fun processUtxoResult(
+        result: List<Service.GetAddressUtxosReply>,
+        tAddress: String,
+        startHeight: BlockHeight
+    ): Int = withContext(IO) {
         var skipped = 0
-        val aboveHeight = startHeight - 1
+        val aboveHeight = startHeight + -1
```
pacu, line 508: is this kind of syntax expected on kotlin or would it be better to have a `decrement` helper for the type?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/BlockHeight.kt
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Represents a block height, which is a UInt32.  SDK clients use this class to represent the "birthday" of a wallet.
+ *
+ * @param value The block height.  Must be in range of a UInt32.
+ */
+/*
+ * For compatibility with Java clients, this class represents the height value as a Long with
+ * assertions to ensure that it is a 32-bit unsigned integer.
+ */
+data class BlockHeight internal constructor(val value: Long) : Comparable<BlockHeight> {
```
pacu, line 14: I'd consider adding decrement helper function given the number of places `height + -1` is used. I haven't noticed occurrences of other decrements that are not just subtracting one.

pacu (changes requested):

pacu:
This is a huge change and improvement! 🥳 

Left a few comments!

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
ccjernigan: This is a weird special case that invented today.  It is a kludge to support support the existing logic in CompactBlockProcessor, although it could probably have a further refactor to make this cleaner.

BlockHeight has an `add` operator, but no subtract operator.  That's to ensure that subtracting two BlockHeights don't violate the representation invariant of the height being above the sapling activation height for the network.  BlockHeights don't store the network, so a subtraction can't know whether the result is valid.  (We only validate when using the `BlockHeight.Companion.new` factory method).  Because I intentionally left out the subtract operator, I'm doing plus -1 to effectively do a subtract.  The add operator allows this, but doesn't allow plus -2 since the only places we ever subtract was at most -1.

I don't like this solution, but it doesn't leak into the public API so it is mostly not too bad.

Two alternatives are
 - Do a second pass to update CompactBlockProcessor to change the bounds so that -1 isn't necessary
 - Keep the network inside the BlockHeight, so that subtraction can be validated.  (I don't like this option).

Thoughts?

pacu: > Do a second pass to update CompactBlockProcessor to change the bounds so that -1 isn't necessary

If the subtractions are bound to the CompactBlockProcessor, then probably we can be fine with BlockHeight not needing any math and refactor the processor, which is the class introducing this requirement. Do you have a ballpark estimate of what the refactor would be?

ccjernigan: I'll have to scope this out and follow up.

There is at least one place where I think this impacts the Rust code, due to inclusive versus exclusive boundary conditions.  So in that particular case, changing it would be more difficult.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/PendingTransactionDb.kt
```diff
@@ -70,10 +70,10 @@ interface PendingTransactionDao {
     suspend fun removeRawTransactionId(id: Long)
 
     @Query("UPDATE pending_transactions SET minedHeight = :minedHeight WHERE id = :id")
-    suspend fun updateMinedHeight(id: Long, minedHeight: Int)
+    suspend fun updateMinedHeight(id: Long, minedHeight: Long)
 
     @Query("UPDATE pending_transactions SET raw = :raw, rawTransactionId = :rawTransactionId, expiryHeight = :expiryHeight WHERE id = :id")
-    suspend fun updateEncoding(id: Long, raw: ByteArray, rawTransactionId: ByteArray, expiryHeight: Int?)
+    suspend fun updateEncoding(id: Long, raw: ByteArray, rawTransactionId: ByteArray, expiryHeight: Long?)
```
pacu, line 76: Does this imply a migration or it automatically figures it out?

pacu (approved):
LGTM! I didn't test it.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/PendingTransactionDb.kt
ccjernigan: SQLite doesn't use strong types, so this should be fine as-is.  https://www.sqlite.org/datatype3.html

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
ccjernigan: This should be resolved now.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
ccjernigan: I think it only gets called once.  Do you have a recommendation for a nicer name?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
ccjernigan: I think it only gets called once.  Do you have a recommendation for a nicer name?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
ccjernigan: I fixed this

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
ccjernigan: I can't find this anymore

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
ccjernigan: So I did a few things to make this better.  It isn't 100% but its most of the way there.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/PendingTransactionDb.kt
pacu: Oh yeah. that gives me nightmares. I'm in denial 😅

pacu (approved):

Issue #674, opened 2022-08-17
Use suspending functions for networking


ccjernigan:
## Is your feature request related to a problem? Please describe.
`CompactBlockProcessor` is using blocking GRPC calls, which means each network call is wrapped in a `withContext`.  This makes reasoning about network connections very complex.

## Describe the solution you'd like
Switch `CompactBlockProcessor` to use `CoroutineLightWalletClient`.

HonzaR:
Note about `FileCompactBlockRepositoryTest#getLatestHeightTest` -> `FileCompactBlockRepository#write` refactoring. Write function could work with flow as well.

Issue #673, opened 2022-08-17
Hide GRPC from public API


ccjernigan:
## Is your feature request related to a problem? Please describe.
The SDK leaks GRPC classes into the public API.  This makes our public API dependent on the public API of another library.  If we ever wanted to switch networking implementations, this would be a difficult and breaking change.

## Describe the solution you'd like
Building upon #615, add additional Unsafe model objects to the lightwallet-client-lib and translate the GRPC objects to and from these representations owned by the library.

HonzaR:
Solved as part of this https://github.com/zcash/zcash-android-wallet-sdk/pull/916.

Issue #615, opened 2022-07-20
Refactor network connectivity


ccjernigan:
## Is your feature request related to a problem? Please describe.
The GRPC connectivity in the app is somewhat isolated, but it still leaks some implementation details into the public API.  It also bypasses some of the type safety that we're trying to enable.

For example
 - `LightWalletService` exposes generated GRPC objects
 - GRPC objects are in the SDK's public API with the namespace `cash.z.wallet.sdk.rpc`
 - LightWalletService does use `BlockHeight` and instead uses Int for blocks

## Describe the solution you'd like
 - Consider moving the networking to a separate Gradle module, which simplifies and isolates the networking layer of the SDK
 - Change the API of LightWalletService to use our own model objects.  If necessary, create additional model objects and convert between the generated GRPC objects to our own.
 - Repackage `cash.z.wallet.sdk.rpc` to `cash.z.wallet.sdk.internal.rpc`
 
 ## Additional context:
- Prepare a separate module for it
- Use Coroutines (start with blocking first, then coroutines refactor in followup PR)
- Add dedicated model classes in  this module (convert GRPC model objects to our own network model objects, which then get converted later to the SDK’s model objects)
- Mapping GRPC errors into a more sane networking API of request/response (draft in place, but some work likely needed on the mapping)
- It will enable us also obtaining a new checkpoint files instead of using the Checkmate tool

PR #675, opened 2022-08-17
[#615] Refactor lightwalletd client

ccjernigan:
I've created a draft PR just to start discussion on how changes to the API look.

This moves the lightwalletd client code to a separate Gradle module, but does not yet do much clean up the public API (e.g. hiding generated GRPC objects).  That will be a followon change via #673

I’ve prototyped a safer version of the API that would be implemented for #673 for two specific calls: obtaining the current block height and obtaining the light wallet info.  These were the easiest endpoints to update and they povide a useful exploration for the future changes needed.

#674 will also update CompactBlockProcessor to use coroutines and simplify dispatching to a separate thread for network requests.

# Author

# Reviewer


sdk-lib/src/androidTest/java/cash/z/ecc/android/sdk/integration/service/ChangeServiceTest.kt
```diff
@@ -79,99 +71,8 @@ class ChangeServiceTest : ScopedTest() {
             twig(it)
         }.getOrElse { return }
 
-        assertTrue(result > network.saplingActivationHeight)
-    }
-
-    @Test
-    fun testCleanSwitch() = runBlocking {
```
pacu, line 86: are you planning to create these later on ? otherwise the name of the test class should be changed

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -797,23 +779,23 @@ object DefaultSynchronizerFactory {
             initializer.rustBackend.cacheDbFile
         )
 
-    fun defaultService(initializer: Initializer): LightWalletService =
-        LightWalletGrpcService.new(initializer.context, initializer.lightWalletEndpoint)
+    fun defaultLightWalletClient(initializer: Initializer): BlockingLightWalletClient =
```
pacu, line 782: I assume this maintains current functionality where a suspending function calls a blocking client. 
If we are planning to move to the coroutines implementation of the service, don't forget to create a github issue about it! :)

pacu (changes requested):

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
ccjernigan: https://github.com/zcash/zcash-android-wallet-sdk/issues/674

sdk-lib/src/androidTest/java/cash/z/ecc/android/sdk/integration/service/ChangeServiceTest.kt
ccjernigan: We won't support switching right now.  Instead, SDK consumers should stop the synchronizer and start it again with new arguments.

In the future, the networking abstraction may take care of automatically switching between multiple endpoints if they are available.

lightwallet-client-lib/src/androidTest/resources/utils/seeds.txt
```diff
@@ -0,0 +1,18 @@
+kitchen renew wide common vague fold vacuum tilt amazing pear square gossip jewel month tree shock scan alpha just spot fluid toilet view dinner
```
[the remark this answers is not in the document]
ccjernigan: Yes, these should be deleted.

gradle.properties
```diff
@@ -59,57 +60,74 @@ ZCASH_RELEASE_KEYSTORE_PASSWORD=
...
+ANDROIDX_TEST_VERSION=1.5.0
+ANDROIDX_TRACING_VERSION=1.2.0-alpha01
+ANDROIDX_UI_AUTOMATOR_VERSION=2.3.0-alpha01
 BIP39_VERSION=1.0.4
 COROUTINES_OKHTTP=1.0
-GOOGLE_MATERIAL_VERSION=1.6.1
-GRPC_VERSION=1.48.1
+GOOGLE_MATERIAL_VERSION=1.7.0
 GRPC_KOTLIN_VERSION=1.3.0
+GRPC_VERSION=1.51.0
```
pacu, line 118: Have you updated the generated files with `protoc` ?

pacu (changes requested):

gradle.properties
ccjernigan: I think they get generated automatically as part of the Gradle build, based upon the contents of src/main/proto.  See https://github.com/zcash/zcash-android-wallet-sdk/blob/7dd6d3c3429a456e6a0308925026f3e0c27dfc3b/lightwallet-client-lib/build.gradle.kts#L166

Is there something specific that you were thinking about?

pacu: Great!
> Is there something specific that you were thinking about?


Nothing in particular, just checking!

pacu (approved):
utACK

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/CoroutineLightWalletClientImpl.kt
```diff
@@ -51,11 +53,23 @@ internal class CoroutineLightWalletClientImpl private constructor(
...
     }
 
-    override suspend fun getLatestBlockHeight(): BlockHeightUnsafe {
-        return BlockHeightUnsafe(
-            requireChannel().createStub(singleRequestTimeout)
-                .getLatestBlock(Service.ChainSpec.newBuilder().build()).height
-        )
+    override suspend fun getLatestBlockHeight(): Response<BlockHeightUnsafe> {
+        return try {
+            if (BenchmarkingExt.isBenchmarking()) {
```
pacu, line 58: I might be missing some context so I could be wrong.  If so, dismiss the comment. 

I find this rather odd given the fact that this is an implementation for an abstraction. You probably want a decorator that wraps this method and used specifically as a backend when doing benchmarking instead of having this `if` statement that depends on something that seems to be a global value.

pacu (changes requested):
Please see comment on height injection for benchmarking.

I also have a doubt on the `unsafe` suffix that some types have. I went through the docs on github didn't find something about what unsafe means to us, why it is there, when should something be flagged as unsafe, etc. 

If there is, awesome! my bad 🙏 if there's not, we should add it. I don't think there are unsafe things on the public APIs (not at least on this PR) but we should document this for other contributors to know.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/CoroutineLightWalletClientImpl.kt
ccjernigan: We don’t yet have a good way to inject this.  We have a plan to improve this, although it will take multiple iterations to get there.

The refactoring the lightwalletd client will give us an interface that different networking implementations can extend.  Once we have that stabilized (which this PR moves us closer to), then we can pass an alternative networking implementation into the synchronizer which will eliminate the need for this hack.

ccjernigan:
> Please see comment on height injection for benchmarking.
> 
> I also have a doubt on the `unsafe` suffix that some types have. I went through the docs on github didn't find something about what unsafe means to us, why it is there, when should something be flagged as unsafe, etc.
> 
> If there is, awesome! my bad 🙏 if there's not, we should add it. I don't think there are unsafe things on the public APIs (not at least on this PR) but we should document this for other contributors to know.

At the moment we aren’t considering lightwalletd client to be a public API. Someone could re-use it (we talked about using it for checkpoint downloads), but ultimately it is serving needs as an internal API used by the SDK.

The `unsafe` approach is new and something we’re experimenting with.  Our thought process is that Models coming from the network are untrusted and not validated.  For example, a BlockHeightUnsafe could include a block outside of the valid range for the current network.  We want to make usage of those objects inside the SDK clear, so that conversion from BlockHeightUnsafe to BlockHeight is set up to handle errors in conversion without crashing the SDK.  We could also end in some other suffix like Network.  Whatever suffix there is, it will probably have some Smurf naming going on.  We want to avoid sending GRPC objects out of the networking module, so that we aren’t explicitly tying the entire SDK’s implementation to GRPC.  Doing this will also make testing easier.

Another way to explain our current thought process is that we have multiple representations of data in the SDK:
 - Network representation (GRPC objects)
 - Lightwalletd client representation (API of the networking module).  Due to the module dependency graph, the public API representation isn’t visible here so lightwalletd client needs to expose its own representation for internal consumption.
 - Database representation
 - Public API representation

Do you have other suggested approaches?

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/BlockingLightWalletClientImpl.kt
```diff
@@ -0,0 +1,191 @@
...
+                .setStartHeight(startHeight.value).build()
+        )
+        return result.addressUtxosList
+    }
+
+    override fun getTAddressTransactions(
+        tAddress: String,
+        blockHeightRange: ClosedRange<BlockHeightUnsafe>
+    ): Sequence<Service.RawTransaction> {
+        if (blockHeightRange.isEmpty() || tAddress.isBlank()) return emptySequence()
```
ccjernigan, line 135: Long-term, should we consider throwing an IllegalArgumentException if these arguments are not correct?

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/CoroutineLightWalletClientImpl.kt
```diff
@@ -0,0 +1,197 @@
...
+    private val singleRequestTimeout: Duration = 10.seconds,
+    private val streamingRequestTimeout: Duration = 90.seconds
+) : CoroutineLightWalletClient {
+
+    private var channel = channelFactory.newChannel(lightWalletEndpoint)
+
+    /* LightWalletService implementation */
+
+    override fun getBlockRange(heightRange: ClosedRange<BlockHeightUnsafe>): Flow<CompactFormats.CompactBlock> {
+        if (heightRange.isEmpty()) {
```
ccjernigan, line 49: Long-term, should this be an IllegalArgumentException?

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/CoroutineLightWalletClientImpl.kt
```diff
@@ -0,0 +1,197 @@
...
+                .setStartHeight(startHeight.value).build()
+        )
+        return result.addressUtxosList
+    }
+
+    override fun getTAddressTransactions(
+        tAddress: String,
+        blockHeightRange: ClosedRange<BlockHeightUnsafe>
+    ): Flow<Service.RawTransaction> {
+        if (blockHeightRange.isEmpty() || tAddress.isBlank()) return emptyFlow()
```
ccjernigan, line 141: Long-term should this be an IllegalArgumentException?

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/GrpcStatusResolver.kt
```diff
@@ -0,0 +1,68 @@
...
+
+import android.util.Log
+import co.electriccoin.lightwallet.client.ApiStatusResolver
+import co.electriccoin.lightwallet.client.model.Response
+import io.grpc.Status
+
+/**
+ * This class provides conversion from GRPC Status to our predefined Server or Client error classes.
+ */
+object GrpcStatusResolver : ApiStatusResolver {
```
ccjernigan, line 11: Should we make this `internal`?

ccjernigan (commented):
There's a lot of great stuff here!  The rework of the server responses is much cleaner, and allows us to better handle errors in the SDK. Thanks for the effort that went into that refactor and digging into the GRPC APIs.

I've added some comments, many of them date back to my original draft of this code.

After we do some testing to make sure we haven't destabilized anything, I think this is pretty close to move forward with merging.  There's plenty we can address in followup PRs since this refactoring was always intended to be multiple steps.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/CoroutineLightWalletClient.kt
```diff
@@ -0,0 +1,94 @@
...
+     */
+    suspend fun fetchTransaction(txId: ByteArray): Service.RawTransaction?
+
+    /**
+     * @param tAddress the transparent address to use.
+     * @param startHeight the starting height to use.
+     *
+     * @return the UTXOs for the given address from the [startHeight].
+     */
+    suspend fun fetchUtxos(
```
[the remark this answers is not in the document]
HonzaR: Fortunately, the GRPC layer provides us a stream version of the UTXOs fetch method as well, so I switched it. Thank you for pointing this out.

docs/Architecture.md
```diff
@@ -7,7 +7,36 @@ Thankfully, the only thing an app developer has to be concerned with is the foll
...
+      lightwalletClientLib[[lightwallet-client-lib]] --> sdkLib[[sdk-lib]];
+      sdkLib[[sdk-lib]] --> demoApp[[demo-app]];
+      sdkLib[[sdk-lib]] --> darksideTestLib[[darkside-test-lib]];
+```
+
+# Data model
+Before diving into some of the module specifics, it is helpful to provide some context on the data model representations as data flows between the different modules of this repository.  There are multiple data representations, including:
+
+1. Network — The wire representation for calls to and from the Lightwalletd server.  These are generated by `protoc` at compile time.  These are not generally a public API.
+2. Unsafe — The representation provided as the public API from `lightwallet-client-lib`.  These values are not necessarily validated, hence the smurf naming with the suffix `Unsafe`.  These are not generally a public API.
```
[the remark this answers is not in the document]
HonzaR: Yep, this is more clear. Thanks. I'm going to add javadoc comments to all of our unsafe model classes with explanation about it too.

.idea/codeStyles/Project.xml
```diff
@@ -8,8 +8,6 @@
           <package name="kotlinx.android.synthetic" alias="false" withSubpackages="true" />
         </value>
       </option>
-      <option name="NAME_COUNT_TO_USE_STAR_IMPORT" value="2147483647" />
```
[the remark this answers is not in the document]
HonzaR: Yep, reversed. Thanks.

lightwallet-client-lib/lint.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lint>
+    <!-- Generated by `./gradlew refreshVersions` to avoid errors when using _ as a version. -->
```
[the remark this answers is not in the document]
HonzaR: It seems like we are good without both lines in `lightwallet-client-lib/lint.xml`
```
<issue id="GradlePluginVersion" severity="ignore" />
<issue id="GradleDependency" severity="ignore" />
```
Should we remove the whole file, or just a part of it? Do we want to preserve the origin file in `sdk-lib/lint.xml` from which this was copied from?

lightwallet-client-lib/build.gradle.kts
```diff
@@ -0,0 +1,234 @@
...
+
+    buildTypes {
+        getByName("debug").apply {
+            // test builds exceed the dex limit because they pull in large test libraries
+            multiDexEnabled = true
+            isMinifyEnabled = false
+        }
+        getByName("release").apply {
+            multiDexEnabled = false
+            isMinifyEnabled = project.property("IS_MINIFY_SDK_ENABLED").toString().toBoolean()
```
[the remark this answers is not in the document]
HonzaR: I just did a first manual testing round of all of our use-cases with the release minified build, in which are the server calls involved and it seems alright.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/BlockingLightWalletClient.kt
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * @param tAddress the transparent address to use.
+     * @param startHeight the starting height to use.
+     *
+     * @return the UTXOs for the given address from the [startHeight].
+     */
+    fun fetchUtxos(
+        tAddress: String,
+        startHeight: BlockHeightUnsafe
+    ): List<Service.GetAddressUtxosReply>
```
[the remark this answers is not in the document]
HonzaR: Yep, done. It's better to have it unified with the other client calls.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/GrpcStatusResolver.kt
HonzaR: Good point. Moved.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/BlockingLightWalletClient.kt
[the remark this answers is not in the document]
HonzaR: Yep, just filed https://github.com/zcash/zcash-android-wallet-sdk/issues/895.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/ApiStatusResolver.kt
```diff
@@ -0,0 +1,10 @@
+package co.electriccoin.lightwallet.client
+
+import co.electriccoin.lightwallet.client.model.Response
+
+/**
+ * This class provides conversion from API statuses to our predefined Server or Client error classes.
+ */
+interface ApiStatusResolver {
```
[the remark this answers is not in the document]
HonzaR: Good point. Moved.

lightwallet-client-lib/lint.xml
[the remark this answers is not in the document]
ccjernigan: Let's remove the lines from both files, but keep the files in place (basically a file with an xml open and close tag but no rules in the middle).

That'll make it easier to add exceptions in the future if we need to.  Occasionally, new versions of AGP introduce lint bugs and having the files and gradle script in place to create new exceptions can be handy.

gradle.properties
HonzaR: Could you guys please check Alex's latest changes related to the proto files update, if they meet our expectations? I will need to make similar changes in this PR, so we don't discuss it twice. ([this commit](https://github.com/zcash/zcash-android-wallet-sdk/pull/804/commits/19c072c5592496ca07ca9d6983c832b4cb01099d) and [this change](https://github.com/zcash/zcash-android-wallet-sdk/pull/804/commits/83051bdedfca6490490e1b7746c5a6f44b048680#r1085327159))

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/BlockingLightWalletClientImpl.kt
```diff
@@ -0,0 +1,191 @@
...
+ * is created, it will use a deadline that is after the given duration from now.
+ * @property streamingRequestTimeout the timeout to use for streaming requests. When a new stub
+ * is created for streaming requests, it will use a deadline that is after the given duration from
+ * now.
+ */
+internal class BlockingLightWalletClientImpl private constructor(
+    private val channelFactory: ChannelFactory,
+    private val lightWalletEndpoint: LightWalletEndpoint,
+    private val singleRequestTimeout: Duration = 10.seconds,
+    private val streamingRequestTimeout: Duration = 90.seconds
```
pacu, line 38: Beware of timeouts being handled slightly different than one might expect where the call can be terminated because it hasn't finished its work in the duration specified regardless it is sending information back or not. This was a particular headache I suffered in the iOS version of GRPC

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/CoroutineLightWalletClientImpl.kt
```diff
@@ -0,0 +1,200 @@
...
+            lightWalletEndpoint: LightWalletEndpoint
+        ): CoroutineLightWalletClientImpl {
+            return CoroutineLightWalletClientImpl(channelFactory, lightWalletEndpoint)
+        }
+    }
+}
+
+private fun Channel.createStub(timeoutSec: Duration = 60.seconds) =
+    CompactTxStreamerGrpcKt.CompactTxStreamerCoroutineStub(this, CallOptions.DEFAULT)
+        .withDeadlineAfter(timeoutSec.inWholeSeconds, TimeUnit.SECONDS)
```
pacu, line 191: Please verify that this is the behavior you desire. I think that `withDeadlineAfter()` will shut down the channel after the given seconds regardless of it actually being transmitting in a very responsive communication.

pacu (changes requested):
Please confirm that the timeouts set up for the channel work as you would expect.

docs/Architecture.md
```diff
@@ -7,7 +7,36 @@ Thankfully, the only thing an app developer has to be concerned with is the foll
...
-# Components
+# Modules
+The SDK is broken down into several logical components, implemented as Gradle modules.  At a high level, the modularization is:
+
+ * sdk-lib — Compiles all of the modules together for the SDK.
+ * lightwallet-client-lib — Provides a set of Kotlin APIs for interacting with lightwalletd over the network.
+ * darkside-test-lib — Contains integration tests for the SDK, running against a localhost lightwalletd instance running in darkside mode.  This is not run as part of the SDK test suite, because it requires some manual setup to enable.
+ * demo-app — Contains a primitive demo application to exercise the SDK.
+
+```mermaid
```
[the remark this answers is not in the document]
ccjernigan: Can you say how?  This is the flow of dependencies so light-client-lib is consumed by sdk-lib.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/model/Response.kt
```diff
@@ -0,0 +1,127 @@
...
+        /**
+         * Use this function to convert Failure into Throwable object.
+         */
+        fun toThrowable() = Throwable("Communication failure with details: $code${description?.let{": $it"} ?: "."}")
+
+        /**
+         * The client was not able to communicate with the server.
+         */
+        class Connection<T>(
+            override val code: Int = CONNECTION_ERROR_CODE,
```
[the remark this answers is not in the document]
HonzaR: Yep, probably not. I'm removing the code parameter from these locally triggered failures with default codes then.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/block/CompactBlockDownloader.kt
```diff
@@ -77,55 +77,26 @@ open class CompactBlockDownloader private constructor(val compactBlockRepository
...
-            result = lightWalletService.getServerInfo()
-        } catch (e: StatusRuntimeException) {
-            retryUpTo(GET_SERVER_INFO_RETRIES) {
-                twig("WARNING: reconnecting to service in response to failure (retry #${it + 1}): $e")
-                lightWalletService.reconnect()
-                result = lightWalletService.getServerInfo()
+    suspend fun getServerInfo(): LightWalletEndpointInfoUnsafe? = withContext(IO) {
+        retryUpTo(GET_SERVER_INFO_RETRIES) {
+            when (val result = lightWalletClient.getServerInfo()) {
+                is Response.Success -> return@withContext result.result
```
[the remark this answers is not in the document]
HonzaR: Yep, agree.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/CoroutineLightWalletClientImpl.kt
HonzaR: Thanks Pacu, yes it will close it. So if I understand it right, you're saying that we should have this function as is in the blocking version of the client, but not in the coroutine one, where we should not use the timeout for the streaming endpoints? So it could look just like this:
`private fun Channel.createStub() = CompactTxStreamerGrpcKt.CompactTxStreamerCoroutineStub(this, CallOptions.DEFAULT)`

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/BlockingLightWalletClientImpl.kt
HonzaR: Yep, understand. The proposed separation of logic sounds good to me. Stronger types are the right long-term solution.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/CoroutineLightWalletClientImpl.kt
pacu: I hope I had an answer! 😂  I always found that the timeout api was not doing what I intended it to do. 

Just waving my hands to bring the attention to it 😅

lightwallet-client-lib/build.gradle.kts
```diff
@@ -0,0 +1,234 @@
...
+            initWith(buildTypes.getByName("release"))
+            matchingFallbacks += listOf("release")
+        }
+    }
+
+    sourceSets.getByName("main") {
+        proto { srcDir("src/main/proto") }
+    }
+
+    packagingOptions {
```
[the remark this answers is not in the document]
HonzaR: Thanks for pointing this out. Moved in a separate [commit](https://github.com/zcash/zcash-android-wallet-sdk/pull/675/commits/2e8d335ef35ee2d7dec290f28d8e3395c784d1c5). It includes `sdk-lib` and `lightwallet-client-lib` modules.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/BlockingLightWalletClientImpl.kt
```diff
@@ -0,0 +1,191 @@
...
+
+            Response.Success(lightwalletEndpointInfo)
+        } catch (e: StatusRuntimeException) {
+            GrpcStatusResolver.resolveFailureFromStatus(e)
+        }
+    }
+
+    override fun submitTransaction(spendTransaction: ByteArray): Response<SendResponseUnsafe> {
+        if (spendTransaction.isEmpty()) {
+            return Response.Failure.Client.SubmitEmptyTransaction()
```
ccjernigan, line 88: Long-term, is this and `fetchTransaction` below instances of where we should be throwing an Exception?

ccjernigan (commented):
I pushed a few very minor commits.

These changes look good to me, and bring a lot of nice enhancements to the SDK internals that we can leverage going forward.  I can't officially approve this PR though, since I created the original draft.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/BlockingLightWalletClientImpl.kt
```diff
@@ -126,7 +126,7 @@ internal class BlockingLightWalletClientImpl private constructor(
         }
 
         val result = requireChannel().createStub().getAddressUtxos(
-            Service.GetAddressUtxosArg.newBuilder().setAddress(tAddress)
+            Service.GetAddressUtxosArg.newBuilder().setAddresses(Constants.TRANSPARENT_ADDRESS_INDEX, tAddress)
```
pacu, line 130: I think just putting zero here is fine. If I'm not mistaken, the builder lets you request UTXOs for many addresses at once and the -somewhat misfortunate- builder API is asking of an index for the request. This index has nothing to do with the t-address index. In fact the `DiversifierIndex` of the t address you might be requesting will probably not match the index you need to build this query.
```suggestion
            Service.GetAddressUtxosArg.newBuilder().setAddresses(0, tAddress)
```

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/Constants.kt
```diff
@@ -3,4 +3,5 @@ package co.electriccoin.lightwallet.client.internal
 internal object Constants {
     const val LOG_TAG = "LightWalletClient" // NON-NLS
     const val ILLEGAL_ARGUMENT_EXCEPTION_MESSAGE = "Illegal argument provided - can't be empty:" // NON-NLS
+    const val TRANSPARENT_ADDRESS_INDEX = 0
```
pacu, line 6: see comment above.

ccjernigan:
> Please confirm that the timeouts set up for the channel work as you would expect.

These timeouts were in place previously.  We'll investigate and followup with a future PR to look into whether these should be adjusted.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/CoroutineLightWalletClientImpl.kt
```diff
@@ -130,7 +128,7 @@ internal class CoroutineLightWalletClientImpl private constructor(
             "${Constants.ILLEGAL_ARGUMENT_EXCEPTION_MESSAGE} address: $tAddress." // NON-NLS
         }
         return requireChannel().createStub().getAddressUtxosStream(
-            Service.GetAddressUtxosArg.newBuilder().setAddress(tAddress)
+            Service.GetAddressUtxosArg.newBuilder().setAddresses(Constants.TRANSPARENT_ADDRESS_INDEX, tAddress)
```
pacu, line 132: this is bringing up some questions on how the android sdk is requesting UTXOs. It has nothing to do with the API itself, but with the caller.

please double check that the SDK requests UTXO for all existing derived addresses. this means calling the rust API to get those addresses and call a function in this lightwalletd client that accepts many addresses as input (probably an array or sequence of them)

lightwallet-client-lib/src/main/proto/compact_formats.proto
```diff
@@ -1,7 +1,11 @@
-syntax = "proto3";
+// Copyright (c) 2019-2020 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+    syntax = "proto3";
 package cash.z.wallet.sdk.rpc;
 option java_package = "cash.z.wallet.sdk.internal.rpc";
-option go_package = "walletrpc";
+option go_package = "lightwalletd/walletrpc";
```
pacu, line 8: is this some change you introduced or are these proto files coming from lightwalletd?

lightwallet-client-lib/src/main/proto/compact_formats.proto
```diff
@@ -1,7 +1,11 @@
-syntax = "proto3";
+// Copyright (c) 2019-2020 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+    syntax = "proto3";
```
pacu, line 5: this indentation does not seem right

pacu (changes requested):
Great work. I spotted a few things that might need a look at! Please comment if you need further explanation

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/BlockingLightWalletClientImpl.kt
HonzaR: Yep, I just processed to right away. The exception throwing approach is better for us in this state of refactoring.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/BlockingLightWalletClientImpl.kt
HonzaR: Ok, understand. Thanks

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/CoroutineLightWalletClientImpl.kt
HonzaR: Thanks for pointing this out. I just quickly looked at it. If I understand it right, the UTXOs are refreshed e.g. after scan is complete [here](https://github.com/zcash/zcash-android-wallet-sdk/blob/01da937fdcf2d5d70686d5617ecad623f138f685/sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt#L405) and the [getTransparentAddress](https://github.com/zcash/zcash-android-wallet-sdk/blob/01da937fdcf2d5d70686d5617ecad623f138f685/sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt#L188) works with `Account.DEFAULT` for now. Moreover, until now we worked with the simple `getAddress` here in networking. So I assume we are checking the only one address, but I might we wrong. So do you want me to file an issue, so we can check it more in details and we're not blocked in this PR?

lightwallet-client-lib/src/main/proto/compact_formats.proto
HonzaR: Nope, I just downloaded the updated proto files from [lightwalletd repo](https://github.com/zcash/lightwalletd/tree/master/walletrpc) and added only one thing in each file and it's `java_package` option, which sets our local package for generated files.

lightwallet-client-lib/src/main/proto/compact_formats.proto
HonzaR: You have a good eye 🙂. Fixed.

HonzaR (approved):
I just self-reviewed again and approved, as Carter is co-author of this PR. Release build uploading to the maven works again after theirs last days infrastructure troubles, so we have this PR also robo tested with the release build.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/internal/CoroutineLightWalletClientImpl.kt
pacu: Yes this doesn't block this PR and can be worked in a subsequent PR but rather urgently if you ask me

HonzaR: Yep, here I just filed the issue for it https://github.com/zcash/zcash-android-wallet-sdk/issues/904. Please, add any additional details, if needed.

PR #717, opened 2022-09-21
Update Rust crates

str4d:
This migrates us to the latest revision of the `zcash/librustzcash` crates.

# Author

- [ ] Rebase and squash: Did you pull in the latest changes from the main branch and squash your commits before assigning a reviewer? _Having your code up to date and squashed will make it easier for others to review. Use best judgement when squashing commits, as some changes (such as refactoring) might be easier to review as a separate commit._

# Reviewer

- [x] Automated tests: Did you review the automated tests?
- [x] Run the app: Did you run the demo app and try the changes? _While the CI server runs the app to look for build failures or crashes, humans running the app are more likely to notice unexpected log messages, UI inconsistencies, or bad output data. Perform this step last, after verifying the code changes are safe to run locally._


sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
```diff
@@ -444,11 +444,14 @@ interface Synchronizer {
          * Synchronizer requires. It contains all information necessary to build a synchronizer and it is
          * mainly responsible for initializing the databases associated with this synchronizer and loading
          * the rust backend.
+         * @param seed the wallet's seed phrase. This only needs to be provided if this method returns an
+         * error indicating that the seed phrase is required for a database migration.
          */
         suspend fun new(
-            initializer: Initializer
+            initializer: Initializer,
+            seed: ByteArray? = null,
```
str4d, line 452: I do not know if this is the correct way to expose this; given that `seed` is not stored in `Initializer` currently, I figured it was better to not require it there. But I have no idea how these APIs are invoked. I _think_ this is the only actually-public API related to initialization.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/PersistentTransactionManager.kt
```diff
@@ -145,8 +145,10 @@ class PersistentTransactionManager(
         tx
     }
 
+    // TODO(str4d): This uses operator overloading to distinguish between it and createToAddress, which breaks when
+    //  spendingKey is removed. Figure out where these methods need to be renamed, and do so.
     override suspend fun encode(
```
str4d, line 150: I'm still figuring out how these APIs and their overloading work. We can't merge this PR until we fix this overloading, due to @nuttycom's desire to move all of these APIs to use the (intentionally unstable) binary USK encoding (which will make these two APIs have identical signatures).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -628,11 +628,19 @@ class SdkSynchronizer internal constructor(
 
     override suspend fun cancelSpend(pendingId: Long) = txManager.cancel(pendingId)
 
+    // TODO(str4d): Rename this to getCurrentAddress (and remove/add in changelog).
+    /**
+     * Returns the current Unified Address for this account.
+     */
```
str4d, line 634: I added this comment when I thought this was the public API that led to `RustWelding.getShieldedAddress`, but it does not. I am still figuring out how it is populated, and what the access patterns are.

str4d:
[Force-pushed](https://github.com/zcash/zcash-android-wallet-sdk/compare/fb506e7bb7a5c726ac51ee9b8e1475b2d8da61ca..868e7164f230bb011bf13c5faa2f2c59d01c26e4) to add a `SeedRequired` exception and pass the relevant status across the FFI to raise it.

str4d:
[Force-pushed](https://github.com/zcash/zcash-android-wallet-sdk/compare/fb506e7bb7a5c726ac51ee9b8e1475b2d8da61ca..12fd90af0f6ec0d2a3fe9ccf1076efc1f9408f50) to add a `SeedRequired` exception and pass the relevant status across the FFI to raise it.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
```diff
@@ -1016,18 +1017,40 @@ class CompactBlockProcessor internal constructor(
     suspend fun getLastScannedHeight() =
         repository.lastScannedHeight()
 
+    // TODO(str4d): CompactBlockProcessor is the wrong place for this, but it's where all the other APIs that need
+    //  access to the RustBackend live. This should be refactored.
+    internal suspend fun createAccount(seed: ByteArray): UnifiedSpendingKey =
```
pacu, line 1022: This is related to the layered architecture we've always wanted. It's probably a good discussion to have in terms of kotlin multiplatform

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/PersistentTransactionManager.kt
pacu: You should just use the USK because these APIs are really internal I don't think that maintaining the String based API will do any good

pacu (changes requested):
I did a first pass over the PR. Tests on my side tell me that it's safe to proceed with USK with total disregard of existing string based API. 

great work!

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -667,20 +690,20 @@ class SdkSynchronizer internal constructor(
...
         val tBalance = processor.getUtxoCacheBalance(tAddr)
-        val zAddr = getAddress(0)
+        val zAddr = getCurrentAddress(usk.account)
 
         // Emit the placeholder transaction, then switch to monitoring the database
-        txManager.initSpend(tBalance.available, zAddr, memo, 0).let { placeHolderTx ->
+        txManager.initSpend(tBalance.available, zAddr, memo, usk.account).let { placeHolderTx ->
             emit(placeHolderTx)
-            txManager.encode(spendingKey, transparentAccountPrivateKey, placeHolderTx).let { encodedTx ->
+            txManager.encode("", usk, placeHolderTx).let { encodedTx ->
```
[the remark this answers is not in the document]
str4d: Indeed, but the problem is that then we have two identical APIs named `encode` that don't encode anything, but actually do the important spend authorizing and transaction creating steps for spending and shielding. I don't yet know whether I can just rename them to be more accurate, or whether the overloaded name is depended on in some way.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/PersistentTransactionManager.kt
ccjernigan: PersistentTransactionManager is removed with Roomoval, so I wouldn't worry too much about it.  It will be removed from the public API entirely.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
ccjernigan: We shouldn't change addresses too much just yet; there are changes happening in the Roomoval PR to make these more type safe.

ccjernigan:
I added a followup issue to address how I think the Synchronization factory methods could evolve in the future: https://github.com/zcash/zcash-android-wallet-sdk/issues/725

ccjernigan (approved):

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
```diff
@@ -0,0 +1,16 @@
+package cash.z.ecc.android.sdk.model
+
+/**
+ * A [ZIP 316](https://zips.z.cash/zip-0316) account identifier.
+ *
+ * @param value A 0-based account index.  Must be >= 0.
+ */
+data class Account(val value: Int) {
+    init {
+        require(value >= 0) { "Account index must be >= 0 but actually is $value" }
```
str4d, line 10: An account must be between `0` and `2^31 - 1` inclusive (`2^31 - 1` is valid for derivation, but is used for legacy `zcashd` purposes, so we may want to exclude that as well for the purposes of the SDK).

ccjernigan: Kotlin Int is 32-bits.  Effectively it is a Java Integer.  So https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Integer.html#MAX_VALUE

sdk-lib/src/androidTest/java/cash/z/ecc/android/sdk/fixture/WalletFixture.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import cash.z.ecc.android.bip39.Mnemonics
+import cash.z.ecc.android.sdk.model.Account
+import cash.z.ecc.android.sdk.model.ZcashNetwork
+import cash.z.ecc.android.sdk.tool.DerivationTool
+
+object WalletFixture {
+    val NETWORK = ZcashNetwork.Mainnet
+    const val SEED_PHRASE =
+        "kitchen renew wide common vague fold vacuum tilt amazing pear square gossip jewel month tree shock scan alpha just spot fluid toilet view dinner"
```
str4d, line 11: In my review, I didn't spot where this was used (or a corresponding deletion of this seed phrase from a test). What motivated its addition?

CHANGELOG.md
```diff
@@ -27,15 +35,29 @@ Change Log
       - `Initializer.Config.newWallet`
       - `Initializer.Config.setViewingKeys`
 - `cash.z.ecc.android.sdk`:
-  - `Synchronizer.shieldFunds` now takes a transparent account private key (representing
-    all transparent secret keys within an account) instead of a transparent secret key.
+  - `Synchronizer.Companion.new` now takes an optional `seed` argument. This should be
+    provided if `Synchronizer.Companion.new` throws an error that a database migration
+    requires the wallet seed.
```
str4d, line 40: This was accurate as of the commit where I added the `seed` argument, but now that `Initializer` has been removed, it needs to be rewritten.

str4d (commented):
utACK f871896c654bfbbde9ea050247b7fa5f9b01c27c. My comments should be addressed before the feature branch is merged, but don't need to block this PR.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
str4d: Okay, so restricting to positive is equivalent to restricting to a 31-bit range, great. The `2^31 - 1` exclusion may want to happen, but that doesn't have to block this PR.

sdk-lib/src/androidTest/java/cash/z/ecc/android/sdk/fixture/WalletFixture.kt
ccjernigan: We need hard-coded fixture values to write some unit tests, and this is consistent with a pattern we are adopting in the Secant app.

These are used by the unit tests for DerivationTool and the UnifiedSpendingKey.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
ccjernigan: I filed a followup issue to come back to this: https://github.com/zcash/zcash-android-wallet-sdk/issues/729

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
```diff
@@ -472,6 +544,57 @@ interface Synchronizer {
...
+        ): Synchronizer = runBlocking {
+            new(context, zcashNetwork, alias, lightWalletEndpoint, seed, birthday)
+        }
+
+        /**
+         * Delete the databases associated with this wallet. This removes all compact blocks and
+         * data derived from those blocks. For most wallets, this should not result in a loss of
+         * funds because the seed and spending keys are stored separately. This call just removes
+         * the associated data but not the seed or spending key, themselves, because those are
+         * managed separately by the wallet.
```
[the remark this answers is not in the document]
ccjernigan: I updated the comment

CHANGELOG.md
```diff
@@ -27,15 +36,30 @@ Change Log
       - `Initializer.Config.newWallet`
       - `Initializer.Config.setViewingKeys`
 - `cash.z.ecc.android.sdk`:
-  - `Synchronizer.shieldFunds` now takes a transparent account private key (representing
-    all transparent secret keys within an account) instead of a transparent secret key.
+  - `Synchronizer.Companion.new` now takes a `seed` argument. A non-null value should be
+    provided if `Synchronizer.Companion.new` throws an error that a database migration
+    requires the wallet seed.
```
str4d, line 41: This rewrite is insufficient. This constructor's API was completely altered due to the removal of `Initializer`, and the caller now needs to provide up-front several variables that previously they might have been ignoring and leaving to the `Initializer` defaults. The change should be described here (but it doesn't need to include a migration pathway; that can go in the migrations doc later).

str4d (commented):
utACK 600e90de6f7574b10400f92a2d6b3f1dbcc4018b modulo changelog comment (I reviewed https://github.com/zcash/zcash-android-wallet-sdk/compare/f871896c654bfbbde9ea050247b7fa5f9b01c27c..600e90de6f7574b10400f92a2d6b3f1dbcc4018b).

CHANGELOG.md
ccjernigan: Good comments.  I will resolve this later, as the changelog will have a lot more changes with the Roomoval.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
```diff
@@ -437,26 +466,68 @@ interface Synchronizer {
...
+        @Suppress("LongParameterList")
         suspend fun new(
-            initializer: Initializer
+            context: Context,
+            zcashNetwork: ZcashNetwork,
+            alias: String = "zcash",
+            lightWalletEndpoint: LightWalletEndpoint,
+            seed: ByteArray?,
+            birthday: BlockHeight?
         ): Synchronizer {
```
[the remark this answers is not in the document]
pacu: I think the way Initializer is implemented in Android didn't do much. The way I ended up using it was to actually construct other objects and reduce the object graph. On iOS you can pass an Initializer to a CompactBlockProcessor or Synchronizer to get the repositories, etc. Now that I think of the Android SDK didn't do that so this change makes sense in this context.

PR #770, opened 2022-11-04
More FFI updates

str4d:
This pulls in the recent bugfixes to the Rust crates, removes FFI methods we don't need, and updates Rust dependencies.

# Author

# Reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,6 @@ Change Log
 - `cash.z.ecc.android.sdk.tool`:
   - `DerivationTool.deriveUnifiedSpendingKey`
   - `DerivationTool.deriveUnifiedFullViewingKey`
-  - `DerivationTool.deriveTransparentAccountPrivateKey`
-  - `DerivationTool.deriveTransparentAddressFromAccountPrivateKey`
   - `DerivationTool.deriveUnifiedAddress`
   - `DerivationTool.deriveUnifiedFullViewingKeys`
   - `DerivationTool.validateUnifiedFullViewingKey`
@@ -60,18 +58,13 @@ Change Log
     public key, and not the extended public key as intended. This made it incompatible
     with ZIP 316.
 - `cash.z.ecc.android.sdk.tool`:
-  - `DerivationTool.deriveSpendingKeys` (use
-    `DerivationTool.deriveUnifiedSpendingKey` instead).
-  - `DerivationTool.deriveViewingKey` (use
-  - `DerivationTool.deriveUnifiedFullViewingKey` instead).
-  - `DerivationTool.deriveTransparentAddressFromPrivateKey` (use
-    `DerivationTool.deriveTransparentAddressFromAccountPrivateKey` instead).
-  - `DerivationTool.deriveTransparentSecretKey` (use
-    `DerivationTool.deriveTransparentAccountPrivateKey` instead).
+  - `DerivationTool.deriveSpendingKeys` (use `DerivationTool.deriveUnifiedSpendingKey` instead).
+  - `DerivationTool.deriveViewingKey` (use `DerivationTool.deriveUnifiedFullViewingKey` instead).
+  - `DerivationTool.deriveTransparentAddress` (use `Synchronizer.getLegacyTransparentAddress` instead).
+  - `DerivationTool.deriveTransparentAddressFromPrivateKey` (use `Synchronizer.getLegacyTransparentAddress` instead).
+  - `DerivationTool.deriveTransparentAddressFromPublicKey` (use `Synchronizer.getLegacyTransparentAddress` instead).
+  - `DerivationTool.deriveTransparentSecretKey` (use `DerivationTool.deriveUnifiedSpendingKey` instead).
   - `DerivationTool.deriveShieldedAddress`
-    - TODO: Do we still need to be able to derive Sapling shielded addresses for legacy
-      support? Currently removed because `UnifiedFullViewingKey` doesn't expose the
-      Sapling FVK on the Kotlin side (unlike the previous `UnifiedViewingKey`).
   - `DerivationTool.deriveUnifiedViewingKeys`
   - `DerivationTool.validateUnifiedViewingKey`
 
--- a/sdk-lib/Cargo.toml
+++ b/sdk-lib/Cargo.toml
@@ -15,7 +15,7 @@ failure = "0.1"
 hdwallet = "0.3.1"
 hdwallet-bitcoin = "0.3"
 hex = "0.4"
-jni = { version = "0.19", default-features = false }
+jni = { version = "0.20", default-features = false }
 log = "0.4"
 log-panics = "2.0.0"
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -64,6 +64,8 @@ Change Log
     `DerivationTool.deriveUnifiedSpendingKey` instead).
   - `DerivationTool.deriveViewingKey` (use
   - `DerivationTool.deriveUnifiedFullViewingKey` instead).
+  - `DerivationTool.deriveTransparentAddress` (use
+    `Synchronizer.getLegacyTransparentAddress` instead).
```
str4d, line 68: Should we remove any of the other transparent address derivation methods? This one was the most confusing, and also happens to be the only one that wasn't altered during UA integration. The other old ones removed below were because we replaced them with UA-compatible equivalents, but those still allow users to generate arbitrary taddrs that don't necessarily appear in the UAs the SDK will generate for accounts (if the UAs contain Sapling, which they do currently).

sdk-lib/Cargo.toml
```diff
@@ -10,33 +10,32 @@ publish = false
 edition = "2018"
 
 [dependencies]
-android_logger = "0.9"
+android_logger = "0.11"
```
ccjernigan, line 13: This was causing build issues so I reverted it a while back.
```
error: type parameters must be declared prior to const parameters
   --> /Users/carter/.cargo/registry/src/github.com-1ecc6299db9ec823/android_logger-0.11.1/src/lib.rs:437:33
    |
437 | fn uninit_array<const N: usize, T>() -> [MaybeUninit<T>; N] {
    |                -----------------^- help: reorder the parameters: lifetimes, then types, then consts: `<T, const N: usize>`

error: could not compile `android_logger` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
```

ccjernigan (changes requested):
CI is picking up on a compile failure in the test module.  You should be able to reproduce locally with the assembleAndroidTest task.

CHANGELOG.md
ccjernigan: It seems like we eventually want to refactor this to simplify it.  Whether that time is now is later is ¯\_(ツ)_/¯ 

Given that the DerivationTool has `deriveUnifiedAddress`, would it make more sense for the remaining methods to take the UnifiedAddress as a parameter and derive the others from it?

Would that line of thought lead us towards having a UnifiedAddress object with accessor methods for getSaplingAddress and getTransparentAddress?

sdk-lib/Cargo.toml
str4d: I think that's due to an older Rust version; I recall the ordering being relaxed at some point. That means this dependency has an MSRV we need to take into account. But that's easy enough: we should always use the latest stable Rust version.

CHANGELOG.md
str4d: We already have those methods on `RustBackend`: `getTransparentReceiver` and `getSaplingReceiver`. If those would make more sense moved to `DerivationTool`, we can do that.

str4d: Re: the others, the question is whether there remains e.g. a use case for deriving a transparent address from a transparent account private key. At the time I ported the existing methods to be compatible with ZIP 316, we weren't planning to pass around an encoded USK. Now that we are, it's likely that these methods simply aren't necessary now, so we could just remove them.

sdk-lib/Cargo.toml
str4d: Aha, it was Rust 1.59 that enabled this: https://blog.rust-lang.org/2022/02/24/Rust-1.59.0.html#const-generics-defaults-and-interleaving

We can make this easy to avoid by setting our MSRV to 1.59 in `Cargo.toml`; then the compiler will return a more useful error if it is too old.

CHANGELOG.md
ccjernigan: I vote remove and simplify the public API.

pacu: > Re: the others, the question is whether there remains e.g. a use case for deriving a transparent address from a transparent account private key.

no use for this since on the SDK this derivation was a workaround and transparent priv keys were never stored just passed through the ffi

str4d: Okay, I will do this.

sdk-lib/src/androidTest/java/cash/z/ecc/android/sdk/util/TestWallet.kt
```diff
@@ -79,9 +77,9 @@ class TestWallet(
 
     val available get() = synchronizer.saplingBalances.value?.available
     val unifiedAddress =
-        runBlocking { DerivationTool.deriveUnifiedAddress(seed, network = network, Account.DEFAULT) }
+        runBlocking { synchronizer.getCurrentAddress(Account.DEFAULT) }
     val transparentAddress =
-        runBlocking { DerivationTool.deriveTransparentAddress(seed, network = network, Account.DEFAULT) }
+        runBlocking { synchronizer.getLegacyTransparentAddress(Account.DEFAULT) }
```
pacu, line 82: You shouldn't use the legacy account. We don't use it neither expose it anymore. 

Both SDKs should be consistent

pacu (changes requested):

sdk-lib/src/androidTest/java/cash/z/ecc/android/sdk/util/TestWallet.kt
str4d: That is not what this API does. It returns the transparent receiver from the current UA for the given account. "Legacy" here just means "you should be using UAs primarily in your UI, but here's a taddr if you need it".

pacu: This is misleading. Transparent addresses are not being called legacy anywhere else that I know of. We are referring to  components of UAs as receivers. Mixing deprecation intentions with these APIs is weird

str4d: "Receiver" is a term of art that I thought would be more confusing in the mobile SDKs as it is a new term, and given that these APIs were intended to be direct swaps in for existing "show transparent address" usage introduced by the auto shielding feature, this name would be more obvious. But if you and @Carter agree that it should be exposed to your users as receivers, and your migration documentation will handle the onboarding, then it doesn't bother me.

ccjernigan: I think our best option is to have a strongly typed UnifiedAddress object that allows fetching the various components from it.  However I'd rather avoid further changing the API at this stage so that we can get this SDK update released.  We have enough churn right now as-is.

To track the future API change we can repurpose this existing issue, although the description needs to change a bit since it was written a while ago: https://github.com/zcash/zcash-android-wallet-sdk/issues/483

PR #775, opened 2022-11-12
Migrate to published Zcash Rust crates with ZIP 317 support

str4d:
ZIP 317 fee rules are disabled by default but exposed via a boolean flag.

# Author

# Reviewer

### DIFF
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackend.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackend.kt
@@ -228,7 +228,7 @@ internal class RustBackend private constructor(
             File(saplingParamDir, SaplingParamTool.SPEND_PARAM_FILE_NAME).absolutePath,
             File(saplingParamDir, SaplingParamTool.OUTPUT_PARAM_FILE_NAME).absolutePath,
             networkId = network.id,
-            useZip317Fees = false
+            useZip317Fees = IS_USE_ZIP_317_FEES
         )
     }
 
@@ -245,7 +245,7 @@ internal class RustBackend private constructor(
                 File(saplingParamDir, SaplingParamTool.SPEND_PARAM_FILE_NAME).absolutePath,
                 File(saplingParamDir, SaplingParamTool.OUTPUT_PARAM_FILE_NAME).absolutePath,
                 networkId = network.id,
-                useZip317Fees = false
+                useZip317Fees = IS_USE_ZIP_317_FEES
             )
         }
     }
@@ -327,6 +327,7 @@ internal class RustBackend private constructor(
      */
     companion object {
         internal val rustLibraryLoader = NativeLibraryLoader("zcashwalletsdk")
+        private const val IS_USE_ZIP_317_FEES = false
 
         /**
          * Loads the library and initializes path variables. Although it is best to only call this
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackend.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackend.kt
@@ -227,7 +227,8 @@ internal class RustBackend private constructor(
             memo ?: ByteArray(0),
             File(saplingParamDir, SaplingParamTool.SPEND_PARAM_FILE_NAME).absolutePath,
             File(saplingParamDir, SaplingParamTool.OUTPUT_PARAM_FILE_NAME).absolutePath,
-            networkId = network.id
+            networkId = network.id,
+            useZip317Fees = false
         )
     }
 
@@ -243,7 +244,8 @@ internal class RustBackend private constructor(
                 memo ?: ByteArray(0),
                 File(saplingParamDir, SaplingParamTool.SPEND_PARAM_FILE_NAME).absolutePath,
                 File(saplingParamDir, SaplingParamTool.OUTPUT_PARAM_FILE_NAME).absolutePath,
-                networkId = network.id
+                networkId = network.id,
+                useZip317Fees = false
             )
         }
     }
@@ -486,7 +488,8 @@ internal class RustBackend private constructor(
             memo: ByteArray,
             spendParamsPath: String,
             outputParamsPath: String,
-            networkId: Int
+            networkId: Int,
+            useZip317Fees: Boolean
         ): Long
 
         @JvmStatic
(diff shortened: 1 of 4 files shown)


sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/WalletTransactionEncoder.kt
```diff
@@ -157,7 +162,8 @@ internal class WalletTransactionEncoder(
                 twig("params exist! attempting to shield...")
                 rustBackend.shieldToAddress(
                     usk,
-                    memo
+                    memo,
+                    useZip317Fees
                 )
             } catch (t: Throwable) {
                 // TODO [#680]: if this error matches: Insufficient balance (have 0, need 1000 including fee)
```
daira, line 169:
```suggestion
                // TODO [#680]: if this error matches: Insufficient balance (have 0, need * including fee)
```

sdk-lib/src/main/rust/lib.rs
```diff
@@ -1091,18 +1126,46 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_createToAdd
...
+                    &network,
+                    prover,
+                    &input_selector,
+                    &usk,
+                    request,
+                    OvkPolicy::Sender,
+                    ANCHOR_OFFSET,
+                )
+                .map_err(|e| format_err!("Error while creating transaction: {}", e))
+            },
```
daira, line 1167: This is an unfortunate duplication. I'm not sure how to fix that given that the types are different.

daira (approved):
utACK with minor comments.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/WalletTransactionEncoder.kt
str4d: This part of the code is not altered by this PR anymore, and the Android SDK will stick with a fixed fee for the next release.

sdk-lib/src/main/rust/lib.rs
```diff
@@ -733,17 +751,21 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_getTotalTra
         let addr = utils::java_string_to_rust(&env, address);
         let taddr = TransparentAddress::decode(&network, &addr).unwrap();
 
+        // We select all transparent funds including unmined coins, as that is the same
+        // set of UTXOs we shield from.
+        let min_confirmations = 0;
```
ccjernigan, line 756: What kind of implications does this have for clients of the SDK?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackend.kt
```diff
@@ -227,7 +227,8 @@ internal class RustBackend private constructor(
             memo ?: ByteArray(0),
             File(saplingParamDir, SaplingParamTool.SPEND_PARAM_FILE_NAME).absolutePath,
             File(saplingParamDir, SaplingParamTool.OUTPUT_PARAM_FILE_NAME).absolutePath,
-            networkId = network.id
+            networkId = network.id,
+            useZip317Fees = false
```
ccjernigan, line 231: I suggest we eliminate the default argument here, define a constant like below, and pass the constant in explicitly where this is called.

I often dislike default arguments because they are a frequent source of subtle bugs.

```
companion object {
    private const val IS_USE_ZIP_317_FEES = false
}
```

ccjernigan (commented):
I have an old review comment that GitHub won't show me, so you'll see something on a line that changed but I can't edit until I submit this review 🤦‍♂️.

Besides that, I have a question about the

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/WalletTransactionEncoder.kt
```diff
@@ -121,7 +123,8 @@ internal class WalletTransactionEncoder(
         usk: UnifiedSpendingKey,
         amount: Zatoshi,
         toAddress: String,
-        memo: ByteArray? = byteArrayOf()
+        memo: ByteArray? = byteArrayOf(),
+        useZip317Fees: Boolean = false
```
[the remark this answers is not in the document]
ccjernigan: This is the stale comment, but the reasoning about why I like to avoid default arguments is still valid and applies to my other comment about default arguments.

sdk-lib/src/main/rust/lib.rs
str4d: None. This change is a no-op; it just uses the `WalletRead` API correctly.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackend.kt
str4d: This isn't a default argument? This is a function call of the `createToAddress` FFI method, passing the explicit value `false`. It only uses keyword argument styling because that's what was already done for `networkId`, and I'm bad at Kotlin and don't know how positional and keyword arguments interact. (An earlier version of the PR did have default arguments elsewhere higher up in the stack, but all those were removed when you requested that SDK users not be allowed to configure this.) Happy to use a constant here though.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/WalletTransactionEncoder.kt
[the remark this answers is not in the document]
str4d: I agree with you in general. However, in all of those stale and now-removed changes, I was following the existing style (as I do everywhere in the Kotlin codebase, because I am not a Kotlin developer and am mostly fumbling in the dark here). If the memo argument did not have a default, I would not have added a default for the ZIP 317 fees flag, so that indicates the memo argument's default should definitely be removed from these internals.

ccjernigan (approved):

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/WalletTransactionEncoder.kt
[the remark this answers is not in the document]
ccjernigan: Yes this is a place where the existing style should get revised.  We're slowly working towards that! 😄

PR #905, opened 2023-02-01
List transparent receivers

### LINKED ISSUE #904: Check addresses for loading UTXOs
### Describe the goal
We'd like to analyse our SDK UTXOs loading mechanisms. The main question is if we use all derived existing addresses for requesting an account UTXOs?

### Any extra information that might be useful in the debugging process.
This thought comes from our Networking refactoring PR https://github.com/zcash/zcash-android-wallet-sdk/pull/675 reviews, e.g. from this [discussion](https://github.com/zcash/zcash-android-wallet-sdk/pull/675#discussion_r1091053812).

pacu:
(eventually when it's done by a real Android developer) Closes #904

This Code is a Draft. I've ported the Rust code from Zcash-light-client-ffi repo to match JNIs rust FFI code so that should work.

The kotlin code is just a draft of how the fetchUtxos should be done instead. This will provide every derived transparent address for a specific account and fetch the corresponding UTXOs. if the caller is not rolling UAs, then the SDK will always request UTXOs for the legacy transparent address and the transparent receiver for the UA of diversifier index 0 for the account. 

# Author

- [ ] Manual tests: Did you update the [manual tests](../blob/main/docs/testing/manual_testing) as appropriate? _While we aim for automated testing of the application, some aspects require manual testing. If you had to manually test something during development of this pull request, write those steps down._

# Reviewer

- [x] Run the app: Did you run the demo app and try the changes? _While the CI server runs the app to look for build failures or crashes, humans running the app are more likely to notice unexpected log messages, UI inconsistencies, or bad output data. Perform this step last, after verifying the code changes are safe to run locally._

### DIFF
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
@@ -380,7 +380,7 @@ class SdkSynchronizer private constructor(
 
     suspend fun refreshUtxos() {
         Twig.debug { "refreshing utxos" }
-        refreshUtxos(getTransparentAddress())
+        refreshUtxos(Account.DEFAULT)
     }
 
     /**
@@ -672,8 +672,8 @@ class SdkSynchronizer private constructor(
         }
     }
 
-    override suspend fun refreshUtxos(tAddr: String, since: BlockHeight): Int? {
-        return processor.refreshUtxos(tAddr, since)
+    override suspend fun refreshUtxos(account: Account, since: BlockHeight): Int? {
+        return processor.refreshUtxos(account, since)
     }
 
     override suspend fun getTransparentBalance(tAddr: String): WalletBalance {
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
@@ -378,11 +378,6 @@ class SdkSynchronizer private constructor(
     // Private API
     //
 
-    suspend fun refreshUtxos() {
-        Twig.debug { "refreshing utxos" }
-        refreshUtxos(Account.DEFAULT)
-    }
-
     /**
      * Calculate the latest balance, based on the blocks that have been scanned and transmit this
      * information into the flow of [balances].
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
@@ -276,12 +276,15 @@ interface Synchronizer {
     suspend fun validateAddress(address: String): AddressType
 
     /**
-     * Download all UTXOs for the given address and store any new ones in the database.
+     * Download all UTXOs for the given account addresses and store any new ones in the database.
      *
-     * @return the number of utxos that were downloaded and addded to the UTXO table.
+     * @param account The Account, for which all addresses blocks will be downloaded.
+     * @param since The BlockHeight, from which blocks will be downloaded.
+     *
+     * @return the number of utxos that were downloaded and added to the UTXO table.
      */
     suspend fun refreshUtxos(
-        tAddr: String,
+        account: Account = Account.DEFAULT,
         since: BlockHeight = network.saplingActivationHeight
     ): Int?
 
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
@@ -3,14 +3,11 @@ package cash.z.ecc.android.sdk.block
 import androidx.annotation.VisibleForTesting
 import cash.z.ecc.android.sdk.BuildConfig
 import cash.z.ecc.android.sdk.annotation.OpenForTesting
-import cash.z.ecc.android.sdk.block.CompactBlockProcessor.State.Disconnected
 import cash.z.ecc.android.sdk.block.CompactBlockProcessor.State.Downloading
 import cash.z.ecc.android.sdk.block.CompactBlockProcessor.State.Enhancing
-import cash.z.ecc.android.sdk.block.CompactBlockProcessor.State.Initialized
 import cash.z.ecc.android.sdk.block.CompactBlockProcessor.State.Scanned
 import cash.z.ecc.android.sdk.block.CompactBlockProcessor.State.Scanning
 import cash.z.ecc.android.sdk.block.CompactBlockProcessor.State.Stopped
-import cash.z.ecc.android.sdk.block.CompactBlockProcessor.State.Validating
 import cash.z.ecc.android.sdk.exception.CompactBlockProcessorException
 import cash.z.ecc.android.sdk.exception.CompactBlockProcessorException.EnhanceTransactionError.EnhanceTxDecryptError
 import cash.z.ecc.android.sdk.exception.CompactBlockProcessorException.EnhanceTransactionError.EnhanceTxDownloadError
@@ -130,7 +127,7 @@ class CompactBlockProcessor internal constructor(
         )
     )
 
-    private val _state: MutableStateFlow<State> = MutableStateFlow(Initialized)
+    private val _state: MutableStateFlow<State> = MutableStateFlow(State.Initialized)
     private val _progress = MutableStateFlow(0)
     private val _processorInfo = MutableStateFlow(ProcessorInfo(null, null, null, null, null))
     private val _networkHeight = MutableStateFlow<BlockHeight?>(null)
@@ -218,9 +215,8 @@ class CompactBlockProcessor internal constructor(
                         delay(napTime)
                     }
                     BlockProcessingResult.NoBlocksToProcess, BlockProcessingResult.FailedEnhance -> {
-                        val noWorkDone =
-                            currentInfo.lastDownloadRange?.isEmpty() ?: true &&
-                                currentInfo.lastScanRange?.isEmpty() ?: true
+                        val noWorkDone = currentInfo.lastDownloadRange?.isEmpty() ?: true &&
+                            currentInfo.lastScanRange?.isEmpty() ?: true
                         val summary = if (noWorkDone) {
                             "Nothing to process: no new blocks to download or scan"
                         } else {
@@ -286,11 +282,11 @@ class CompactBlockProcessor internal constructor(
 
         if (!updateRanges()) {
             Twig.debug { "Disconnection detected! Attempting to reconnect!" }
-            setState(Disconnected)
+            setState(State.Disconnected)
             downloader.lightWalletClient.reconnect()
             BlockProcessingResult.Reconnecting
         } else if (currentInfo.lastDownloadRange.isEmpty() && currentInfo.lastScanRange.isEmpty()) {
-            setState(Scanned(currentInfo.lastScanRange))
+            setState(State.Scanned(currentInfo.lastScanRange))
             BlockProcessingResult.NoBlocksToProcess
         } else {
             if (BenchmarkingExt.isBenchmarking()) {
@@ -407,7 +403,7 @@ class CompactBlockProcessor internal constructor(
      */
     private suspend fun validateAndScanNewBlocks(lastScanRange: ClosedRange<BlockHeight>?): BlockProcessingResult =
         withContext(IO) {
-            setState(Validating)
+            setState(State.Validating)
             val result = validateNewBlocks(lastScanRange)
             if (result == BlockProcessingResult.Success) {
                 // in theory, a scan should not fail after validation succeeds but maybe consider
@@ -452,8 +448,8 @@ class CompactBlockProcessor internal constructor(
             BlockProcessingResult.FailedEnhance
         }
     }
-    // TODO [#683]: we still need a way to identify those transactions that failed to be enhanced
-    // TODO [#683]: https://github.com/zcash/zcash-android-wallet-sdk/issues/683
+// TODO [#683]: we still need a way to identify those transactions that failed to be enhanced
+// TODO [#683]: https://github.com/zcash/zcash-android-wallet-sdk/issues/683
 
     private suspend fun enhance(transaction: TransactionOverview) = withContext(Dispatchers.IO) {
         transaction.minedHeight?.let { minedHeight ->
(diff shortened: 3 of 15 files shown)


HonzaR (commented):
I just pushed a few mostly minor and documentation changes. On top of that there were `Synchronizer` and `RustBackend` jvm parts missing. I also left one more comment in the code, which I find good to point out now. Anyway, thank you Pacu for this PR. It's great to have this covered.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -386,7 +386,7 @@ class SdkSynchronizer private constructor(
 
     suspend fun refreshUtxos() {
```
ccjernigan, line 387: Should we eliminate this method and just keep the one with the explicit account argument?  I think we're trending towards being explicit about requiring the account argument, although it isn't 100% consistent everywhere.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
```diff
@@ -574,11 +574,14 @@ class CompactBlockProcessor internal constructor(
                 @Suppress("TooGenericExceptionCaught")
                 try {
                     retryUpTo(3) {
+                        val tAddresses = rustBackend.listTransparentReceivers(account.value)
```
ccjernigan, line 577: I remember that recently we had changes in how the transparent addresses were derived, as part of supporting unified addresses.

Does this method get both the old and the new transparent address?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackend.kt
```diff
@@ -112,6 +112,16 @@ internal class RustBackend private constructor(
 
     override fun getSaplingReceiver(ua: String) = getSaplingReceiverForUnifiedAddress(ua)
 
+    override suspend fun listTransparentReceivers(account: Int): Array<String> {
```
ccjernigan, line 115: Should we make the account argument here be the type safe `Account` object?

It would be nice to have the type safe objects for the public API to the RustBackend (but note that we cannot use them for the `external` functions since those are for the JNI)

sdk-lib/src/main/rust/lib.rs
```diff
@@ -1316,3 +1316,74 @@ fn parse_network(value: u32) -> Result<Network, failure::Error> {
...
+/// - `db_data` must be non-null and valid for reads for `db_data_len` bytes, and it must have an
+///   alignment of `1`. Its contents must be a string representing a valid system path in the
+///   operating system's preferred representation.
+/// - The memory referenced by `db_data` must not be mutated for the duration of the function call.
+/// - The total size `db_data_len` must be no larger than `isize::MAX`. See the safety
+///   documentation of pointer::offset.
+/// - Call [`zcashlc_free_keys`] to free the memory associated with the returned pointer
+///   when done using it.
+#[no_mangle]
+pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_listTransparentReceivers(
```
ccjernigan, line 1335: @str4d Would you also be able to give this JNI method a quick review?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
```diff
@@ -618,8 +619,9 @@ class CompactBlockProcessor internal constructor(
             twig("Found UTXO at height ${utxo.height.toInt()} with ${utxo.valueZat} zatoshi")
             @Suppress("TooGenericExceptionCaught")
             try {
+
                 rustBackend.putUtxo(
```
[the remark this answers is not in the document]
ccjernigan: Great that you noticed this.  I think checking the return value is important, and we should at a minimum log the failure.

Failure writing to the Rust backend should probably be consistent (it's writing to the SQLite database which isn't expected to be flaky as far as I know).  Will a retry improve reliability in this particular case?  I suspect the retries are important for the network downloads, but maybe not the database writes.

@str4d Would a failure writing to the rust backend here succeed on a second try?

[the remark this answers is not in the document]
HonzaR: I'm also thinking about why do we return `true/false` as well as throwing error in the rust layer? Could this be unified, or is it a necessary pattern here? 

Or may be another question, is the Rust really throwing an exception and we should `try-catch` it? Because I can see that the Rust does not throw exceptions as we know them from the Android world, it just returns `Result<T, E>` or use `panic`, which is probably the reason why we have to have the `try-catch` here, right? Thank you @str4d.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
HonzaR: Yes, good idea.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackend.kt
HonzaR: Yes, I like this idea, so I went further and changed all related RustBacked methods, which used Int type to Account type. The changes were pretty straightforward, see related commit.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
HonzaR: What I can see is that to get a single transparent address, which we used before these changes -  `getTransparentReceiverForUnifiedAddress()`, we got a transparent address for a unified address from the Rust layer. Now we use `listTransparentReceivers()` to get all transparent addresses from the Rust layer, which implementation seems different to me, but I might be wrong. So please @str4d, could you also take a quick look at this thought? Thank you.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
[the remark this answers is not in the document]
ccjernigan: Thank you for this explanation.

A quick solution might be: tweak many of the RustBackend public APIs to have void return values (leaving the `external fun` declarations along with the FFI code alone for a second refactor).  We'd need to add a comment in RustBackend explaining why the booleans are being ignored.  Then it would be contained within a single file.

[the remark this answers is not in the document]
str4d: > Would a failure writing to the rust backend here succeed on a second try?

It would depend on the failure. At a glance, the kinds of errors that can occur from this function include:
- Passing an unknown network ID.
- Wallet DB path not existing or not being openable.
- SQLite errors of basically any kind, including violations of table constraints (which would likely re-occur on retry) and transient write failures (which might not).
- Attempting to store a UTXO that is not P2PKH or P2SH.

The non-SQLite errors occurring would indicate logic bugs on the Android side, so assuming those don't occur... the answer is "maybe"? For example, if the failure occurs because the UTXO already existed in the wallet DB (failing the `prevout` uniqueness constraint), due to e.g. a race condition on the Android side, that will continue to fail.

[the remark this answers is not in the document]
HonzaR: Thank you @str4d for this information. We have two more things for you, see below. Could you please take a quick look at them too, once you have some spare time so we can have this merged? Thank you.

I removed the new boolean check of this method and filed a new issue for changing related RustBackend method return types to void, as discussed here. https://github.com/zcash/zcash-android-wallet-sdk/issues/920.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
nuttycom: I agree that it would be a good idea to just remove this method. Unresolving this conversation so that this can be addressed.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
```diff
@@ -277,12 +277,15 @@ interface Synchronizer {
...
      *
-     * @return the number of utxos that were downloaded and addded to the UTXO table.
+     * @param account The Account, for which all addresses blocks will be downloaded.
+     * @param since The BlockHeight, from which blocks will be downloaded.
+     *
+     * @return the number of utxos that were downloaded and added to the UTXO table.
      */
     suspend fun refreshUtxos(
-        tAddr: String,
+        account: Account = Account.DEFAULT,
```
nuttycom, line 288: I'm skeptical of default arguments; they're almost always a bad idea and lead users down unpleasant paths.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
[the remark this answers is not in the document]
nuttycom: Should #920 block this PR from merging then? Otherwise this should still check the boolean result until the alternative approach is implemented.

nuttycom (changes requested):
A couple of minor changes requested: please remove the defaults for the account argument in new & updated APIs, and there's a small utility method that can be removed as it duplicates existing functionality in librustzcash.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
ccjernigan: I generally agree with no default arguments (except in our test fixtures and occasionally in Compose UI method signatures)

In this particular case, I think we should keep the account default argument until the SDK actually supports accounts.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/jni/RustBackendWelding.kt
```diff
@@ -54,21 +54,23 @@ internal interface RustBackendWelding {
 
     fun isValidUnifiedAddr(addr: String): Boolean
 
-    suspend fun getCurrentAddress(account: Int = 0): String
+    suspend fun getCurrentAddress(account: Account = Account.DEFAULT): String
```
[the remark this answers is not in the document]
ccjernigan: We could remove the default from here, because this is an internal API of the SDK and isn't public facing.  This is a good example where default arguments could cause problems.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
nuttycom: The problem here is that you have multiple methods that all provide a default for their argument, which is dangerous: all call sites have to choose to either supply an argument, or to use the default. I'd much prefer that the callers be modified to be called with `Account.DEFAULT` than to leave this in place - it will prove a land mine when adding account support to the SDK.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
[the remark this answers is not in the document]
ccjernigan: If desired, we could add a comment noting issue 920 and that the return value is not useful in this context.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
ccjernigan: I agree and recommend this as a [followup issue](https://github.com/zcash/zcash-android-wallet-sdk/issues/939), so that we can get this PR merged which will fix an important bug in the SDK.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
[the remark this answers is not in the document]
HonzaR: Yep, comments added to have this clear for everyone. We're good without boolean checking here, as we previously already had the try-catch result checking approach. So the https://github.com/zcash/zcash-android-wallet-sdk/issues/920 is not blocking this PR.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
HonzaR: Yes, I agree. Thank you for pointing this out.

sdk-lib/src/main/rust/lib.rs
```diff
@@ -1316,3 +1316,74 @@ fn parse_network(value: u32) -> Result<Network, failure::Error> {
...
+                        |env| env.new_string(""),
+                    ))
+                },
+            Err(e) => Err(format_err!("Error while fetching address: {}", e)),
+        }
+    });
+    unwrap_exc_or(&env, res, ptr::null_mut())
+}
+
+fn convert_network_into_zcash_address_network(zcash_network: zcash_primitives::consensus::Network) -> zcash_address::Network {
```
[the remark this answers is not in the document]
HonzaR: Thanks for pointing us there. I just made the [change](https://github.com/zcash/zcash-android-wallet-sdk/pull/905/commits/044df67103fe86e74de89010be5221e60f4f0c81) in the Rust layer, but I'm very new to Rust programming, so please, could you guys take a look at it and suggest a change. Thank you.

HonzaR (approved):

sdk-lib/src/main/rust/lib.rs
```diff
@@ -1341,7 +1341,8 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_jni_RustBackend_listTranspa
 ) -> jobjectArray {
     let res = panic::catch_unwind(|| {
         let network = parse_network(network_id as u32)?;
-        let zcash_network = convert_network_into_zcash_address_network(network);
+        // Valid value of the Option<Network> is guaranteed by the previous network_id parsing
+        let zcash_network = Network::address_network(&network);
```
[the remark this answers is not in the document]
nuttycom: Alternately, if you're confident that a failure here to convert is a programming error that should crash the wallet process, you could implement it as:
```suggestion
        let zcash_network = network.address_network().expect("network_id parsing should not have resulted in an unrecognized network type.");
```

nuttycom (approved):
utACK, thanks for addressing the comments!

ccjernigan (approved):

PR #931, opened 2023-03-02
[#674] Use suspending functions for networking

HonzaR:
- Transformed from blocking to the coroutine version of the networking client (the blocking one was removed entirely)
- Temporarily commented out some features of the old UI demo-app to avoid necessary changes in it. It will be solved separately at https://github.com/zcash/zcash-android-wallet-sdk/issues/973
- Left a few FIXMEs also in tests which were disabled
- Running our benchmark locally still works as expected
- Note that fetching UTXOs still suffers with https://github.com/zcash/zcash-android-wallet-sdk/issues/941
- Closes https://github.com/zcash/zcash-android-wallet-sdk/issues/674

# Author

- [ ] Manual tests: Did you update the [manual tests](../blob/main/docs/testing/manual_testing) as appropriate? _While we aim for automated testing of the application, some aspects require manual testing. If you had to manually test something during development of this pull request, write those steps down._
- [ ] Documentation: Did you update documentation as appropriate? (e.g [README.md](../blob/main/README.md), etc.)

# Reviewer


sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/storage/block/FileCompactBlockRepository.kt
```diff
@@ -18,51 +18,61 @@ import cash.z.ecc.android.sdk.jni.findBlockMetadata
...
-            val tmpFile = block.createTemporaryFile(blocksDirectory)
-            // write compact block bytes
-            tmpFile.writeBytesSuspend(block.toByteArray())
-            // buffer metadata
-            metaDataBuffer.add(block.toJniMetaData())
-
-            val isFinalizeSuccessful = tmpFile.finalizeFile()
-            check(isFinalizeSuccessful) {
-                "Failed to finalize file: ${tmpFile.absolutePath}"
+    override suspend fun write(blocks: Flow<CompactBlockUnsafe>): Flow<Int> {
```
ccjernigan, line 41: If we're returning a flow, I don't think the `suspend` is necessary anymore.  At the same time, I do have some concern about returning a flow in this case.  The implementation is correct, but it has potential to be a sharp corner in the future.

Since the flow has side effects in the database and filesystem (it doesn't just emit a final value), collecting the flow multiple times would cause those side effects to occur multiple times.  We don't often invoke a flow for their side effects, but rather for their final value which is where it could be a surprise for a future person working on this code.

Would it make more sense to keep the suspend modifier for the function, do the collection internally (e.g. make the write() method perform side effects) rather than defer the side effects to when the Flow<Int> is collected?

ccjernigan (commented):
This is a great improvement and will simplify a lot of our networking!  Thanks for the effort that's gone into this.

I pushed a couple of minor fixes.

I have one question inline regarding the return value of the CompactBlockRepository, and where we want the side-effects to occur.

Issue #239, opened 2021-08-20
Get rid of the PendingTransaction Database


gmale:
As a wallet developer (or SDK maintainer), I would like one database to contain the wallet information, rather than two so that there are less moving parts and reduced complexity.

Critieria:
* the number of databases is reduced from 3 to a number less than 3
* transactions that failed to submit due to intermittent wifi can be automatically resubmitted on behalf of the user by the wallet until they expire
* the SDK is able to write data to support tracking failed transactions without interfering with wallet scanning or otherwise corrupting the wallet database

Implementation Details
* Of course, we can't just delete the database or even just move it's tables to the Data.DB
    * If we simply move its tables we'll violate the read/write separation and begin generating corrupted database issues because the Kotlin code will attempt to modify the database while the rust code is scanning
    * So first, we ought to [make the database transactions more robust](https://github.com/zcash/zcash-android-wallet-sdk/issues/240) so that corruption cannot occur
* The Pending Transaction DB also addresses a few concerns that are not currently supported
    * Mempool support
        * The wallet certainly should know about every transaction that it has sent
    * Retry support
        * If a transaction fails to encode, then there is no need to retry. If it fails to submit, retry could happen automatically
        * Transactions can safely be retried until they expire but doing so requires maintaining data on disk
            * Note: cutting the block time in half also reduced the retry window by half
            * This smaller window for retries might make failures something we don't need to persist outside of app memory
        * When a transaction failed, the reason should be persisted

## Background
Originally, the [SDK was designed with an intention to keep writes and reads separate](https://github.com/zcash/zcash-android-wallet-sdk/tree/296782286b473c309764e6f26ac0bf6711880302#structure): the "kotlin code" would only ever write compact blocks and read decrypted wallet data produced by the "rust code." Conversely, the "rust code" would only ever write decrypted wallet data and read compact blocks produced by the "kotlin code."
[...]

PR #990, opened 2023-04-17
[#239] Remove pending transaction database

ccjernigan:
I added one followup issue to return an ID after a transaction is submitted, so that clients could implement their own tracking of pending transactions if desired.  See https://github.com/zcash/zcash-android-wallet-sdk/issues/991

> **Note**

# Author

# Reviewer

- [x] Review the **automated tests**


sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/DatabaseCoordinator.kt
```diff
@@ -186,6 +183,28 @@ internal class DatabaseCoordinator private constructor(context: Context) {
...
+     * @return true only if any database deleted, false otherwise
+     */
+    internal suspend fun deletePendingTransactionDatabase(
+        network: ZcashNetwork,
+        alias: String
+    ): Boolean {
+        deleteFileMutex.withLock {
+            val dataDeleted = deleteDatabase(pendingTransactionsDbFile(network, alias))
+
+            val cacheDeleted = fsBlockDbRoot(network, alias).deleteRecursivelySuspend()
```
ccjernigan, line 202: I will fix this; we don't want to delete the blocks

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/OutboundTransactionManagerImpl.kt
```diff
@@ -0,0 +1,88 @@
...
+    override suspend fun isValidUnifiedAddress(address: String) =
+        encoder.isValidUnifiedAddress(address)
+
+    //
+    // Helper functions
+    //
+
+    companion object {
+        /** Error code for an error while encoding a transaction */
+        const val ERROR_ENCODING = 2000
```
ccjernigan, line 75: We aren't using these anymore—I will delete them.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -611,58 +545,42 @@ class SdkSynchronizer private constructor(
...
         }
     }
 
-    override fun shieldFunds(
+    @Throws(TransactionEncoderException::class)
+    override suspend fun shieldFunds(
         usk: UnifiedSpendingKey,
         memo: String
-    ): Flow<PendingTransaction> {
+    ) {
```
pacu, line 570: We should return a transaction data type for wallets to show or look up is needed.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/OutboundTransactionManager.kt
```diff
@@ -0,0 +1,83 @@
...
+    suspend fun isValidTransparentAddress(address: String): Boolean
+
+    /**
+     * Return true when the given address is a valid ZIP 316 Unified Address.
+     *
+     * @param address the address to validate.
+     *
+     * @return true when the given address is a valid ZIP 316 Unified Address.
+     */
+    suspend fun isValidUnifiedAddress(address: String): Boolean
```
pacu, line 72: Why are this functions here? That's curious

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/TransactionOverview.kt
```diff
@@ -5,6 +5,9 @@ package cash.z.ecc.android.sdk.model
  *
  * Note that both sent and received transactions will have a positive net value.  Consumers of this class must check
  * [isSentTransaction] if displaying negative values is desired.
+ *
+ * Pending transactions are identified by a null [minedHeight].  Pending transactions are considered expired if the
+ * last synced block exceeds the [expiryHeight].
```
pacu, line 10: We probably want a `isPending` helper

pacu (changes requested):
Left some comments on specific details. My main objection regarding this change is that even thought the pendingDb might not be needed anymore, the concept of "pending transactions" is not going away in any of it two forms.

A pending transaction can be a transaction that hasn't yet been mined or that it is pending to be confirmed given a number of `min_confirmations` that we've hardcoded at 10 in the rust welding layer.

I don't see how this PR addresses the pendingDb on those roles.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/TransactionOverview.kt
ccjernigan: I avoided adding isPending here because I think it needs to know the latest scanned height, not the latest height of the blockchain.

Although I could add a method with a parameter for lastScannedHeight, I have a concern that if scanning becomes out-of-order, it isn't clear that clients would be able to pass in the right argument to determine this.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/OutboundTransactionManager.kt
ccjernigan: Agreed, although I'm avoiding moving them for a separate issue which will place them on a type safe Address object.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/TransactionOverview.kt
ccjernigan: If we invert the problem, I wonder if an `isConfirmed` flag is better than `isPending`?

We still lack the information here to report confirmation in this object as it is, but perhaps that could be injected by the synchronizer doing a map on the flow that gets returned.

pacu: We probably want to disambiguate what Pending and Unconfirmed mean to us. They previously were meshed together and this is a great opportunity to do so.

We already have a `expired_unmined` column. So we can call `Pending` to transactions that have not been mined and have not expired either. 

Then `Unconfirmed` is a little bit more complicated because we need an "anchor" value to say N transactions have passed since this transaction was mined. Information that the "v_transaction" view does not bring with it.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/OutboundTransactionManagerImpl.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    override suspend fun isValidUnifiedAddress(address: String) =
+        encoder.isValidUnifiedAddress(address)
+
+    //
+    // Helper functions
+    //
+
+    companion object {
+
```
pacu, line 74: dangling space

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -197,8 +188,11 @@ class SdkSynchronizer private constructor(
     // Transactions
     //
 
-    override val clearedTransactions get() = storage.allTransactions
-    override val pendingTransactions = txManager.getAll()
+    override val transactions
+        get() = combine(processor.networkHeight, storage.allTransactions) { networkHeight, allTransactions ->
```
pacu, line 192: I think that, given the immutable nature of the blockchain, it's safe to use the latest height of the chain instead of the latest scanned height to settle confirmed, pending and expired. 

Also, a performance concern: how does this combination of stream work? 

When you scan you will be progressing in the last known scanned height with every scanned batch, would that cause a re-combination of these two streams every time one of them emits a new value?

pacu (changes requested):
I think the overall idea is good! I got some more doubts regarding the stream combination and usage of scanned height instead of latest chain height

pacu (approved):

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/OutboundTransactionManagerImpl.kt
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    override suspend fun submit(encodedTransaction: EncodedTransaction) {
+        when (val response = service.submitTransaction(encodedTransaction.raw.byteArray)) {
+            is Response.Success -> {
+                Twig.debug { "SUCCESS: submit transaction completed with response: ${response.result}" }
+            }
+
+            is Response.Failure -> {
```
HonzaR, line 50: Should we also report the submission failure result to the caller, or is it done another way?

HonzaR (commented):
It's a nice simplification of the Synchronizer. Thank you. I just left two question comments inline. Other than that our CI unit testing reports some dangling tests. And while testing the changes with demo-app manually it reports `NullPointerException: null cannot be cast to non-null type kotlin.Int` in `Synchronizer.toWalletSnapshot()`, which causes the demo-app unable to start.

.idea/codeStyles/Project.xml
```diff
@@ -8,8 +8,6 @@
           <package name="kotlinx.android.synthetic" alias="false" withSubpackages="true" />
         </value>
       </option>
-      <option name="NAME_COUNT_TO_USE_STAR_IMPORT" value="2147483647" />
```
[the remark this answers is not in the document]
ccjernigan: Thanks for noticing this.  Android Studio keeps deleting these lines without my permission.  I reverted this which should fix it.

HonzaR (commented):
I just re-reviewed and found out that the app is crashing when submitting a transaction with insufficient balance, i.e. we do not catch its balance exception now.
```
RuntimeException: Error while creating transaction: Insufficient balance (have 0, need 100001000 including fee)
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
```diff
@@ -175,29 +161,28 @@ interface Synchronizer {
...
         usk: UnifiedSpendingKey,
         amount: Zatoshi,
         toAddress: String,
         memo: String = ""
-    ): Flow<PendingTransaction>
+    ): Long
 
-    fun shieldFunds(
+    /**
+     * @return The transaction id of the sent transaction.
```
pacu, line 179: I'd clarify this is the internal id and not the "txid" on chain to make it really obvious, because this number can change if the chain re-orgs so it shouldn't be stored long term

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/TransactionOverview.kt
```diff
@@ -21,7 +26,62 @@ data class TransactionOverview internal constructor(
...
+}
+
+enum class TransactionState {
+    Confirmed,
+    Pending,
+    Expired;
+
+    companion object {
+
+        private const val MIN_CONFIRMATIONS = 10
```
pacu, line 72: I think this is already a constant called defaultStaleTolerance or so.

pacu (changes requested):
Left some small remarks to improve docs and other things

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/TransactionOverview.kt
ccjernigan: The only thing I can find is ZcashSdk.EXPIRY_OFFSET which is about how many blocks before a transaction expires.

pacu (changes requested):
this PR can't be merged until we fix an issue in the `v_transactions` View where pending transactions with outgoing funds shows the value of the notes spent instead without factoring in the change that will be incoming so the row for a specific transaction has a higher value that the one the user has input in the transaction form or payment request completion

MIGRATIONS.md
```diff
@@ -3,7 +3,11 @@ Troubleshooting Migrations
 
 Migration to Version 1.17
 ---------------------------------
-Synchronizer APIs for listing sent and received transactions have been removed.  Clients should use `Synchronizer.clearedTransactions`, filtering on the field `TransactionOverview.isSentTransaction`.
+Synchronizer APIs for listing sent and received transactions have been removed.  Clients should use `Synchronizer.transactions`, filtering on the field `TransactionOverview.isSentTransaction`.
+
+Synchronizer APIs for pending transactions have been removed.  Clients should use `Synchronizer.transactions`, filtering on the field `TransactionOverview.isMined`
```
pacu, line 8: I don't think this is correct since pending transactions can be mined but not yet confirmed.

pacu (changes requested):
small doc change!

ccjernigan:
> this PR can't be merged until we fix an issue in the `v_transactions` View where pending transactions with outgoing funds shows the value of the notes spent instead without factoring in the change that will be incoming so the row for a specific transaction has a higher value that the one the user has input in the transaction form or payment request completion

This has been rebased on the changes to adopt the updated database views.

MIGRATIONS.md
ccjernigan: This has been updated

pacu (approved):

PR #1017, opened 2023-05-02
Migrate to `zcash_client_sqlite 0.7`

str4d:


sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/derived/SaplingReceivedNoteTable.kt
```diff
@@ -5,7 +5,7 @@ import cash.z.ecc.android.sdk.internal.db.queryAndMap
 import cash.z.ecc.android.sdk.model.ZcashNetwork
 import java.util.Locale
 
-internal class ReceivedNoteTable(
+internal class SaplingReceivedNoteTable(
```
nuttycom, line 8: Does it still make sense to use the `sapling_received_notes` table directly? The `tx_outputs` view is potentially a better source of this data.

nuttycom (approved):
utACK with question

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/derived/TxOutputsView.kt
```diff
@@ -17,47 +17,48 @@ internal class SentNoteTable(
...
         )
 
-        private val SELECT_BY_TRANSACTION_ID = String.format(
+        private val SELECT_BY_TRANSACTION_ID_AND_NOT_CHANGE = String.format(
             Locale.ROOT,
-            "%s = ?", // $NON-NLS
-            SentNoteTableDefinition.COLUMN_INTEGER_TRANSACTION_ID
+            "%s = ? AND %s == 0", // $NON-NLS
+            TxOutputsViewDefinition.COLUMN_INTEGER_TRANSACTION_ID,
+            TxOutputsViewDefinition.COLUMN_INTEGER_IS_CHANGE
```
nuttycom, line 34: I *think* this is generally okay, since shielding transactions should treat all the shielded value as "change".

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -324,18 +324,16 @@ class SdkSynchronizer private constructor(
...
-                val sentNoteIds = storage.getSentNoteIds(transactionOverview.id)
-
-                sentNoteIds.map { rustBackend.getSentMemoAsUtf8(it) }.filterNotNull()
-            }
-            false -> {
-                val receivedNoteIds = storage.getReceivedNoteIds(transactionOverview.id)
-
-                receivedNoteIds.map { rustBackend.getReceivedMemoAsUtf8(it) }.filterNotNull()
+        return storage.getNoteIds(transactionOverview.id).map {
+            when (transactionOverview.isSentTransaction) {
```
nuttycom, line 328: I wonder if the `get*MemoAsUtf8` methods really make sense now, given `v_tx_outputs`? Should we just be getting the memos from there?

nuttycom (approved):
utACK. I think in the future that we'll want to change how the data access API exposes memos, but for now this looks good.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
ccjernigan: Eventually yes, that would be more efficient because we could do a single database query.  We don't have any Kotlin implementation for dealing with memo bytes so this is the path of least implementation.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/derived/SaplingReceivedNoteTable.kt
pacu: Yes, it is but the changes are not small so that can be adopted later as long as the SDK is cherrypicking data from them as it is currently now.

Issue #1094, opened 2023-06-22
Consider fake SDK sync related components


HonzaR:
## Is your feature request related to a problem? Please describe.
In our tests as well as in the Compose Previews we'd appreciate having a way how to mock our key synchronizer components.

## Describe the solution you'd like
It requires mocking a few components like networking client (https://github.com/zcash/zcash-android-wallet-sdk/issues/898), Synchornizer, DerivedDataRepository, CompactBlockProcessor (CompactBlockDownloader & CompactBlockRepository), or TransactionManager.

The rust backend is already mocked as `FakeRustBacked`.

Issue #776, opened 2022-11-14
Add Support for ZIP 317 Fees in Public API


ccjernigan:
## Is your feature request related to a problem? Please describe.
ZIP 317 adds variable fees.  The SDK has been mostly implemented assuming fixed fees.  The underlying librustzcash supports ZIP 317 but work needs to be done in the public API to make variable fees better supported.

## Describe the solution you'd like
 - Refactor the current fixed fee constant in the SDK so that clients know it isn't always going to be the fee amount
 - Update the demo app to support ZIP 317 fees
 - Documentation: https://github.com/zcash/zips/blob/main/zip-0317.rst

str4d:
Once the SDK's API is updated as appropriate, the final step is to change this constant to `true`:
https://github.com/zcash/zcash-android-wallet-sdk/blob/1a0508f57dfa6eec68668f4705990f59f62ff3c5/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt#L362

PR #1116, opened 2023-07-11
Fast spendability

str4d:
Depends on https://github.com/zcash/librustzcash/pull/872.

# Author

- [x] Check the **code coverage**[^3] report for the automated tests

# Reviewer


sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/CompactBlockProcessor.kt
```diff
@@ -754,27 +753,12 @@ class CompactBlockProcessor internal constructor(
                         // Enrich batch model with fetched blocks. It's useful for later blocks deletion
                         downloadStageResult.batch.blocks = downloadStageResult.stageResult.downloadedBlocks
 
-                        // Run validation stage
+                        // Run scanning stage (which also validates the fetched blocks)
```
HonzaR, line 756: Great improvement for the scanBlocks function.

HonzaR (commented):
Thanks for making these changes! I just made a first walk through the changes and tested them. It seems good to me. I left only one minor question on the `Cargo.toml` update.

backend-lib/Cargo.toml
```diff
@@ -39,6 +40,16 @@ tracing-subscriber = "0.3"
...
 
+[patch.crates-io]
+incrementalmerkletree = { git = "https://github.com/zcash/incrementalmerkletree.git", rev = "67111e29403c33f2e36d6924167f1d5f04ad0fc2" }
+shardtree = { git = "https://github.com/zcash/incrementalmerkletree.git", rev = "67111e29403c33f2e36d6924167f1d5f04ad0fc2" }
+orchard = { git = "https://github.com/zcash/orchard.git", rev = "6ef89d5f154de2cf7b7dd87edb8d8c49158beebb" }
+zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "352e1c709ac4c95074812dcabeba3cd5565371d2" }
+zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "352e1c709ac4c95074812dcabeba3cd5565371d2" }
+zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "352e1c709ac4c95074812dcabeba3cd5565371d2" }
+zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "352e1c709ac4c95074812dcabeba3cd5565371d2" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "352e1c709ac4c95074812dcabeba3cd5565371d2" }
```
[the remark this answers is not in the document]
str4d: Yes, but only if the targeted revision is guaranteed to be stable. This revision comes from an open PR, and while I'm trying to keep it stable, there's no guarantee yet.

Separately, we do not guarantee stability of database migrations for pinned revisions, so if we do merge this PR as-is then you'd be blocked from making an SDK release until you move to the released Rust crates, and any wallet you test with that version of the SDK may need to have its database wiped and be recovered from seed.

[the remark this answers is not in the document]
HonzaR: Ok, understand. Thanks for clarifying. Then consider this as a feature branch and build next sub tasks upon this one. Once done, we can merge this all together.

backend-lib/src/main/rust/lib.rs
```diff
@@ -726,27 +657,20 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_ge
...
-            .map(|amount| amount.into())
+        if let Some(wallet_summary) = db_data
+            .get_wallet_summary(0)
+            .map_err(|e| format_err!("Error while fetching balance: {}", e))?
+        {
+            wallet_summary
+                .account_balances()
+                .get(&account)
+                .ok_or_else(|| format_err!("Unknown account"))
+                .map(|balances| Amount::from(balances.sapling_balance.total()).into())
```
nuttycom, line 668: Shouldn't this be `balance.total()` if it's intended for use as the top-line value? Or is the meaning of this API really `getShieldedBalance`?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeBackendImpl.kt
```diff
@@ -22,32 +23,20 @@ internal class TypesafeBackendImpl(private val backend: Backend) : TypesafeBacke
...
-        return DerivationTool.getInstance().deriveUnifiedFullViewingKeys(seed, network, numberOfAccounts)
-    }
-
-    override suspend fun initBlocksTable(checkpoint: Checkpoint) {
-        backend.initBlocksTable(
-            checkpoint.height.value,
-            checkpoint.hash,
-            checkpoint.epochSeconds,
-            checkpoint.tree
+        checkpoint: Checkpoint,
```
nuttycom, line 28: Given that the only thing that you do with the `Checkpoint` is extract the tree state from it, I feel like the caller could be responsible for invoking the `checkpoint.treeState().encoded` dereference and this would then be more flexible; you don't need an actual checkpoint to initialize the wallet, though if you have a checkpoint you can certainly use that. We don't initialize the blocks table with a block hash from the checkpoint any longer, so we no longer verify that downloaded blocks chain from a checkpointed hash. I don't think that's a problem though; even if we did chain from a `Checkpoint` object, it would be possible to construct a malicious `Checkpoint` to chain from.

nuttycom (approved):
utACK f88d7ff2ae613a6835ea04aa2b27eaa92c3a0bc1 with a minor question; this review applies exclusively to f88d7ff2ae613a6835ea04aa2b27eaa92c3a0bc1 as I have not reviewed the other commits on this branch.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/derived/DerivedDataDb.kt
```diff
@@ -47,27 +47,15 @@ internal class DerivedDataDb private constructor(
...
-                // TODO [#681]: consider converting these to typed exceptions in the welding layer
-                // TODO [#681]: https://github.com/zcash/zcash-android-wallet-sdk/issues/681
-                tryWarn(
-                    message = "Did not initialize the blocks table. It probably was already initialized.",
-                    ifContains = "table is not empty"
-                ) {
-                    backend.initBlocksTable(checkpoint)
+            // If a seed is provided, fill in the accounts.
+            seed?.let {
+                for (i in 1..numberOfAccounts) {
```
str4d, line 56: TODO: This will blindly create new accounts even if old ones exist. The `1` here needs to be replaced with the number of existing accounts. I'm not sure how the SDK accesses this currently. cc @HonzaR.

backend-lib/src/main/rust/lib.rs
str4d: I thought this API was just the shielded balance (i.e. the balance that can be spent to recipients in our wallet), given that there are existing separate `getTotalTransparentBalance` and `getVerifiedTransparentBalance` APIs.

nuttycom: Okay, yeah, as far as maintaining the current API goes I agree that this is the correct change. In the future, if we keep this API and don't replace it entirely with a wallet summary API, it should probably be renamed `getShieldedBalance` or something of the sort.

HonzaR (approved):
[d7cc472](https://github.com/zcash/zcash-android-wallet-sdk/pull/1116/commits/d7cc472535a7a5f0dddd3dcd4bd38ddbfb7aac5d) looks good to me. Thank you.

str4d:
[Force-pushed](https://github.com/zcash/zcash-android-wallet-sdk/compare/2c7679b57bc45ca4243917b9ed1880b96cc61bba..0b6480486592ae69e9f7331e5b8438fa93cc179b) to add https://github.com/zcash/librustzcash/pull/953 to the Rust bugfix commit.

HonzaR (approved):
Selfe-reviewed. There are these bigger follow-ups:
- #1224 (Refactor and re-enable disabled darkside tests), which is our long-term task; the darkside test module has been abandoned for a long time, but we are prioritizing it now.
- #1094 (Consider fake SDK sync-related components),  which enables us to prepare tests for the CompactBlockProcessor's functions
- And a few more are mentioned in the codebase

nuttycom (commented):
Incomplete review, flushing comments.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -183,31 +206,108 @@ class CompactBlockProcessor internal constructor(
         // Clear any undeleted left over block files from previous sync attempts
         deleteAllBlockFiles(
             downloader = downloader,
-            lastKnownHeight = getLastScannedHeight(repository)
+            lastKnownHeight = when (val result = getMaxScannedHeight(backend)) {
+                is GetMaxScannedHeightResult.Success -> result.height
+                else -> null
+            }
```
nuttycom, line 212: This looks incorrect to me; given out-of-order sync, it could delete block files that have been downloaded but not yet scanned because the max-scanned height may be near the tip of the chain, but un-scanned files can exist below this height. Use `getFullyScannedHeight` instead.

nuttycom (changes requested):
Potential bug.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
nuttycom: Oh, actually, I didn't realize that `lastKnownHeight` is just used for an error message. Never mind.

nuttycom (approved):
utACK 9284e60e2aef950c2c17f79d17affc7bbd631108

backend-lib/src/main/rust/lib.rs
```diff
@@ -797,9 +727,7 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_ge
         let addr = utils::java_string_to_rust(&env, address);
         let taddr = TransparentAddress::decode(&network, &addr).unwrap();
 
-        // We select all transparent funds including unmined coins, as that is the same
-        // set of UTXOs we shield from.
-        let min_confirmations = 0;
+        let min_confirmations = NonZeroU32::new(1).unwrap();
```
str4d, line 730: There's a bug here: in the [iOS SDK](https://github.com/zcash-hackworks/zcash-light-client-ffi/blob/055b239fdd4218cdc9d7cbf354689f50891dc726/rust/src/lib.rs#L1339-L1346), I use `1` here and then use the `target` height instead of the `anchor` height to ensure I'm actually getting the zero-conf balance. We should make that fix here for now (so this is consistent with the iOS SDK), and then in a subsequent refactor we can migrate this API to `get_wallet_summary`.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1379,17 +1468,16 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_sh
...
-        let db_data = wallet_db(&env, network, db_data)?;
-        let mut db_data = db_data.get_update_ops()?;
+        let mut db_data = wallet_db(&env, network, db_data)?;
         let usk = decode_usk(&env, usk)?;
         let memo_bytes = env.convert_byte_array(memo).unwrap();
         let spend_params = utils::java_string_to_rust(&env, spend_params);
         let output_params = utils::java_string_to_rust(&env, output_params);
 
-        let min_confirmations = 0;
+        let min_confirmations = NonZeroU32::new(1).unwrap();
```
str4d, line 1480: Same bug: in the iOS SDK I use `target` instead of `anchor` to re-enable zero-conf.

Really, what we should do is revert the `NonZeroU32` change for the `shield_transparent_funds` API (or implement https://github.com/zcash/librustzcash/issues/873). That doesn't block the RC.

str4d (commented):
Reviewed everything as of 9284e60e2aef950c2c17f79d17affc7bbd631108 except `CompactBlockProcessor.kt`. Changes requested (I can't mark that status as I opened this PR).

nuttycom (changes requested):
Requesting changes for @str4d: https://github.com/zcash/zcash-android-wallet-sdk/pull/1116#pullrequestreview-1618200035

CHANGELOG.md
```diff
@@ -1,8 +1,40 @@
...
+  which contains all blocks in case of `SpendBeforeSync` synchronization algorithm. No internal change was made in 
+  case of older Linear synchronization algorithm.
+- `CompactBlockProcessor.ProcessorInfo.isSyncing`. Use `Synchronizer.status` instead.
+- `CompactBlockProcessor.ProcessorInfo.syncProgress`. Use `Synchronizer.progress` instead.
+- `alsoClearBlockCache` parameter from rewind functions of `Synchronizer` and `CompactBlockProcessor` as it take no 
+  effect on the current rewind functionality result.
+- Internally, we removed access to the shared block table from the Kotlin layer, which resulted in eliminating these 
+  APIs:
+  - `SdkSynchornizer.findBlockHash()`
+  - `SdkSynchornizer.findBlockHashAsHex()`
```
daira, line 21:
```suggestion
  - `SdkSynchronizer.findBlockHash()`
  - `SdkSynchronizer.findBlockHashAsHex()`
```

CHANGELOG.md
```diff
@@ -1,8 +1,40 @@
...
+`CompactBlockProcessor.SYNC_BATCH_SIZE`. Issue **#1206**.
+
+### Removed
+- `CompactBlockProcessor.ProcessorInfo.lastSyncHeight` which the SDK is no longer able to provide because of the new
+  **SpendBeforeSync** synchronization algorithm adoption. Use `CompactBlockProcessor.ProcessorInfo.overallSyncRange` 
+  which contains all blocks in case of `SpendBeforeSync` synchronization algorithm. No internal change was made in 
+  case of older Linear synchronization algorithm.
+- `CompactBlockProcessor.ProcessorInfo.isSyncing`. Use `Synchronizer.status` instead.
+- `CompactBlockProcessor.ProcessorInfo.syncProgress`. Use `Synchronizer.progress` instead.
+- `alsoClearBlockCache` parameter from rewind functions of `Synchronizer` and `CompactBlockProcessor` as it take no
```
daira, line 16:
```suggestion
- `alsoClearBlockCache` parameter from rewind functions of `Synchronizer` and `CompactBlockProcessor`, as it has no 
```

CHANGELOG.md
```diff
@@ -1,8 +1,40 @@
...
+
+### Removed
+- `CompactBlockProcessor.ProcessorInfo.lastSyncHeight` which the SDK is no longer able to provide because of the new
+  **SpendBeforeSync** synchronization algorithm adoption. Use `CompactBlockProcessor.ProcessorInfo.overallSyncRange` 
+  which contains all blocks in case of `SpendBeforeSync` synchronization algorithm. No internal change was made in 
+  case of older Linear synchronization algorithm.
+- `CompactBlockProcessor.ProcessorInfo.isSyncing`. Use `Synchronizer.status` instead.
+- `CompactBlockProcessor.ProcessorInfo.syncProgress`. Use `Synchronizer.progress` instead.
+- `alsoClearBlockCache` parameter from rewind functions of `Synchronizer` and `CompactBlockProcessor` as it take no 
+  effect on the current rewind functionality result.
```
daira, line 17:
```suggestion
- `alsoClearBlockCache` parameter from rewind functions of `Synchronizer` and `CompactBlockProcessor`, as it has no 
  effect on the current behaviour of these functions.
```

CHANGELOG.md
```diff
@@ -1,8 +1,40 @@
...
+### Changed
+- `CompactBlockProcessor.quickRewind()` and `CompactBlockProcessor.rewindToNearestHeight()` now might fail due to
+  internal changes in getting scanned height. Thus, these functions return `Boolean` results.
+- `Synchronizer.new()` requires a new `walletInitMode` parameter of type `WalletInitMode`, which describes wallet
+  initialization mode. See related function and sealed class documentation.
+
+### Fixed
+- `Synchronizer.getMemos()` now correctly returns a flow of strings for sent and received transactions. Issue **#1154**.
+- `CompactBlockProcessor` now triggers transaction polling while block synchronization is in progress as expected.
+  Clients will be notified briefly after every new transaction is discovered via `Synchronizer.transactions` API.
```
daira, line 32:
```suggestion
  Clients will be notified shortly after every new transaction is discovered via `Synchronizer.transactions` API.
```

backend-lib/Cargo.lock
```diff
@@ -775,7 +760,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "969c513e03167e65d4bb59f5c51ec3820210975044ad7f218ab801fc169760fa"
 dependencies = [
  "base58",
- "hdwallet",
+ "hdwallet 0.3.1",
```
daira, line 763: `hdwallet` and `hdwallet-bitcoin` are in the same repo, so I'm confused that they aren't being bumped together. What depends on the older version of `hdwallet-bitcoin`?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeBackendImpl.kt
HonzaR: I like that idea. The `createAccountAndGetSpendingKey` API changed.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniScanProgress.kt
```diff
@@ -0,0 +1,30 @@
...
+class JniScanProgress(
+    val numerator: Long,
+    val denominator: Long
+) {
+    init {
+        require(numerator >= 0L) {
+            "Numerator $numerator is outside of allowed range [0, Long.MAX_VALUE]"
+        }
+        require(denominator >= 1L) {
+            "Denominator $denominator is outside of allowed range [1, Long.MAX_VALUE]"
```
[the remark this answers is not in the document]
HonzaR: Understand, the requirement changed. Test added.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniScanRange.kt
```diff
@@ -0,0 +1,29 @@
...
+) {
+    init {
+        // We require some of the parameters below to be in the range of unsigned integer, because of the Rust layer
+        // implementation.
+        require(startHeight.isInUIntRange()) {
+            "Height $startHeight is outside of allowed UInt range"
+        }
+        require(endHeight.isInUIntRange()) {
+            "Height $endHeight is outside of allowed UInt range"
+        }
```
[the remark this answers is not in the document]
HonzaR: Yep, added.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
HonzaR: Yes, it only deletes persisted leftover block files from the disc, and the `lastKnownHeight` is for unified error reporting across sync substages.

sdk-lib/src/test/java/cash/z/ecc/android/sdk/model/BlockHeightTest.kt
```diff
@@ -68,4 +68,26 @@ class BlockHeightTest {
...
+                ZcashNetwork.Mainnet.saplingActivationHeight.value
+        )
+        val two = BlockHeight.new(ZcashNetwork.Mainnet, ZcashNetwork.Mainnet.saplingActivationHeight.value)
+
+        assertEquals(ZcashNetwork.Mainnet.saplingActivationHeight.value, (one - two).value)
+    }
+
+    @Test
+    fun subtraction_of_long_succeeds() {
+        assertEquals(419_323L, (ZcashNetwork.Mainnet.saplingActivationHeight - 123L).value)
```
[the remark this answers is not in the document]
HonzaR: Oh, thanks! Typo fixed. Tests reworked.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1173,7 +1173,7 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_ge
         match db_data
             .get_wallet_summary(0)
             .map_err(|e| format_err!("Error while fetching scan progress: {}", e))?
-            .and_then(|summary| summary.scan_progress().filter(|r| r.denominator() > 0))
+            .and_then(|summary| summary.scan_progress().filter(|r| r.denominator() > &0))
```
HonzaR, line 1176: @nuttycom Please review this Rust FFI change, as I just used what the compiler suggested to silent error:
```
error[E0308]: mismatched types
    --> src/main/rust/lib.rs:1176:86
     |
1176 |             .and_then(|summary| summary.scan_progress().filter(|r| r.denominator() > 0))
     |                                                                                      ^
     |                                                                                      |
     |                                                                                      expected `&u64`, found integer
     |                                                                                      help: consider borrowing here: `&0`
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/TransactionEncoderImpl.kt
```diff
@@ -97,8 +98,8 @@ internal class TransactionEncoderImpl(
         backend.isValidUnifiedAddr(address)
 
     override suspend fun getConsensusBranchId(): Long {
-        val height = repository.lastScannedHeight()
-        if (height < backend.network.saplingActivationHeight) {
+        val height = backend.getMaxScannedHeight()
```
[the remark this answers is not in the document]
HonzaR: Resolved as part of https://github.com/zcash/zcash-android-wallet-sdk/pull/1232

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/WalletService.kt
```diff
@@ -35,7 +34,42 @@ suspend inline fun retryUpTo(
...
+    while (failedAttempts < retries) {
+        @Suppress("TooGenericExceptionCaught")
+        try {
+            block(failedAttempts)
+            return
+        } catch (t: Throwable) {
+            failedAttempts++
+            if (failedAttempts > retries) {
+                exceptionWrapper(t)
+            }
```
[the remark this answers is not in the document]
HonzaR: Good catch! I just changed the inner condition to `failedAttempts == retries ` to still be able to trigger the exception wrapper in the last run of the `while` loop.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/ExceptionExt.kt
```diff
@@ -44,3 +44,28 @@ internal inline fun <R> tryWarn(
         }
     }
 }
+
+internal const val PREV_HASH_MISMATCH = "The parent hash of proposed block does not correspond to the block hash at " +
+    "height" // $NON-NLS
+internal const val BLOCK_HEIGHT_DISCONTINUITY = "Block height discontinuity at height" // $NON-NLS
+internal const val TREE_SIZE_MISMATCH = "note commitment tree size provided by a compact block did not match the " +
+    "expected size at height" // $NON-NLS
```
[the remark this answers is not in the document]
HonzaR: Thanks. We also checked with the iOS side when preparing these. I just added an inline comment to warn about changing them. Should we add such a comment to the librustzcash code as well? Something like: 
```
/// Note: Remember to reflect potential changes in the texts in the connected Android and iOS SDKs and others and state it in the changelog.
```
cc @LukasKorba

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -172,21 +174,17 @@ class SdkSynchronizer private constructor(
...
-    override val saplingBalances = _saplingBalances.asStateFlow()
-    override val transparentBalances = _transparentBalances.asStateFlow()
+    override val orchardBalances = processor.orchardBalances.asStateFlow()
+    override val saplingBalances = processor.saplingBalances.asStateFlow()
+    override val transparentBalances = processor.transparentBalances.asStateFlow()
+
     override val transactions
         get() = combine(processor.networkHeight, storage.allTransactions) { networkHeight, allTransactions ->
-            val latestBlockHeight = networkHeight ?: storage.lastScannedHeight()
+            val latestBlockHeight = networkHeight ?: backend.getMaxScannedHeight()
```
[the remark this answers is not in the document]
HonzaR: The `?: backend.getMaxScannedHeight()` is a fallback value only, i.e. it's used only when the cached `networkHeight` (current chain tip) is null for some reason, e.g. not fetched yet or failed to fetch.

backend-lib/src/main/rust/lib.rs
HonzaR: I couldn't align the Android SDK FFI implementation with the iOS side on the provided link. `NonZeroU32::MIN` and `anyhow!` caused compile errors. So I just changed the lines 737 and 738 only. See related commit and update if necessary. Thank you.

CHANGELOG.md
HonzaR: Thank you, Daira, for proofreading our texts!

backend-lib/Cargo.lock
HonzaR: @str4d @nuttycom could you please assist us here?

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniScanProgress.kt
```diff
@@ -17,8 +17,10 @@ class JniScanProgress(
         require(numerator >= 0L) {
             "Numerator $numerator is outside of allowed range [0, Long.MAX_VALUE]"
         }
-        require(denominator >= 1L) {
```
nuttycom: This should be okay now that you've fixed it in the Rust layer.

backend-lib/src/main/rust/lib.rs
nuttycom: Yup, sorry I missed that.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniScanProgress.kt
HonzaR: Yep, I then reverted this change back in [1b6039c](https://github.com/zcash/zcash-android-wallet-sdk/pull/1116/commits/1b6039c36800beea17dbe156e85e1e4f45c1212f).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/derived/DerivedDataDb.kt
```diff
@@ -77,7 +77,11 @@ internal class DerivedDataDb private constructor(
                 }
                 repeat(missingAccounts) {
                     runCatching {
-                        backend.createAccountAndGetSpendingKey(checkedSeed, checkpoint, recoverUntil)
+                        backend.createAccountAndGetSpendingKey(
+                            seed = checkedSeed,
+                            treeState = checkpoint.treeState().encoded,
```
nuttycom, line 82: This should probably be `checkpoint.treeState()` instead, since this is intended to be the typesafe API; sorry, I forgot to trim the `.encoded` when I copy-pasted the line.

HonzaR: Good idea, the `createAccountAndGetSpendingKey` API updated.

daira (approved):
utACK modulo previous comments.

backend-lib/Cargo.lock
HonzaR: I can see that the iOS FFI uses version `0.4.1`. See [link](https://github.com/zcash-hackworks/zcash-light-client-ffi/blob/055b239fdd4218cdc9d7cbf354689f50891dc726/rust/Cargo.lock#L704). @str4d @nuttycom could you clarify please?

backend-lib/src/main/rust/lib.rs
HonzaR: @str4d Could you please prepare the necessary FFI changes here?

backend-lib/src/main/rust/lib.rs
HonzaR: @str4d Same case as above.

backend-lib/Cargo.lock
nuttycom: @daira upgrading to `hdwallet 0.4.1` introduces a duplicate `secp256k1` dependency, and that's one that we can't upgrade (at least right now) - which do you prefer?

str4d: I checked this, and upgrading is actually what we do need to do in this case, because it de-duplicates.

backend-lib/src/main/rust/lib.rs
str4d: Fixed.

backend-lib/Cargo.lock
daira: This seems fine to me:
```diff
zcash-android-wallet-sdk/backend-lib$ git diff
diff --git a/backend-lib/Cargo.toml b/backend-lib/Cargo.toml
index c87cff94..765a57ae 100644
--- a/backend-lib/Cargo.toml
+++ b/backend-lib/Cargo.toml
@@ -12,8 +12,8 @@ rust-version = "1.60"
 
 [dependencies]
 failure = "0.1"
-hdwallet = "0.3.1"
-hdwallet-bitcoin = "0.3"
+hdwallet = "0.4.1"
+hdwallet-bitcoin = "0.4"
 hex = "0.4"
 jni = { version = "0.20", default-features = false }
 prost = "0.12"

zcash-android-wallet-sdk/backend-lib$ cargo update -p hdwallet-bitcoin
    Updating crates.io index
    Updating base58 v0.1.0 -> v0.2.0
    Removing block-buffer v0.7.3
    Removing block-padding v0.1.5
    Removing byte-tools v0.3.1
    Removing digest v0.8.1
    Removing generic-array v0.12.4
    Removing hdwallet v0.3.1
    Updating hdwallet-bitcoin v0.3.0 -> v0.4.1
    Removing opaque-debug v0.2.3
    Removing ripemd160 v0.8.0

zcash-android-wallet-sdk/backend-lib$ cargo update -p hdwallet
    Updating crates.io index
```

str4d: Fixed. What I was remembering is that we don't want to upgrade to `secp256k1 0.27`, which is still the case.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1423,7 +1509,7 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_sh
 
         let shielding_threshold = NonNegativeAmount::from_u64(100000).unwrap();
 
-        zip317_helper(
+        let txid = zip317_helper(
```
daira, line 1512: Comment on the definition of `zip317_helper`:
`fixed::FeeRule::standard()` is deprecated. Consider using `fixed::FeeRule::non_standard(zip317::MINIMUM_FEE)`.

backend-lib/src/main/rust/lib.rs
```diff
@@ -797,16 +727,14 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_ge
...
-        let min_confirmations = 0;
+        let min_confirmations = NonZeroU32::new(1).unwrap();
 
         let amount = (&db_data)
             .get_target_and_anchor_heights(min_confirmations)
             .map_err(|e| format_err!("Error while fetching anchor height: {}", e))
             .and_then(|opt_anchor| {
                 opt_anchor
-                    .map(|(_, a)| a)
+                    .map(|(target, _)| target) // Include unconfirmed funds.
```
daira, line 737: I continue to believe that showing the user balances that are completely unconfirmed is a bad idea. As I said [here](https://github.com/zcash/librustzcash/pull/914#discussion_r1314177526):
> If I send you funds in a bunch of transactions that each have 50 unpaid actions and expire at the next block, I can be pretty confident that at most one (probably zero) of them will be mined. So, using this value as defined for anything opens the wallet up to the transparent equivalent of a Faerie Gold attack (and to various social engineering attacks, such as refund scams, if it is ever shown to a user).

daira (commented):
Flushing comment

backend-lib/src/main/rust/lib.rs
```diff
@@ -1148,23 +1016,245 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_re
...
+        name: &str,
+    ) -> Result<Vec<u8>, failure::Error> {
+        let field = env.get_field(obj, name, "[B")?.l()?.into_raw();
+        Ok(env.convert_byte_array(field)?[..].try_into()?)
+    }
+
+    Ok(CommitmentTreeRoot::from_parts(
+        BlockHeight::from_u32(long_as_u32("completingBlockHeight")?),
+        sapling::Node::read(&byte_array(env, obj, "rootHash")?[..])?,
+    ))
```
daira, line 1039: The names `long_as_u32` and `byte_array` don't imply that they do field access. Also, there seems to be no reason for the former to be a closure and the latter to be a function, or for the `[..].try_into()?` in the latter (since [JNIEnv::convert_byte_array](https://docs.rs/jni/latest/jni/struct.JNIEnv.html#method.convert_byte_array) already returns a `Result` of `Vec<u8>`).

```suggestion
    let long_field = |name| -> Result<u32, failure::Error> {
        Ok(u32::try_from(env.get_field(obj, name, "J")?.j()?)?)
    };

    let byte_array_field = |name| -> Result<Vec<u8>, failure::Error> {
        let field = env.get_field(obj, name, "[B")?.l()?.into_raw();
        Ok(env.convert_byte_array(field)?)
    }

    Ok(CommitmentTreeRoot::from_parts(
        BlockHeight::from_u32(long_field("completingBlockHeight")?),
        sapling::Node::read(&byte_array_field("rootHash")?[..])?,
    ))
```
Non-blocking.

daira: Note that there is similar code in `decode_blockmeta`. For consistency it can also be changed to:
```rust
fn decode_blockmeta(env: &JNIEnv<'_>, obj: JObject<'_>) -> Result<BlockMeta, failure::Error> {
    let long_field = |name| -> Result<u32, failure::Error> {
        Ok(u32::try_from(env.get_field(obj, name, "J")?.j()?)?)
    };

    let byte_array_field = |name| -> Result<Vec<u8>, failure::Error> {
        let field = env.get_field(obj, name, "[B")?.l()?.into_raw();
        Ok(env.convert_byte_array(field)?)
    };

    Ok(BlockMeta {
        height: BlockHeight::from_u32(long_field("height")?),
        block_hash: BlockHash(byte_array_field("hash")?[..].try_into()?),
        block_time: long_field("time")?,
        sapling_outputs_count: long_field("saplingOutputsCount")?,
        orchard_actions_count: long_field("orchardOutputsCount")?,
    })
}
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -646,18 +961,28 @@ class CompactBlockProcessor internal constructor(
...
          * Default size of batches of blocks to request from the compact block service. Then it's also used as a default
-         * size of batches of blocks to validate and scan via librustzcash. For scanning action applies this - the
-         * smaller this number the more granular information can be provided about scan state. Unfortunately, it may
-         * also lead to a lot of overhead during scanning.
+         * size of batches of blocks to scan via librustzcash. For scanning action applies this - the smaller this
+         * number the more granular information can be provided about scan state. Unfortunately, it may also lead to
+         * a lot of overhead during scanning.
          */
-        internal const val SYNC_BATCH_SIZE = 10
+        internal const val SYNC_BATCH_SIZE = 100
```
str4d, line 985: We should check during the RC period how this value affects performance on devices.

str4d (commented):
utACK d775421f5a41ead7822758f641e5d8d8755720b2

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -183,31 +206,108 @@ class CompactBlockProcessor internal constructor(
...
+                            processNewBlocksInSbSOrder(
+                                backend = backend,
+                                downloader = downloader,
+                                repository = repository,
+                                network = network,
+                                lastValidHeight = lowerBoundHeight,
+                                firstUnenhancedHeight = _processorInfo.value.firstUnenhancedHeight
+                            )
+                        }
+                        GetSubtreeRootsResult.Linear -> {
```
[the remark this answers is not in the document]
HonzaR: It's the design of Kotlin [Sealed classes](https://kotlinlang.org/docs/sealed-classes.html). They can hold `class` or `object` inside. Then when comparing with e.g. the `when` condition like here, you prefix it with `is` for classes, [is doc](https://kotlinlang.org/docs/keyword-reference.html#hard-keywords).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
HonzaR: Yep. We already did measurements, and it gives faster new block processing by a third, as we state in the updated changelog. The `10` was an initial value for the linear sync. But with the SbS, `100` works better. Agree on the later testing with other values.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/WalletService.kt
```diff
@@ -35,7 +34,42 @@ suspend inline fun retryUpTo(
...
+            block(failedAttempts)
+            return
+        } catch (t: Throwable) {
+            failedAttempts++
+            if (failedAttempts == retries) {
+                exceptionWrapper(t)
+            }
+            val duration = (initialDelayMillis.toDouble() * 2.0.pow(failedAttempts.toDouble() - 1)).toLong()
             Twig.warn(t) { "Retrying ($failedAttempts/$retries) in ${duration}s..." }
             delay(duration)
```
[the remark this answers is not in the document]
HonzaR: Oh, good catch! We're not throwing the exception wrapper as in the case of the previous `retryUpToAndThrow`, but only using it as a callback here, so I've added `return` after the wrapper invocation.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1148,23 +1016,245 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_re
...
+            Err(e) => Err(format_err!(
+                "Failed to read block metadata from WalletDb: {:?}",
+                e
+            )),
+        }
+    });
+    unwrap_exc_or(&env, res, -1)
+}
+
+fn encode_scan_progress(env: &JNIEnv<'_>, progress: Ratio<u64>) -> Result<jobject, failure::Error> {
```
daira, line 1150: Since the Rust type [does not enforce](https://github.com/zcash/librustzcash/blob/1e39daf6f83c4b62d645e74385d24299dc4414c7/zcash_client_backend/src/data_api.rs#L131-L136) any constraints on the numerator and denominator, the constraints imposed by [JniScanProgress](https://github.com/zcash/zcash-android-wallet-sdk/pull/1116/files#diff-7762a27caf428321f053453236c0a045b09f17f84385cf953891b989bb714656R17-R28) should be documented (even though this function is internal).
```suggestion
/// Returns a `JniScanProgress` object, provided that numerator is nonnegative, denominator
/// is positive, and the represented ratio is in the range 0.0 to 1.0 inclusive.
fn encode_scan_progress(env: &JNIEnv<'_>, progress: Ratio<u64>) -> Result<jobject, failure::Error> {
```

backend-lib/src/main/rust/lib.rs
```diff
@@ -1148,23 +1016,245 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_re
...
+    let output = env.new_object(
+        "cash/z/ecc/android/sdk/internal/model/JniScanProgress",
+        "(JJ)V",
+        &[
+            JValue::Long(*progress.numerator() as i64),
+            JValue::Long(*progress.denominator() as i64),
+        ],
+    )?;
```
daira, line 1160: I don't know what this will do if the constructor throws an `IllegalArgumentException` (which is what Kotlin's [`require`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require.html) does if it fails). It's not documented by [JNIEnv::new_object](https://docs.rs/jni/latest/jni/struct.JNIEnv.html#method.new_object). I assume you would have to call [`JNIEnv::exception_occurred`](https://docs.rs/jni/latest/jni/struct.JNIEnv.html#method.exception_occurred), [`exception_clear`](https://docs.rs/jni/latest/jni/struct.JNIEnv.html#method.exception_clear), etc.

backend-lib/src/main/rust/lib.rs
daira: It was that the function was generic on the array length in the case of `decode_blockmeta`, and you can't have generic closures. But the code above solves that more simply by having the closure return `Result<Vec<u8>, failure::Error>`. It's no less efficient because the closure will be inlined anyway, given that it has only one use.

nuttycom (approved):

backend-lib/src/main/rust/lib.rs
daira: Oh, what happens is that on exit from a JNI native method, the JVM checks whether there is a "pending" exception and re-throws it. So if `encode_scan_progress` is used from another JNI native method, the `IllegalArgumentException` will be re-thrown. However, in order for the Rust code not to do the wrong thing in the meantime, `JNIEnv::new_object` would have to check for a pending exception and returns an `Err` if there is one. Which it does, but I had to check the [implementations](https://github.com/jni-rs/jni-rs/blob/v0.20.0/src/wrapper/macros.rs) of the jni-rs macros to confirm that.

str4d (commented):
Re-utACK 5037102fafac757be418fc72617ca8e1532b8af7

backend-lib/src/main/rust/lib.rs
daira: Anyway this is non-blocking.

backend-lib/src/main/rust/utils.rs
```diff
@@ -18,6 +18,17 @@ pub(crate) fn java_string_to_rust(env: &JNIEnv<'_>, jstring: JString<'_>) -> Str
         .into()
 }
 
+pub(crate) fn rust_bytes_to_java(
+    env: &JNIEnv<'_>,
+    data: &[u8],
+) -> Result<jbyteArray, failure::Error> {
+    // SAFETY: jbyte (i8) has the same size and alignment as u8.
```
daira, line 25:
```suggestion
    // SAFETY: jbyte (i8) has the same size and alignment as u8, and a well-defined
    // twos-complement representation with no "trap representations".
```

backend-lib/src/test/java/cash/z/ecc/android/sdk/internal/model/JniBlockMetaTest.kt
```diff
@@ -0,0 +1,32 @@
+package cash.z.ecc.android.sdk.internal.model
```
daira, line 1: Please add a similar test for `JniScanProgress`.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -183,31 +206,108 @@ class CompactBlockProcessor internal constructor(
         // Clear any undeleted left over block files from previous sync attempts
```
daira, line 206:
```suggestion
        // Clear any undeleted left over block files from previous sync attempts.
        // lastKnownHeight is only used for error reporting.
```
Non-blocking.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -183,31 +206,108 @@ class CompactBlockProcessor internal constructor(
...
+                                val result = putSaplingSubtreeRoots(
+                                    backend = backend,
+                                    startIndex = 0,
+                                    subTreeRootList = (subTreeRootResult as GetSubtreeRootsResult.SpendBeforeSync)
+                                        .subTreeRootList,
+                                    lastValidHeight = lowerBoundHeight
+                                )
+                            ) {
+                                PutSaplingSubtreeRootsResult.Success -> {
+                                    // Lets continue with the next step
```
daira, line 243:
```suggestion
                                    // Let's continue with the next step
```
Non-blocking.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -183,31 +206,108 @@ class CompactBlockProcessor internal constructor(
...
+                                downloader = downloader,
+                                repository = repository,
+                                network = network,
+                                lastValidHeight = lowerBoundHeight,
+                                firstUnenhancedHeight = _processorInfo.value.firstUnenhancedHeight
+                            )
+                        }
+                        is GetSubtreeRootsResult.OtherFailure -> {
+                            // The server possibly replied with some unsupported error. We still approach
+                            // spend-before-sync synchronization.
```
daira, line 273:
```suggestion
                        GetSubtreeRootsResult.Linear, is GetSubtreeRootsResult.OtherFailure -> {
                            // This is caused by an empty response result or another unsupported error.
                            // Although the spend-before-sync synchronization algorithm is not supported,
                            // we can get the entire block range as we previously did for the linear sync type.
```
(Assuming this syntax works; if it doesn't then never mind.) Non-blocking.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -301,149 +370,396 @@ class CompactBlockProcessor internal constructor(
...
+        while (verifyRangeResult is VerifySuggestedScanRange.ShouldVerify) {
+            Twig.info { "Starting verification of range: $verifyRangeResult" }
+
+            // Remove existing blocks as they'll be re-downloaded
+            downloader.rewindToHeight(verifyRangeResult.scanRange.range.start)
             deleteAllBlockFiles(
                 downloader = downloader,
-                lastKnownHeight = lastScannedHeight
+                lastKnownHeight = verifyRangeResult.scanRange.range.start
             )
```
daira, line 428: This is non-atomic; does that matter? I.e. what happens if we rewind but the process is terminated before deleting some or all of the block files?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -301,149 +370,396 @@ class CompactBlockProcessor internal constructor(
...
+                        is GetMaxScannedHeightResult.Success -> result.height
+                        else -> null
+                    }
+                    lastScannedHeight?.let {
+                        downloader.rewindToHeight(lastScannedHeight)
+                    }
+                    deleteAllBlockFiles(
+                        downloader = downloader,
+                        lastKnownHeight = lastScannedHeight
+                    )
```
daira, line 484: Same potential non-atomicity issue here.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -301,149 +370,396 @@ class CompactBlockProcessor internal constructor(
...
```
daira, line 595: Same potential non-atomicity issue here.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -301,149 +370,396 @@ class CompactBlockProcessor internal constructor(
...
+                    return (syncingResult as SyncingResult.Failure).toBlockProcessingResult()
+                }
+            }
+        }
         return BlockProcessingResult.Success
```
daira, line 600: My review of this function was not particularly thorough; I'm missing a lot of the needed context.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -532,10 +847,10 @@ class CompactBlockProcessor internal constructor(
...
             try {
-                retryUpTo(UTXO_FETCH_RETRIES) {
+                retryUpToAndThrow(UTXO_FETCH_RETRIES) {
                     val tAddresses = backend.listTransparentReceivers(account)
 
-                    downloader.lightWalletClient.fetchUtxos(
+                    downloader.fetchUtxos(
                         tAddresses,
                         BlockHeightUnsafe.from(startHeight)
                     ).onEach { response ->
```
daira, line 856: Nitpick: there's a grammar error in the second sentence of the error message on line 903 ("It seems to unavailable on lightwalletd.").

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -646,18 +961,28 @@ class CompactBlockProcessor internal constructor(
...
          */
         internal const val MAX_REORG_SIZE = 100
 
         /**
          * Default size of batches of blocks to request from the compact block service. Then it's also used as a default
-         * size of batches of blocks to validate and scan via librustzcash. For scanning action applies this - the
-         * smaller this number the more granular information can be provided about scan state. Unfortunately, it may
-         * also lead to a lot of overhead during scanning.
+         * size of batches of blocks to scan via librustzcash. For scanning action applies this - the smaller this
+         * number the more granular information can be provided about scan state. Unfortunately, it may also lead to
```
daira, line 982:
```suggestion
         * size of batches of blocks to scan via librustzcash. The smaller the batch size for scanning, the more
         * granular information can be provided about scan state. Unfortunately, small batches may also lead to
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -670,6 +995,268 @@ class CompactBlockProcessor internal constructor(
          */
         internal const val REWIND_DISTANCE = 10
 
+        /**
+         * Limit millis value for restarting currently running block synchronization.
```
daira, line 999:
```suggestion
         * Timeout in milliseconds for restarting the currently running block synchronization.
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -670,6 +995,268 @@ class CompactBlockProcessor internal constructor(
...
+                    }
+                }
+            }
+
+            return latestBlockHeight
+        }
+
+        /**
+         * This operation downloads note commitment tree data from the lightwalletd server to decide if we communicate
+         * with linear or spend-before-sync node
```
daira, line 1053:
```suggestion
         * This operation downloads note commitment tree data from the lightwalletd server, to decide whether
         * we are communicating with a linear or a spend-before-sync node.
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -670,6 +995,268 @@ class CompactBlockProcessor internal constructor(
...
+            network: ZcashNetwork
+        ): GetSubtreeRootsResult {
+            Twig.debug { "Fetching SubtreeRoots..." }
+
+            var result: GetSubtreeRootsResult = GetSubtreeRootsResult.Linear
+
+            retryUpToAndContinue(GET_SUBTREE_ROOTS_RETRIES) {
+                downloader.getSubtreeRoots(
+                    startIndex = 0,
+                    maxEntries = if (network.isTestnet()) {
```
daira, line 1069: Please add a comment documenting what this is a workaround for, and pointing to an issue to remove it when possible.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -670,6 +995,268 @@ class CompactBlockProcessor internal constructor(
...
+                        65536
+                    } else {
+                        0
+                    },
+                    shieldedProtocol = ShieldedProtocolEnum.SAPLING
+                ).onEach { response ->
+                    when (response) {
+                        is Response.Success -> {
+                            Twig.verbose {
+                                "SubtreeRoot got successfully: it's completingHeight: ${response.result
```
daira, line 1079:
```suggestion
                                "SubtreeRoot fetched successfully: its completingHeight is ${response.result
```
(possessive "its"). Non-blocking.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -670,6 +995,268 @@ class CompactBlockProcessor internal constructor(
...
+                    .toList()
+                    .map {
+                        SubtreeRoot.new(it, network)
+                    }.let {
+                        result = if (it.isEmpty()) {
+                            GetSubtreeRootsResult.Linear
+                        } else {
+                            GetSubtreeRootsResult.SpendBeforeSync(it)
+                        }
+                    }
```
daira, line 1116: This looks indented by 4 spaces too many.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -784,21 +1356,26 @@ class CompactBlockProcessor internal constructor(
...
+                            resultState = resultState
                         )
                     )
 
                     // Increment and compare the range for triggering the enhancing
                     enhancingRange = enhancingRange.start..continuousResult.batch.range.endInclusive
 
-                    // Enhance is run in case of the range is on or over its limit, or in case of any failure
+                    // Enhancing is run in case of the range is on or over its limit, or in case of any failure
                     // state comes from the previous stages, or if the end of the sync range is reached
```
daira, line 1376:
```suggestion
                    // Enhancing is run when the range is on or over its limit, or there is any
                    // failure from previous stages, or the end of the sync range is reached.
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -810,52 +1387,83 @@ class CompactBlockProcessor internal constructor(
...
+        private fun getBatchCount(syncRanges: List<ClosedRange<BlockHeight>>): Long {
+            var allRangesBatchCount = 0L
+            var allMissingBlocksCount = 0L
+
+            syncRanges.forEach { range ->
+                val missingBlockCount = range.endInclusive.value - range.start.value + 1
+                val batchCount = (
+                    missingBlockCount / SYNC_BATCH_SIZE +
+                        (if (missingBlockCount.rem(SYNC_BATCH_SIZE) == 0L) 0 else 1)
+                    )
```
daira, line 1441:
```suggestion
                val batchCount = (missingBlockCount + SYNC_BATCH_SIZE - 1) / SYNC_BATCH_SIZE
```
Non-blocking.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -810,52 +1387,83 @@ class CompactBlockProcessor internal constructor(
...
+        }
 
+        /**
+         * Prepare list of all [BlockBatch] internal objects to be processed during a range of
+         * blocks processing
+         *
+         * @param syncRange Current range to be processed
+         * @param network The network we are operating on
+         *
+         * @return List of [BlockBatch] to for synchronization
```
daira, line 1460:
```suggestion
         * @return List of [BlockBatch] to prepare for synchronization
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -879,89 +1492,103 @@ class CompactBlockProcessor internal constructor(
...
+            var downloadException: CompactBlockProcessorException.FailedDownloadException? = null
+
+            retryUpToAndContinue(
+                retries = RETRIES,
+                exceptionWrapper = {
+                    downloadException = CompactBlockProcessorException.FailedDownloadException(it)
+                    downloadException!!
+                }
+            ) { failedAttempts ->
+                @Suppress("MagicNumber")
```
daira, line 1509: Where's the magic number here? (0 isn't magic.) Oh well, non-blocking.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -1143,38 +1791,50 @@ class CompactBlockProcessor internal constructor(
      *
```
daira, line 1791: This sets global parameters, rather than returning a new instance of `ProcessorInfo`, correct? Please make that clear in the documentation.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -1143,38 +1791,50 @@ class CompactBlockProcessor internal constructor(
...
+            overallSyncRange = overallSyncRange,
             firstUnenhancedHeight = firstUnenhancedHeight
         )
     }
 
-    private suspend fun handleChainError(errorHeight: BlockHeight) {
-        // TODO [#683]: consider an error object containing hash information
-        // TODO [#683]: https://github.com/zcash/zcash-android-wallet-sdk/issues/683
+    /**
+     * Emit an instance of progress.
```
daira, line 1815:
```suggestion
     * Set the progress for this [CompactBlockProcessor].
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -1143,38 +1791,50 @@ class CompactBlockProcessor internal constructor(
...
+     *
+     * @param progress the block syncing progress of type [PercentDecimal] in the range of [0, 1]
+     */
+    private fun setProgress(progress: PercentDecimal = _progress.value) {
+        _progress.value = progress
+    }
+
+    /**
+     * Transmits the given state for this processor.
```
daira, line 1824:
```suggestion
     * Sets the state of this [CompactBlockProcessor].
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -1196,145 +1856,84 @@ class CompactBlockProcessor internal constructor(
     /**
      * Rewind back at least two weeks worth of blocks.
```
daira, line 1857: `rewindToNearestHeight` rewinds to the "nearest" supported height. Is that guaranteed to be _at least_ two weeks back?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/WalletService.kt
```diff
@@ -35,7 +34,43 @@ suspend inline fun retryUpTo(
...
+    }
+}
+
+/**
+ * Execute the given block and if it fails, retry up to [retries] more times. If none of the
+ * retries succeed, then leave the block execution unfinished and continue.
+ *
+ * @param retries the number of times to retry the block after the first attempt fails.
+ * @param exceptionWrapper a function that can wrap the final failure to add more useful information
+ *  * or context. Default behavior is to just return the final exception.
```
daira, line 50:
```suggestion
 * or context. Default behavior is to just return the final exception.
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/WalletService.kt
```diff
@@ -35,7 +34,43 @@ suspend inline fun retryUpTo(
...
+    initialDelayMillis: Long = 500L,
+    block: (Int) -> Unit
+) {
+    var failedAttempts = 0
+    while (failedAttempts < retries) {
+        @Suppress("TooGenericExceptionCaught")
+        try {
+            block(failedAttempts)
+            return
+        } catch (t: Throwable) {
```
daira, line 67: I know that in Java, it's considered bad form to catch `Throwable` because that includes for example `InterruptedException` and `ThreadDeath`, which interferes with thread termination.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/ScanProgress.kt
```diff
@@ -0,0 +1,28 @@
...
+internal data class ScanProgress(
+    private val numerator: Long,
+    private val denominator: Long
+) {
+    override fun toString() = "ScanProgress($numerator/$denominator) -> ${getSafeRation()}"
+
+    /**
+     * Returns progress ratio in [0, 1] range. Any out-of-range value is treated as 0.
+     */
+    fun getSafeRation() = numerator.toFloat().div(denominator).let { ration ->
```
daira, line 12: Should this be `Ratio` rather than `Ration`?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/BlockHeight.kt
```diff
@@ -39,6 +39,24 @@ data class BlockHeight internal constructor(val value: Long) : Comparable<BlockH
         return BlockHeight(value + other)
     }
 
+    operator fun minus(other: BlockHeight) = BlockHeight(value - other.value)
+
+    operator fun minus(other: Int): BlockHeight {
+        require(other >= 0) {
+            "Cannot subtract negative value $other to BlockHeight"
```
daira, line 46:
```suggestion
            "Cannot subtract negative value $other from BlockHeight"
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/BlockHeight.kt
```diff
@@ -39,6 +39,24 @@ data class BlockHeight internal constructor(val value: Long) : Comparable<BlockH
...
+        }
+
+        return BlockHeight(value - other.toLong())
+    }
+
+    operator fun minus(other: Long): BlockHeight {
+        require(other >= 0) {
+            "Cannot subtract negative value $other to BlockHeight"
```
daira, line 54:
```suggestion
            "Cannot subtract negative value $other from BlockHeight"
```
Non-blocking.

sdk-lib/src/test/java/cash/z/ecc/android/sdk/model/BlockHeightTest.kt
```diff
@@ -68,4 +68,32 @@ class BlockHeightTest {
...
+        )
+    }
+
+    @Test
+    fun subtraction_of_int_succeeds() {
+        assertEquals(
+            ZcashNetwork.Mainnet.saplingActivationHeight.value,
+            (BlockHeight(419_323) - 123).value
+        )
+    }
```
daira, line 98: Check that subtracting a negative value fails.

daira (approved):
utACK with comments.

backend-lib/src/main/rust/lib.rs
HonzaR: Updating the fee model will be part of the this planned issue https://github.com/zcash/zcash-android-wallet-sdk/issues/776

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
HonzaR: Yep, it works like this. Thanks for pointing this out. I just used the shorter version and merged their comments.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
HonzaR: In this scenario, the remaining files will be deleted in the next synchronization attempt in about 30 seconds in most cases or in the next client app start. Both of these cases are fine, as the helper block files are rewritten in the device storage, and they are used only as links to the rust layer to avoid passing all block files to the database.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
HonzaR: Yeah, I understand. Thank you anyway!

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
HonzaR: This testnet workaround is not needed now, according to Kris's and Lukas's communication about iOS implementation.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
HonzaR: Good catch. Thanks! It's possibly a leftover from the previous iteration.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
HonzaR: It depends on the Rust implementation of `getNearestRewindHeight.` Kotlin side guarantees that (except the activation height limit case). So, I slightly changed the comment to reflect this. Thanks!

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/WalletService.kt
HonzaR: Yes, totally agree. But I'd keep this implementation, as we really need to catch any kind of exception and report it to the caller. That's why we have suppression for the lint checks here `@Suppress("TooGenericExceptionCaught")`. We could switch this older pattern to Kotlin's [`runCatching`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/run-catching.html), but that does the same inside:
```
/**
 * Calls the specified function [block] and returns its encapsulated result if invocation was successful,
 * catching any [Throwable] exception that was thrown from the [block] function execution and encapsulating it as a failure.
 */
@InlineOnly
@SinceKotlin("1.3")
public inline fun <R> runCatching(block: () -> R): Result<R> {
    return try {
        Result.success(block())
    } catch (e: Throwable) {
        Result.failure(e)
    }
}
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/ScanProgress.kt
HonzaR: Correct, good catch. Thanks!

sdk-lib/src/test/java/cash/z/ecc/android/sdk/model/BlockHeightTest.kt
HonzaR: That is already covered in lines 58-70 with the older tests `subtraction_of_int_fails` and `subtraction_of_long_fails` tests.

PR #1247, opened 2023-09-18
[#1241][#1243][#1222] Rewinding when ContinuityError

### LINKED ISSUE #1222: zcash-android-wallet-sdk Enrich `BlockProcessingResult.SyncFailure` with root cause
## Is your feature request related to a problem? Please describe.
We'd like to enrich `BlockProcessingResult.SyncFailure` with a specific type of root cause of this failure. It could be a failure from downloading, scanning, etc. Thanks to this, we'll be able to decide more precisely about subsequent rewind action, e.g. in case of lost connection we don't need to rewind.

## Describe the solution you'd like
Add the `cause` parameter to the `SyncFailure` data class:
```
data class SyncFailure(val failedAtHeight: BlockHeight?, cause: enum, val error: Throwable) : BlockProcessingResult()
```

### LINKED ISSUE #1239: Post-v2.0.0-rc1 cleanup to be addressed prior to v2.0.0 release
This issue is to track that comments from https://github.com/zcash/zcash-android-wallet-sdk/pull/1116 that were not resolved prior to the v2.0.0-rc1 release are addressed before v2.0.0 final.

### LINKED ISSUE #1241: Remove rewind for every verify scan range
The verify scan range doesn't require a rewind to be called. This should be handled by isContinuityError in scan.

### LINKED ISSUE #1243: SuggestScanRanges must be called after Rewind
When continuity error occurs it must trigger rewind (current processing batch start -10 blocks) immediately followed by suggestScanRanges call to update the ranges.

HonzaR:
- Closes #1241
- The original solution came from the pseudocode requirement: _Download the blocks in `scan_range` into the block source, overwriting any existing blocks in this range._ Removed now.
- Closes #1243
- Closes #1222 as it was created to determine which failure type comes and don’t rewind for all of them
- Closes #1239 since it is the last required post-v2.0.0-rc1 cleanup.
- Rewind is done only when Continuity-error appears now. In case of other sync failures, the sync loop sleeps for a short time and then retries. Internal actions like fetching subtree roots, fetching chain tip, downloading, scanning, etc., still have their internal retry mechanisms.
- For the calculation of the rewind height, we use the existing `checkContinuityErrorResult` method, originally used only for validation use cases but later incorrectly used for other failures too.
- Tested manually in several scenarios, e.g. lost internet connection or app going to background
---

> **Note**

# Author

# Reviewer

### DIFF
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
@@ -409,13 +409,6 @@ class CompactBlockProcessor internal constructor(
         while (verifyRangeResult is VerifySuggestedScanRange.ShouldVerify) {
             Twig.info { "Starting verification of range: $verifyRangeResult" }
 
-            // Remove existing blocks as they'll be re-downloaded
-            downloader.rewindToHeight(verifyRangeResult.scanRange.range.start)
-            deleteAllBlockFiles(
-                downloader = downloader,
-                lastKnownHeight = verifyRangeResult.scanRange.range.start
-            )
-
             var syncingResult: SyncingResult = SyncingResult.AllSuccess
             runSyncingAndEnhancingOnRange(
                 backend = backend,
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
@@ -24,6 +24,7 @@ import cash.z.ecc.android.sdk.exception.RustLayerException
 import cash.z.ecc.android.sdk.ext.ZcashSdk
 import cash.z.ecc.android.sdk.ext.ZcashSdk.MAX_BACKOFF_INTERVAL
 import cash.z.ecc.android.sdk.ext.ZcashSdk.POLL_INTERVAL
+import cash.z.ecc.android.sdk.ext.ZcashSdk.POLL_INTERVAL_SHORT
 import cash.z.ecc.android.sdk.internal.Twig
 import cash.z.ecc.android.sdk.internal.TypesafeBackend
 import cash.z.ecc.android.sdk.internal.block.CompactBlockDownloader
@@ -304,20 +305,29 @@ class CompactBlockProcessor internal constructor(
                         }
                         consecutiveChainErrors.set(0)
                         val napTime = calculatePollInterval()
-                        Twig.debug {
+                        Twig.info {
                             "$summary Sleeping for ${napTime}ms " +
                                 "(latest height: ${_processorInfo.value.networkBlockHeight})."
                         }
                         delay(napTime)
+                    }
+                    is BlockProcessingResult.ContinuityError -> {
+                        Twig.error {
+                            "Failed while processing blocks at height: ${result.failedAtHeight} with continuity " +
+                                "error: ${result.error}"
+                        }
+                        // No nap time set to immediately continue with the following block synchronization attempt
                     }
                     is BlockProcessingResult.SyncFailure -> {
                         Twig.error {
                             "Failed while processing blocks at height: ${result.failedAtHeight} with: " +
                                 "${result.error}"
                         }
-                        // TODO [#1222]: Enrich BlockProcessingResult.SyncFailure with root cause
-                        // TODO [#1222]: https://github.com/zcash/zcash-android-wallet-sdk/issues/1222
-                        checkErrorResult(result.failedAtHeight)
+                        val napTime = calculatePollInterval(true)
+                        Twig.info {
+                            "Sleeping for ${napTime}ms (latest height: ${_processorInfo.value.networkBlockHeight})."
+                        }
+                        delay(napTime)
                     }
                     is BlockProcessingResult.Success -> {
                         // Do nothing.
@@ -329,7 +339,7 @@ class CompactBlockProcessor internal constructor(
         stop()
     }
 
-    suspend fun checkErrorResult(failedHeight: BlockHeight?) {
+    suspend fun checkContinuityErrorResult(failedHeight: BlockHeight?) {
         if (consecutiveChainErrors.get() >= RETRIES) {
             val errorMessage = "ERROR: unable to resolve reorg at height $failedHeight after " +
                 "${consecutiveChainErrors.get()} correction attempts!"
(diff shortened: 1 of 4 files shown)


sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -569,21 +576,25 @@ class CompactBlockProcessor internal constructor(
...
-                    }
-                    lastScannedHeight?.let {
-                        downloader.rewindToHeight(lastScannedHeight)
-                    }
+                }
+                is SyncingResult.ContinuityError -> {
+                    val failedHeight = (syncingResult as SyncingResult.ContinuityError).failedAtHeight
+                    checkContinuityErrorResult(failedHeight)
+                    // This step is independent of the rewind action as it only removes temporary persisted block files
+                    // from the device storage. The files will be re-downloaded in the following synchronization cycle.
```
nuttycom, line 584: Can you point to where the rewind will be called?

HonzaR: There is `handleChainError` in the `checkContinuityErrorResult`. https://github.com/zcash/zcash-android-wallet-sdk/blob/613c951282b692bc6c9836198808d1668a591e59/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt#L1820

nuttycom: Aha, okay, I didn't expand `checkContinuityErrorResult`. Thanks. The naming there is a little confusing, because it's actually handling the continuity error, not just checking for it.

HonzaR: Yeah, understand. What about renaming `checkContinuityErrorResult` to `processContinuityError`?

LukasKorba (approved):

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -304,20 +305,29 @@ class CompactBlockProcessor internal constructor(
...
+                    }
                     is BlockProcessingResult.SyncFailure -> {
                         Twig.error {
                             "Failed while processing blocks at height: ${result.failedAtHeight} with: " +
                                 "${result.error}"
                         }
-                        // TODO [#1222]: Enrich BlockProcessingResult.SyncFailure with root cause
-                        // TODO [#1222]: https://github.com/zcash/zcash-android-wallet-sdk/issues/1222
-                        checkErrorResult(result.failedAtHeight)
+                        val napTime = calculatePollInterval(true)
```
str4d, line 326: Previously `checkErrorResult` would only retry a few times, and then would fail the sync process. AFAICT this no longer happens in this code pathway: for arbitrary non-continuity failures, we will now retry indefinitely (even if the errors are permanent). Is this desirable, or do we need to keep the old retry behaviour here?

The old behaviour also attempted to resolve the error by rewinding. This would only help with some kinds of recoverable errors, and we don't (and in some cases can't) distinguish between them, so I'm not against us just not trying to rewind, but indefinite retries without any rewinding seems like it won't be as useful as eventually telling the user "there's an error, you need to manually do something to fix it".

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -329,7 +339,7 @@ class CompactBlockProcessor internal constructor(
         stop()
     }
 
-    suspend fun checkErrorResult(failedHeight: BlockHeight?) {
+    suspend fun checkContinuityErrorResult(failedHeight: BlockHeight?) {
         if (consecutiveChainErrors.get() >= RETRIES) {
```
str4d, line 343: Conversely, we _don't_ want to abort after a set number of `ContinuityError` occurrences; reorgs can potentially occur frequently (e.g. during network weather), and the wallet needs to follow them no matter how many occur.

It feels like `checkContinuityErrorResult` should be a separate handler (or just call `handleChainError` directly, perhaps renamed) instead of repurposing `checkErrorResult`.

HonzaR: Okay, that is a new requirement for the rewind logic in comparison with the existing implementation. I've pushed a change that keeps trying to reorg with a changed rewind height (i.e. calls `handleChainError` directly). Thanks for pointing this out!

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
HonzaR: That is true. There were more types of possible implementations of how to get out of the _rewind for every error_ problem. The previous one used a minimal changes approach with the _nap time_ implementation. But agree that it would change the SDK error-handling behavior. So I just pushed a change that adds fail logic after the `RETRIES` limit is reached, and thus keeps the same approach as before but does not rewind.

str4d (approved):
utACK c134ad7db6646abdf4df041d8dd8e25e5b90005a

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -302,42 +311,71 @@ class CompactBlockProcessor internal constructor(
...
-                        // TODO [#1222]: Enrich BlockProcessingResult.SyncFailure with root cause
-                        // TODO [#1222]: https://github.com/zcash/zcash-android-wallet-sdk/issues/1222
-                        checkErrorResult(result.failedAtHeight)
+                        val failed = checkErrorAndFail(result.failedAtHeight, result.error)
+                        if (!failed) {
+                            val napTime = calculatePollInterval(true)
+                            Twig.info {
+                                "Sleeping for ${napTime}ms (latest height: ${_processorInfo.value.networkBlockHeight})."
+                            }
+                            delay(napTime)
```
daira, line 342: Nit: this code occurs three times and could be factored into a helper method.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -569,21 +604,25 @@ class CompactBlockProcessor internal constructor(
...
+                        lastKnownHeight = failedHeight
                     )
+                    return (syncingResult as SyncingResult.ContinuityError).toBlockProcessingResult()
+                }
+                is SyncingResult.Failure -> {
+                    // Other failure types
                     return (syncingResult as SyncingResult.Failure).toBlockProcessingResult()
                 }
+                else -> {
+                    // The rest types of result are not expected here
```
daira, line 624: Log this at info level.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -458,21 +489,25 @@ class CompactBlockProcessor internal constructor(
...
```
daira, line 509: Log this at info level.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
HonzaR: Just added a link to the filed issue to the code base. Thanks.

PR #1344, opened 2024-01-08
Replace `Backend.{createToAddress, shieldToAddress}` with proposal-based methods

str4d:
Migrates to the latest in-progress revision of the Zcash Rust crates, so this is not suitable for inclusion in an SDK release yet.

Closes Electric-Coin-Company/zcash-android-wallet-sdk#1338.

> **Note**

# Author

# Reviewer


backend-lib/build.gradle.kts
```diff
@@ -94,10 +102,37 @@ cargo {
...
+        }
+    }
+}
+
 dependencies {
     api(projects.lightwalletClientLib)
 
     implementation(libs.androidx.annotation)
+    // TODO: We don't need gRPC here, only protoc
+    implementation(libs.bundles.grpc)
```
str4d, line 135: @HonzaR I just copied stuff from `lightwallet-client-lib/build.gradle.kts` until I got the protobufs to generate; I expect we can refine things here since we only care about the protobufs in `backend-lib`, not gRPC.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/ProposalUnsafe.kt
```diff
@@ -0,0 +1,29 @@
...
+    init {
+        require(inner.feeRule != FeeRule.FeeRuleNotSpecified) {
+            "Fee rule must be specified"
+        }
+    }
+
+    companion object {
+        fun parse(encoded: ByteArray): ProposalUnsafe {
+            val inner = Proposal.parseFrom(encoded)
+            return ProposalUnsafe(inner)
```
str4d, line 18: Despite this being called `ProposalUnsafe`, I expect we'll do most validation in this class because we have access to the parsed protobuf (which can't be used outside `backend-lib` without requiring `sdk-lib` to depend on the protobuf library, and I figure it's better to keep it being a backend-internal detail). The wrapping `Proposal` in `sdk-lib` is wrapping the exposed types here in safe API types like `Zatoshi`.

backend-lib/src/main/proto/proposal.proto
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) 2023 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+syntax = "proto3";
+package cash.z.wallet.sdk.ffi;
+
+// A data structure that describes the inputs to be consumed and outputs to
+// be produced in a proposed transaction.
+message Proposal {
```
nuttycom, line 10: I'd prefer if [zcash/librustzcash#1060] could merge before we commit `proposal.proto` to this repository, since there are substantial changes there.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/Proposal.kt
```diff
@@ -0,0 +1,21 @@
+package cash.z.ecc.android.sdk.internal.model
+
+import cash.z.ecc.android.sdk.model.Zatoshi
+
+class Proposal(
+    private val inner: ProposalUnsafe
+) {
+    companion object {
+        fun parse(proposal: ProposalUnsafe): Proposal {
+            return Proposal(proposal)
```
nuttycom, line 10: Is the intent here to just create a wrapper so that the underlying `ProposalUnsafe` is not mutable, and that `Proposal` is all that is exposed in public APIs?

I don't understand why this method is called `parse`, given that it doesn't appear to do any parsing or validation. Also, given that `toUnsafe` is exposed, I'm not sure what this class is supposed to be doing.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/ProposalUnsafe.kt
```diff
@@ -0,0 +1,29 @@
...
+) {
+    init {
+        require(inner.feeRule != FeeRule.FeeRuleNotSpecified) {
+            "Fee rule must be specified"
+        }
+    }
+
+    companion object {
+        fun parse(encoded: ByteArray): ProposalUnsafe {
+            val inner = Proposal.parseFrom(encoded)
```
nuttycom, line 17: Isn't protobuf parsing fallible? How is such an error represented here, is it thrown? Is there no way of indicating that in the type signature?

backend-lib/src/main/rust/lib.rs
```diff
@@ -343,7 +336,8 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustDerivation
 
         let ufvks: Vec<_> = (0..accounts)
             .map(|account| {
-                let account_id = AccountId::from(account);
+                let account_id =
+                    AccountId::try_from(account).map_err(|_| format_err!("Invalid account ID"))?;
```
nuttycom, line 340: Since `accounts` has been cast from a jint, it cannot be more than 2^31-1, so this error should never occur. This could also be an `expect`.

backend-lib/src/main/rust/lib.rs
```diff
@@ -103,6 +93,13 @@ fn block_db(env: &JNIEnv<'_>, fsblockdb_root: JString<'_>) -> Result<FsBlockDb,
         .map_err(|e| format_err!("Error opening block source database connection: {:?}", e))
 }
 
+fn account_id_from_jint(account: jint) -> Result<AccountId, failure::Error> {
+    u32::try_from(account)
+        .map_err(|_| ())
+        .and_then(|id| AccountId::try_from(id).map_err(|_| ()))
```
nuttycom, line 99: This `try_from` should never be able to fail, because the `u32::try_from` will have ensured that the resulting u32 falls into the valid range of account IDs. I think that using `map` and `expect` here would actually be appropriate, because if the inner `try_from` fails that _should_ indicate that something is crashingly wrong.

backend-lib/src/main/rust/lib.rs
```diff
@@ -589,10 +578,10 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_is
         let network = parse_network(network_id as u32)?;
         let addr = utils::java_string_to_rust(&env, addr);
 
-        match RecipientAddress::decode(&network, &addr) {
+        match Address::decode(&network, &addr) {
             Some(addr) => match addr {
-                RecipientAddress::Shielded(_) => Ok(JNI_TRUE),
-                RecipientAddress::Transparent(_) | RecipientAddress::Unified(_) => Ok(JNI_FALSE),
+                Address::Sapling(_) => Ok(JNI_TRUE),
+                Address::Transparent(_) | Address::Unified(_) => Ok(JNI_FALSE),
```
nuttycom, line 584: nit: although this is not a change to behavior from this PR, and is instead just highlighted because of the rename, I'm now confused about what the behavior here is intended to be.

Does the name of the `isValidShieldedAddress` method need to change to `isValidSaplingAddress`? I don't understand why a Unified address is not considered a valid shielded address, given that it must have a shielded receiver.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1326,58 +1317,44 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_de
...
-    usk: jbyteArray,
+    account: jint,
     to: JString<'_>,
     value: jlong,
     memo: jbyteArray,
-    spend_params: JString<'_>,
-    output_params: JString<'_>,
     network_id: jint,
     use_zip317_fees: jboolean,
 ) -> jbyteArray {
```
nuttycom, line 1347: nit: It might be nice to have a type alias for `jbyteArray` that indicates that this is a serialized proposal protobuf. The fact that the return type didn't change was a little disorienting.

nuttycom (approved):
utACK with nonblocking nits. Some of these, on further reflection, are just design issues that should probably be cleaned up separately; the one that I'd like clarification on is the `UnsafeProposal` vs `Proposal` situation and the meaning of the `parse` method on `Proposal`; please document those types/methods more completely.

nuttycom (changes requested):
Now that [zcash/librustzcash#1060] has merged, please update `proposal.proto` before this merges.

backend-lib/src/main/proto/proposal.proto
str4d: As the PR in question has now been merged, I'll rebase this PR to include it.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/ProposalUnsafe.kt
```diff
@@ -0,0 +1,29 @@
+package cash.z.ecc.android.sdk.internal.model
+
+import cash.z.wallet.sdk.ffi.ProposalOuterClass.FeeRule
+import cash.z.wallet.sdk.ffi.ProposalOuterClass.Proposal
+
+class ProposalUnsafe(
```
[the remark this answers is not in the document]
str4d: I don't intend for it to be exposed to the SDK user, but IDK if I succeeded within Kotlin.

backend-lib/src/main/rust/lib.rs
str4d: Yeah, the FFI method should be renamed as well.

str4d:
[Rebased on `main`](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/e503d1612963d7efabd5c201215f1f4960d3b339..c00b0de23c559ae4bb998a13c06f5c239a25e3c1).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/Proposal.kt
str4d: The `ProposalUnsafe` class only exists because of the split between `backend-lib` and `sdk-lib`. I cannot access the type-safe types like `Zatoshi` from inside `backend-lib`, but I also don't want to expose the proposal protobuf outside `backend-lib`. So this `Proposal.parse` function is really the only one I want, and we will need to do some safety checks here depending on what we want to expose, but for now I'll rename it to something else.

backend-lib/src/main/rust/lib.rs
str4d: Hm, this is actually exposed all the way to the outer SDK API. I'm only going to rename the FFI layer and then the SDK part can be decided later.

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/c00b0de23c559ae4bb998a13c06f5c239a25e3c1..0a7cf9626c4ecd1e31e81ce746841e7614dd5d7c) to address review comments.

nuttycom (changes requested):
`proposal.proto` is still the older version, not what's on current ][zcash/librustzcash@main]

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/0a7cf9626c4ecd1e31e81ce746841e7614dd5d7c..b3cfad5ed4fac0b9521eec59dedc406ebacf8003) to update `proposal.proto` (I had done so locally, and then stashed it to split the changes across the separate commits and forgot to unstash it).

nuttycom (approved):
utACK b3cfad5ed4fac0b9521eec59dedc406ebacf8003

nuttycom:
There appear to be a bunch of linting errors being reported.

backend-lib/build.gradle.kts
HonzaR: I just pushed a little improvement of the dependency that we need here.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/ProposalUnsafe.kt
HonzaR: Yes, that is the way we do it with the other models as well.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/ProposalUnsafe.kt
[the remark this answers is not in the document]
HonzaR: It's not ideal, as the class is still reachable from clients, but this is how we used to with, e.g., jni models too. We're helping ourselves with the `internal` in the package.

HonzaR (commented):
Thanks for these changes. I just pushed a few minor commits.

IIUC our intention here is to have a two-step transaction submitting? I.e. 1. get transaction proposal, 2. submit transition. 

To achieve this, we need to update the `Synchronizer`'s API to provide the new propose API methods, which would call related missing propose methods in the `txManager` and tweak the existing `sendToAddress` and `shieldToAddress` Is that correct?

str4d:
> IIUC our intention here is to have a two-step transaction submitting? I.e. 1. get transaction proposal, 2. submit transition.
>
> To achieve this, we need to update the Synchronizer's API to provide the new propose API methods, which would call related missing propose methods in the txManager and tweak the existing sendToAddress and shieldToAddress Is that correct

It's not strictly necessary, but it would make the most sense to do this. If we don't make this change then we end up calling the propose API twice, once to get the fee and then again during transaction creation, and the second time the fee could technically change. If we expose a `Proposal` type to SDK users that has the fee API on it, then wallets can use that to display and confirm fee, and then pass the `Proposal` back to the SDK for transaction creation.

Eventually we will want to enable the SDK user to request that proofs start being created in the background at the time the `Proposal` is created, so then if they pass that `Proposal` back the SDK can look up that work and create the final transaction more quickly. So if there are changes to how the public two-step API works that would be necessary to enable that background caching and subsequent association, that don't make implementing this change harder now, then we might as well incorporate it into the new API.

str4d:
Rebased on #1357 and now targeting the 2.1.0 feature branch. I will rebase again once that merges.

str4d:
[Rebased on `feature-2.1.0`](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/8f5b3c23192522c9b01999ae3789368e25acff97..14da738f5d2a9cd4cf206dd793d3bee675b885d5) now that #1357 is merged.

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/14da738f5d2a9cd4cf206dd793d3bee675b885d5..457984ee85f83327a6654e1cc0b5826f46081bd1) to fix lints.

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/14da738f5d2a9cd4cf206dd793d3bee675b885d5..f3a56d11417a6275ca14366a11d3af1323ac4f811) to fix lints.

HonzaR (approved):
Additionally, I tested it with Zashi via included builds, and it works as expected.

PR #1357, opened 2024-01-19
Migrate to `jni 0.21`

str4d:
In addition to the necessary breaking changes, we also clean up the style to match what is used in the `jni` crate examples.

> **Note**

# Author

# Reviewer

### DIFF
--- a/backend-lib/Cargo.lock
+++ b/backend-lib/Cargo.lock
@@ -496,7 +496,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ac3e13f66a2f95e32a39eaa81f6b95d42878ca0e1db0c7543723dfe12557e860"
 dependencies = [
  "libc",
- "windows-sys",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -740,7 +740,7 @@ version = "0.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5444c27eef6923071f7ebcc33e3444508466a76f7a2b93da00ed6e19f30c1ddb"
 dependencies = [
- "windows-sys",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -788,16 +788,18 @@ checksum = "af150ab688ff2122fcef229be89cb50dd66af9e01a4ff320cc137eecc9bacc38"
 
 [[package]]
 name = "jni"
-version = "0.20.0"
+version = "0.21.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "039022cdf4d7b1cf548d31f60ae783138e5fd42013f6271049d7df7afadef96c"
+checksum = "1a87aa2bb7d2af34197c04845522473242e1aa17c12f4935d5856491a7fb8c97"
 dependencies = [
  "cesu8",
+ "cfg-if",
  "combine",
  "jni-sys",
  "log",
  "thiserror",
  "walkdir",
+ "windows-sys 0.45.0",
 ]
 
 [[package]]
@@ -835,7 +837,7 @@ version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2b6f1427d9c43b1cce87434c4d9eca33f43bdbb6246a762aa823a582f74c1684"
 dependencies = [
- "windows-sys",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -1453,7 +1455,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -1678,7 +1680,7 @@ dependencies = [
  "fastrand",
  "redox_syscall",
  "rustix",
- "windows-sys",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


backend-lib/Cargo.lock
```diff
@@ -788,16 +788,18 @@ checksum = "af150ab688ff2122fcef229be89cb50dd66af9e01a4ff320cc137eecc9bacc38"
...
+checksum = "1a87aa2bb7d2af34197c04845522473242e1aa17c12f4935d5856491a7fb8c97"
 dependencies = [
  "cesu8",
+ "cfg-if",
  "combine",
  "jni-sys",
  "log",
  "thiserror",
  "walkdir",
+ "windows-sys 0.45.0",
```
daira, line 802: This is kind-of annoying. Oh well.

backend-lib/src/main/rust/lib.rs
```diff
@@ -80,16 +80,16 @@ fn print_debug_state() {
 }
 
 fn wallet_db<P: Parameters>(
-    env: &JNIEnv<'_>,
+    env: &mut JNIEnv,
     params: P,
```
daira, line 84: I'm surprised this wasn't &mut already, given that [a `JNIEnv` can only be used by a single thread](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/invocation.html#attaching_to_the_vm).

backend-lib/src/main/rust/lib.rs
```diff
@@ -80,16 +80,16 @@ fn print_debug_state() {
 }
 
 fn wallet_db<P: Parameters>(
-    env: &JNIEnv<'_>,
+    env: &mut JNIEnv,
```
daira, line 83: I'm surprised this wasn't &mut already, given that [a `JNIEnv` can only be used by a single thread](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/invocation.html#attaching_to_the_vm).

backend-lib/src/main/rust/lib.rs
```diff
@@ -146,13 +146,13 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_in
 ///
 /// Returns 0 if successful, or -1 otherwise.
 #[no_mangle]
-pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_initBlockMetaDb(
-    env: JNIEnv<'_>,
-    _: JClass<'_>,
-    fsblockdb_root: JString<'_>,
+pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_initBlockMetaDb<'local>(
+    mut env: JNIEnv<'local>,
```
daira, line 150: I was confused by this being `mut env` rather than `&mut env`. I guess [this section of the `jni` crate docs](https://docs.rs/jni/latest/jni/struct.JNIEnv.html#cannot-borrow-as-mutable) partially explains it: if this parameter were `&mut env` then it would need to be last. But it is not clear to me *why* that is; or what the trade-offs are between making it `mut env`, and making it `&mut env` with reordered parameters.

daira: If it is `mut env` then you always need to create a new `JNIEnv` in the caller like this, correct?
```rust
let env = jvm.attach_current_thread().unwrap();
```

backend-lib/src/main/rust/lib.rs
```diff
@@ -198,28 +196,24 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_in
...
-            JValue::Object(unsafe { JObject::from_raw(bytes) }),
-        ],
-    )?;
-    Ok(output.into_raw())
+        &[JValue::Int(u32::from(account) as i32), (&bytes).into()],
+    )
 }
 
-fn decode_usk(env: &JNIEnv<'_>, usk: jbyteArray) -> Result<UnifiedSpendingKey, failure::Error> {
+fn decode_usk(env: &JNIEnv, usk: JByteArray) -> Result<UnifiedSpendingKey, failure::Error> {
```
daira, line 216: It's kind-of an implementation detail that this doesn't need to call `JNIEnv` methods that take `&mut self`. But since it's not public, that doesn't matter I guess.

backend-lib/Cargo.lock
str4d: They have already bumped to `windows-sys 0.48` in their development branch, so this will likely be resolved in the next release (unless they take the opportunity to jump to `windows-sys 0.52` in which case we'll have the reverse issue).

str4d: Actually I've just checked, and all of our other dependencies will switch to `windows-sys 0.52` with a `cargo update`. So `jni` will hopefully merge https://github.com/jni-rs/jni-rs/pull/513 before their next release.

backend-lib/src/main/rust/lib.rs
```diff
@@ -101,9 +101,9 @@ fn account_id_from_jint(account: jint) -> Result<AccountId, failure::Error> {
 }
 
 #[no_mangle]
-pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_initOnLoad(
-    _env: JNIEnv<'_>,
-    _: JClass<'_>,
+pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_initOnLoad<'local>(
```
daira, line 104: The `jni` crate docs use `extern "system"`, which is correct.

On Unix it will work with `extern "C"`, but that's wrong in general. Specifically, on Win32 `JNICALL` declares the `stdcall` convention, not the `C` convention. `extern "system"` works on all platforms.

I know Windows is currently a Tier 3 platform. Nevertheless, please replace all `extern "C"`s for JNI native method definitions with `extern "system"`. This was already incorrect for `jni 0.20` so it could be done in a separate PR if that's easier.

backend-lib/src/main/rust/utils/exception.rs
```diff
@@ -52,7 +52,7 @@ pub fn unwrap_exc_or<T>(env: &JNIEnv, res: ExceptionResult<T>, error_val: T) ->
```
daira, line 52: Delete the commented-out code or change it to use `env: &mut JNIEnv` so that it would compile. (Probably delete it: writing `unwrap_exc_or_default` is not much more concise than just passing the default, and I prefer the explicitness of the latter.)

backend-lib/src/main/rust/lib.rs
```diff
@@ -939,57 +969,57 @@ pub unsafe extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_re
...
 }
 
 fn decode_sapling_subtree_root(
-    env: &JNIEnv<'_>,
-    obj: JObject<'_>,
+    env: &mut JNIEnv,
+    obj: JObject,
 ) -> Result<CommitmentTreeRoot<sapling::Node>, failure::Error> {
-    let long_as_u32 = |name| -> Result<u32, failure::Error> {
+    fn long_as_u32(env: &mut JNIEnv, obj: &JObject, name: &str) -> Result<u32, failure::Error> {
```
daira, line 979: Perhaps these helpers should be in `utils`, since they're used more than once?

daira (approved):
utACK with comments. If the calling convention isn't fixed in this PR, please open an issue to do so.

backend-lib/src/main/rust/lib.rs
str4d: There are now Android Windows targets. The host is irrelevant here; what matters is the target being compiled for, which per the available Android targets and the definition of `system` means that it will always be equivalent to `C`.

HonzaR (approved):
I've just tested these changes with the basic demo app use cases. All worked as expected.

backend-lib/src/main/rust/lib.rs
str4d: Opened #1358.

PR #1495, opened 2024-06-04
ISSUE-1253 Transaction composable implementation

Milan-Nerevar:
Snímka obrazovky 2024-06-04 o 16 33 40

### DIFF
--- a/demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/ui/screen/transactions/view/TransactionsView.kt
+++ b/demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/ui/screen/transactions/view/TransactionsView.kt
@@ -1,18 +1,16 @@
 package cash.z.ecc.android.sdk.demoapp.ui.screen.transactions.view
 
 import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyColumn
-import androidx.compose.foundation.lazy.items
+import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.filled.ArrowBack
-import androidx.compose.material.icons.filled.ArrowBack
 import androidx.compose.material.icons.outlined.Autorenew
-import androidx.compose.material3.Button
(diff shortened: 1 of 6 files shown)


demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/ui/screen/transactions/view/TransactionsView.kt
```diff
@@ -23,63 +21,91 @@ import androidx.compose.runtime.Composable
...
-    MaterialTheme {
-        // TODO [#1090]: Demo: Add Addresses and Transactions Compose Previews
-        // TODO [#1090]: https://github.com/zcash/zcash-android-wallet-sdk/issues/1090
-        // TransactionsView()
+fun Transactions(synchronizer: Synchronizer, onBack: () -> Unit) {
+    val queryScope = rememberCoroutineScope()
+
+    // TODO [#846]: Slow addresses providing
+    // TODO [#846]: https://github.com/zcash/zcash-android-wallet-sdk/issues/846
+    val walletAddresses: WalletAddresses? = flow<WalletAddresses?> {
```
Milan-Nerevar, line 46: this should be done in VM, never in UI too, but just copied the logic

demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/ui/screen/transactions/view/TransactionsView.kt
```diff
@@ -23,63 +21,91 @@ import androidx.compose.runtime.Composable
...
+
+    val transactions = if (walletAddresses == null) {
+        emptyList()
+    } else {
+        synchronizer
+            .transactions
+            .collectAsStateWithLifecycle(initialValue = emptyList())
+            .value
+            .map { transactionOverview ->
+                transactionOverview.toTransactionState(
```
Milan-Nerevar, line 58: mapping should never be done in UI but in VM instead

HonzaR (commented):
Thank you for these changes. They bring a logic that enables us to extend it later easily.

I agree with all three of your comments inline. It's good to point this out. The Demo app itself is in the shadow of the Zashi wallet from the separate repository. That is the showcase project that the other wallet creates should follow. The Demo app is a testing tool for the SDK APIs. That's why the codebase itself does not necessarily need to follow all the best-practices. We can file this as a potential follow-up and improve it later.

Would you like to sync it with the original SDK repo state (we pushed some commits, I think)? And possibly resolve the CI actions failures.

HonzaR (approved):
This is a good improvement in our Demo app transaction list, which we can build upon quickly later. Thank you.

PR #1532, opened 2024-07-20
ZIP 320 (TEX address) support

str4d:
> **Note**
>This code review checklist is intended to serve as a starting point for the author and reviewer, although it may not be appropriate for all types of changes (e.g. fixing a spelling typo in documentation).  For more in-depth discussion of how we think about code review, please see [Code Review Guidelines](../blob/main/docs/CODE_REVIEW_GUIDELINES.md).

# Author

- [x] **Self-review** your own code in GitHub's web interface[^1]
- [ ] Add **automated tests** as appropriate
- [ ] Update the [**manual tests**](../blob/main/docs/testing/manual_testing)[^2] as appropriate
- [ ] Check the **code coverage**[^3] report for the automated tests
- [ ] Update **documentation** as appropriate (e.g [README.md](../blob/main/README.md), [Architecture.md](../blob/main/docs/Architecture.md), etc.)
- [x] **Run the demo app** and try the changes
- [ ] Pull in the latest changes from the **main** branch and **squash** your commits before assigning a reviewer[^4]

# Reviewer

- [ ] Check the code with the [Code Review Guidelines](../blob/main/docs/CODE_REVIEW_GUIDELINES.md) **checklist**
- [ ] Perform an **ad hoc review**[^5]
- [ ] Review the **automated tests**
- [ ] Review the **manual tests**
- [ ] Review the **documentation**, [README.md](../blob/main/README.md), [Architecture.md](/blob/main/docs/Architecture.md), etc. as appropriate
- [ ] **Run the demo app** and try the changes[^6]

[^1]: _Code often looks different when reviewing the diff in a browser, making it easier to spot potential bugs._
[^2]: _While we aim for automated testing of the SDK, some aspects require manual testing. If you had to manually test 
something during development of this pull request, write those steps down._
[^3]: _While we are not looking for perfect coverage, the tool can point out potential cases that have been missed. Code coverage can be generated with: `./gradlew check` for Kotlin modules and `./gradlew connectedCheck -PIS_ANDROID_INSTRUMENTATION_TEST_COVERAGE_ENABLED=true` for Android modules._
[^4]: _Having your code up to date and squashed will make it easier for others to review. Use best judgement when squashing commits, as some changes (such as refactoring) might be easier to review as a separate commit._
[^5]: _In addition to a first pass using the code review guidelines, do a second pass using your best judgement and experience which may identify additional questions or comments. Research shows that code review is most effective when done in multiple passes, where reviewers look for different things through each pass._
[...]

PR #1533, opened 2024-07-20
Add `Synchronizer` APIs for USD/ZEC exchange rate

str4d:
> **Note**

# Author

# Reviewer

### DIFF
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/TorClient.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/TorClient.kt
@@ -1,0 +1,48 @@
+package cash.z.ecc.android.sdk.internal.model
+
+import cash.z.ecc.android.sdk.internal.ext.existsSuspend
+import cash.z.ecc.android.sdk.internal.ext.mkdirsSuspend
+import cash.z.ecc.android.sdk.internal.jni.RustBackend
+import dalvik.annotation.optimization.CriticalNative
+import java.io.File
+import java.math.BigDecimal
+
+class TorClient(
+    private val nativeHandle: Long?,
+) {
+    fun dispose() {
+        nativeHandle?.let { freeTorRuntime(it) }
+    }
+
+    fun getExchangeRateUsd(): BigDecimal {
+        return getExchangeRateUsd(nativeHandle!!)
+    }
+
+    companion object {
+        suspend fun new(torDir: File): TorClient {
+            RustBackend.loadLibrary()
+
+            // Ensure that the directory exists.
+            torDir.mkdirsSuspend()
+            if (!torDir.existsSuspend()) {
+                error("${torDir.path} directory does not exist and could not be created.")
+            }
+
+            return TorClient(createTorRuntime(torDir.path))
+        }
+
+        //
+        // External Functions
+        //
+
+        @JvmStatic
+        private external fun createTorRuntime(torDir: String): Long
+
+        @CriticalNative
+        @JvmStatic
+        private external fun freeTorRuntime(nativeHandle: Long)
+
+        @JvmStatic
+        private external fun getExchangeRateUsd(nativeHandle: Long): BigDecimal
+    }
+}
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/TorClient.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/TorClient.kt
@@ -4,32 +4,37 @@ import cash.z.ecc.android.sdk.internal.ext.existsSuspend
 import cash.z.ecc.android.sdk.internal.ext.mkdirsSuspend
 import cash.z.ecc.android.sdk.internal.jni.RustBackend
 import dalvik.annotation.optimization.CriticalNative
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 import java.io.File
 import java.math.BigDecimal
 
 class TorClient(
     private val nativeHandle: Long?,
 ) {
-    fun dispose() {
-        nativeHandle?.let { freeTorRuntime(it) }
-    }
+    suspend fun dispose() =
+        withContext(Dispatchers.IO) {
+            nativeHandle?.let { freeTorRuntime(it) }
+        }
 
-    fun getExchangeRateUsd(): BigDecimal {
-        return getExchangeRateUsd(nativeHandle!!)
-    }
+    suspend fun getExchangeRateUsd(): BigDecimal =
+        withContext(Dispatchers.IO) {
+            getExchangeRateUsd(nativeHandle!!)
+        }
 
     companion object {
-        suspend fun new(torDir: File): TorClient {
-            RustBackend.loadLibrary()
+        suspend fun new(torDir: File): TorClient =
+            withContext(Dispatchers.IO) {
+                RustBackend.loadLibrary()
 
-            // Ensure that the directory exists.
-            torDir.mkdirsSuspend()
-            if (!torDir.existsSuspend()) {
-                error("${torDir.path} directory does not exist and could not be created.")
+                // Ensure that the directory exists.
+                torDir.mkdirsSuspend()
+                if (!torDir.existsSuspend()) {
+                    error("${torDir.path} directory does not exist and could not be created.")
+                }
+
+                TorClient(createTorRuntime(torDir.path))
             }
-
-            return TorClient(createTorRuntime(torDir.path))
-        }
 
         //
         // External Functions
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/TorClient.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/TorClient.kt
@@ -9,7 +9,7 @@ import kotlinx.coroutines.withContext
 import java.io.File
 import java.math.BigDecimal
 
-class TorClient(
+class TorClient private constructor(
     private val nativeHandle: Long?,
 ) {
     suspend fun dispose() =
@@ -40,14 +40,22 @@ class TorClient(
         // External Functions
         //
 
+        /**
+         * @throws RuntimeException as a common indicator of the operation failure
+         */
         @JvmStatic
+        @Throws(RuntimeException::class)
         private external fun createTorRuntime(torDir: String): Long
 
         @CriticalNative
         @JvmStatic
         private external fun freeTorRuntime(nativeHandle: Long)
 
+        /**
+         * @throws RuntimeException as a common indicator of the operation failure
+         */
         @JvmStatic
+        @Throws(RuntimeException::class)
         private external fun getExchangeRateUsd(nativeHandle: Long): BigDecimal
     }
 }
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -1535,7 +1535,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_putUtxo<'
                     .map_err(|_| anyhow!("Invalid UTXO value"))?,
                 script_pubkey,
             },
-            BlockHeight::from(height as u32),
+            Some(BlockHeight::from(height as u32)),
         )
         .ok_or_else(|| anyhow!("UTXO is not P2PKH or P2SH"))?;
 
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -14,6 +14,7 @@ use jni::{
 };
 use prost::Message;
 use secrecy::{ExposeSecret, SecretVec};
+use tor_rtcompat::BlockOn;
 use tracing::{debug, error};
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::reload;
@@ -34,6 +35,7 @@ use zcash_client_backend::{
     fees::{standard::SingleOutputChangeStrategy, DustOutputPolicy},
     keys::{DecodingError, Era, UnifiedAddressRequest, UnifiedFullViewingKey, UnifiedSpendingKey},
     proto::{proposal::Proposal, service::TreeState},
+    tor::http::cryptex,
     wallet::{NoteId, OvkPolicy, WalletTransparentOutput},
     zip321::{Payment, TransactionRequest},
     ShieldedProtocol,
@@ -64,6 +66,7 @@ use zcash_proofs::prover::LocalTxProver;
 
 use crate::utils::{catch_unwind, exception::unwrap_exc_or};
 
+mod tor;
 mod utils;
 
 const ANCHOR_OFFSET_U32: u32 = 10;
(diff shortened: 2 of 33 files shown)


str4d:
[Rebased on `feature-2.2.0`](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/c7845e45ce2fd9aa5be649db8347d2560f0697b6..56949983aec351b3847599230e113759bfec52a2) to bring in Kotlin changes from #1532. I also migrated to the latest revision of the Rust crates, which includes bugfixes to ZIP 320 support.

HonzaR:
Rebased on `feature-2.0.0` and resolved conflicts in Changelog.

docs/Setup.md
```diff
@@ -15,12 +15,12 @@ Start by making sure the command line with Gradle works first, because **all the
...
     1. Android Studio has an embedded JVM, although running Gradle tasks from the command line requires a separate JVM to be installed.  Our Gradle scripts are configured to use toolchains to automatically install the correct JVM version.
 1. Configure Rust
-    1. [Install Rust](https://www.rust-lang.org/learn/get-started). You will need Rust 1.60 to 1.67.1.  Older and newer versions of Rust will fail.  If you use system packages, check the provided version.
+    1. [Install Rust](https://www.rust-lang.org/learn/get-started). You will need at least Rust 1.80.  Older versions of Rust will fail.  If you use system packages, check the provided version.
```
HonzaR, line 18: Thank you for updating the documentation as well!

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -189,12 +195,14 @@ class SdkSynchronizer private constructor(
     }
 
     private val _status = MutableStateFlow<Synchronizer.Status>(DISCONNECTED)
+    private val _exchangeRateUsd = MutableStateFlow<Pair<BigDecimal, Instant>?>(null)
```
HonzaR, line 198: The same comment as above for the `Pair` usage.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -417,6 +425,30 @@ class SdkSynchronizer private constructor(
         processor.refreshWalletSummary()
     }
 
+    override suspend fun refreshExchangeRateUsd() {
```
HonzaR, line 428: We should update the Javadoc in the parent `Synchronizer` to reflect the fact that this function may produce exceptions. Ideally, we throw a custom exception, as commented above.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/TorClient.kt
```diff
@@ -0,0 +1,48 @@
+package cash.z.ecc.android.sdk.internal.model
+
+import cash.z.ecc.android.sdk.internal.ext.existsSuspend
+import cash.z.ecc.android.sdk.internal.ext.mkdirsSuspend
+import cash.z.ecc.android.sdk.internal.jni.RustBackend
+import dalvik.annotation.optimization.CriticalNative
+import java.io.File
+import java.math.BigDecimal
+
+class TorClient(
```
HonzaR, line 10: We could have private constructor here.

HonzaR (commented):
I left a few comments inline. @str4d One question is related to the Rust FFI. We'll address the other comments in the connected `exchange-rate-implementation` branch.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
str4d: Fine with me (this type is entirely Kotlin-side, I just did what was sufficient to get this working).

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/TorClient.kt
```diff
@@ -0,0 +1,48 @@
...
+
+        @JvmStatic
+        private external fun createTorRuntime(torDir: String): Long
+
+        @CriticalNative
+        @JvmStatic
+        private external fun freeTorRuntime(nativeHandle: Long)
+
+        @JvmStatic
+        private external fun getExchangeRateUsd(nativeHandle: Long): BigDecimal
```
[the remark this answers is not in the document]
str4d: Any FFI method that ends with some variant of `unwrap_exc_or` can result in a `RuntimeException`. In this case, `freeTorRuntime` is the only FFI method that won't throw an exception (at least from the Rust side; I'm making no guarantees about the JNI internals, but I suspect any issue there is definitionally unexpected).

str4d:
I just realised that this PR was also rebased onto the invalid feature branch state, which is now conflicting very badly after the feature branch state was fixed. I'll rebase this branch to fix.

str4d:
[Rebased on the now-fixed state of `feature-2.2.0`](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/0a67f36a9f3eb8214e0fb76007396d8f39a2fbd9..97b8d1971a3a78808aa1703807d6e6edd35b647d) to undo the merge conflicts.

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/97b8d1971a3a78808aa1703807d6e6edd35b647d..f9307d5ced09b05794fefcc7e261b47b8250f85d) to fix up a changelog entry that was incorrectly being added to the 2.1.2 section.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/exchange/UsdExchangeRateFetcher.kt
```diff
@@ -0,0 +1,68 @@
...
+                    Twig.info { "[USD] Fetch start" }
+                    tor.getExchangeRateUsd()
+                } catch (e: Exception) {
+                    Twig.error(e) { "[USD] Fetch failed" }
+                    return@retry FetchFiatCurrencyResult.Error(e)
+                } finally {
+                    tor.dispose()
+                }
+
+            Twig.debug { "[USD] Fetch success: $rate" }
```
str4d, line 35: If an error occurs in `tor.getExchangeRateUsd()`, the retry logic will discard the `tor` client and create a new one with `TorClient.new`. This is going to noticeably slow down the retry logic, as on each retry it is forcing the Tor client to:
- serialize its state
- shut down its circuits
- load it state
- build new circuits, which can take several seconds

This should instead have separate retry logic for `TorClient.new` (trying a few times to create the Tor client) and `tor.getExchangeRateUsd()` (trying a few times in case a particular Tor request failed, but reusing the already-running Tor client).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -426,27 +468,9 @@ class SdkSynchronizer private constructor(
...
-                throw e
-            } finally {
-                tor.dispose()
-            }
-        Twig.info { "Latest USD/ZEC exchange rate is $rate" }
-
-        _exchangeRateUsd.value = Pair(rate, Instant.now())
+        coroutineScope.launch {
+            refreshExchangeRateUsd.emit(Unit)
+        }
```
str4d, line 473: I'm now confused by what this does. Is `refreshExchangeRateUsd()` no longer a manual refresh action, but instead kicks off a stream of automated refreshes (which would explain why errors are no longer raised from here)? If that is the case, then the method name and documentation is incorrect, and should have been altered in #1537 (to something like `startAutomaticRefreshOfExchangeRateUsd`).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -194,15 +202,49 @@ class SdkSynchronizer private constructor(
...
+    override val exchangeRateUsd =
+        channelFlow {
+            var lastValue = ObserveFiatCurrencyResult()
+            refreshExchangeRateUsd
+                .flatMapLatest {
+                    flow {
+                        emit(lastValue.copy(isLoading = true))
+                        lastValue =
+                            when (val result = fetchExchangeChangeUsd()) {
+                                is FetchFiatCurrencyResult.Error -> lastValue.copy(isLoading = false)
```
str4d, line 225: Is this implementing a partial copy? i.e. if we have older valid data, and then a refresh ends in an error, we preserve that older data (so queries to the `StateFlow` still return an exchange rate)?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -194,15 +202,49 @@ class SdkSynchronizer private constructor(
...
+
+                                is FetchFiatCurrencyResult.Success ->
+                                    lastValue.copy(
+                                        isLoading = false,
+                                        currencyConversion = result.currencyConversion
+                                    )
```
str4d, line 231: And then this is where we update older valid data (or initialize the data after the first successful refresh)?

str4d (commented):
Reviewed f9307d5ced09b05794fefcc7e261b47b8250f85d.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1843,6 +1843,79 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_branchIdF
...
+    _: JClass<'local>,
+    tor_dir: JString<'local>,
+) -> jlong {
+    let res = catch_unwind(&mut env, |env| {
+        let tor_dir = utils::java_string_to_rust(env, &tor_dir);
+        let tor_dir = Path::new(&tor_dir);
+
+        let tor = crate::tor::TorRuntime::create(tor_dir)?;
+
+        Ok((Box::into_raw(Box::new(tor)) as usize) as jlong)
```
daira, line 1863:
```suggestion
        let ptr = Box::into_raw(Box::new(tor)) as usize;
        assert_eq!((ptr as jlong) as usize, ptr);
        Ok(ptr as jlong)
```

sdk-incubator-lib/src/main/java/cash/z/ecc/android/sdk/model/FiatCurrencyConversionExt.kt
```diff
@@ -0,0 +1,60 @@
...
+        if (monetarySeparators.isGroupingValid()) {
+            "#${monetarySeparators.grouping}##0${monetarySeparators.decimal}0#"
+        } else {
+            "0${monetarySeparators.decimal}0#"
+        }
+
+    val decimalFormat =
+        DecimalFormat(localizedPattern, symbols).apply {
+            isParseBigDecimal = true
+            roundingMode = RoundingMode.HALF_EVEN // aka Bankers rounding
```
daira, line 40:
```suggestion
            roundingMode = RoundingMode.HALF_EVEN // a.k.a. Banker's rounding
```

daira (approved):
utACK modulo @str4d's comments, and with some minor suggestions.

Milan-Cerovsky (approved):

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/exchange/UsdExchangeRateFetcher.kt
Milan-Cerovsky: good catch

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
Milan-Cerovsky: I see. It still is what the name says: `refreshExchangeRateUsd`. When it fires it signalizes a call to refresh then `exchangeRateUsd`.

How it works:
`refreshExchangeRateUsd` is observed by `exchangeRateUsd` but there is a catch.
`private val refreshExchangeRateUsd = MutableSharedFlow<Unit>(replay = 1)` is a hot flow with a replay cache of 1, but this time it's empty. Because it is empty observing`refreshExchangeRateUsd.flatMapLatest` would not fire.
This is solved by:
`private val refreshExchangeRateUsd = MutableSharedFlow<Unit>(replay = 1).apply { tryEmit(Unit) }`

Now according to documentation:
```
public inline fun <T, R> Flow<T>.flatMapLatest(
    @BuilderInference crossinline transform: suspend (value: T) -> Flow<R>
): Flow<R>

Returns a flow that switches to a new flow produced by transform function every time the original flow emits a value. When the original flow emits a new value, the previous flow produced by transform block is cancelled.
```

So making `refreshExchangeRateUsd` fire again in effect makes the suspending function UsdExchangeRateFetcher.invoke be killed because we are calling it within observing it via`flatMapLatest`.

Also another welcome and intended change is: You can just start collecting and do not have to call `refreshExchangeRateUsd()` manually in order for the tor fetcher to be called. It happens automatically.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
Milan-Cerovsky: exactly

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
Milan-Cerovsky: yes exactly :) 

We always show latest data and let the client decide whether it's stale or not and how to handle it. That means when we encounter an error during fetch we do not invalidate the previous data.

HonzaR (commented):
On the tex addresses: I re-tested sending funds to a tex addr from Binance. I can now confirm the same behavior as on iOS: the first - shielding one successes, the second - sending one fails after expiration.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
str4d: Mmk, so the effect of this change is:
- On first subscribe to `exchangeRateUsd`, a refresh occurs automatically.
- All subsequent refreshes are individually triggered by `refreshExchangeRateUsd`

That seems fine to me.

str4d:
@HonzaR is that just what you see in the UI, or did you also confirm what happened to the actual transactions? The iOS issue is that the second transaction shows up in the UI as expired, but is actually mined.

HonzaR:
> @HonzaR is that just what you see in the UI, or did you also confirm that this happened to the actual transactions? The iOS issue is that the second transaction shows up in the UI as expired, but is actually mined.

It's just about UI. I can confirm that the transaction was mined and that I received the funds at the tex address.

str4d:
I'm confused about @Milan-Cerovsky merging `main` into this branch instead of the `feature-2.2.0 branch`, as it now makes this PR harder to review. But also this PR has been hanging around way longer than I wanted it to, and AFAICT all the necessary changes for the USD/ZEC feature have been implemented and reviewed. So I'm going to merge this PR into the feature branch; any remaining changes should be made as individual PRs targeting the feature branch.

Issue #1555, opened 2024-08-07
Keep `BLOCKS_METADATA_BUFFER_SIZE` the same as current `SYNC_BATCH_SIZE`


HonzaR:
## Is your feature request related to a problem? Please describe.
We'd like to follow up on this [issue](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/issues/1364), where we defined the ideal situation - using `BLOCKS_METADATA_BUFFER_SIZE` the same size as the sync batch. We use `SYNC_BATCH_SIZE` and then `SYNC_BATCH_SMALL_SIZE` for the sandblasting period, so the `BLOCKS_METADATA_BUFFER_SIZE` should reflect this.

Currently, we only raised the `BLOCKS_METADATA_BUFFER_SIZE` to 1000, which brings great sync machine speed up (about 50%) and still works well for the sandblasting period.

PR #1548, opened 2024-08-05
[#1365] Raise sync batch size outside of sandblasting to 1000

HonzaR:
Closes #1365

Sync speed results
---
**New wallet (cca 22K blocks):**

_Before_:
1. 14:15:53.199 - 14:09:02.966 = 6 minutes, 50 seconds, and 233
2. 14:24:15.356 - 14:18:32.068 = 5 minutes, 43 seconds, and 288

_After_:
1. 14:01:34.786 - 13:55:54.256 = 5 minutes, 40 seconds, and 530
2. 15:18:51.461 - 15:14:26.988 = 4 minut, 24 sekund a 473

**Wallet restoration (cca 117K blocks):**

_Before:_
16:32:09.854 - 16:02:00.592 = 30 minut, 9 sekund a 262

_After:_
15:51:07.254 - 15:22:28.317 = 28 minut, 38 sekund a 937

> **Note**
>This code review checklist is intended to serve as a starting point for the author and reviewer, although it may not be appropriate for all types of changes (e.g. fixing a spelling typo in documentation).  For more in-depth discussion of how we think about code review, please see [Code Review Guidelines](../blob/main/docs/CODE_REVIEW_GUIDELINES.md).

# Author

- [x] **Self-review** your own code in GitHub's web interface[^1]
- [ ] Add **automated tests** as appropriate
- [ ] Update the [**manual tests**](../blob/main/docs/testing/manual_testing)[^2] as appropriate
- [ ] Check the **code coverage**[^3] report for the automated tests
- [x] Update **documentation** as appropriate (e.g [README.md](../blob/main/README.md), [Architecture.md](../blob/main/docs/Architecture.md), etc.)
- [x] **Run the demo app** and try the changes
- [x] Pull in the latest changes from the **main** branch and **squash** your commits before assigning a reviewer[^4]

# Reviewer

- [x] Check the code with the [Code Review Guidelines](../blob/main/docs/CODE_REVIEW_GUIDELINES.md) **checklist**
- [ ] Perform an **ad hoc review**[^5]
- [ ] Review the **automated tests**
- [ ] Review the **manual tests**
- [x] Review the **documentation**, [README.md](../blob/main/README.md), [Architecture.md](/blob/main/docs/Architecture.md), etc. as appropriate
- [ ] **Run the demo app** and try the changes[^6]

[^1]: _Code often looks different when reviewing the diff in a browser, making it easier to spot potential bugs._
[...]

PR #1556, opened 2024-08-07
[#1364] Increase BLOCKS_METADATA_BUFFER_SIZE

### LINKED ISSUE #1364: Increase `BLOCKS_METADATA_BUFFER_SIZE` to match (or be closer to) download batch size
The block downloader writes the blocks to disk and caches their metadata. We then flush the block metadata buffer to the Rust side to make the cache aware of the downloaded blocks.

Currently we flush the buffer when we reach 10 blocks of metadata. For the current download batch size of 100, this means we call `RustBackend.writeBlockMetadata` 10 times per batch. Rust FFI calls are globally synchronized though the `zc-io` thread, and the buffer flushes block the writes to disk in `FileCompactBlockRepository.write()`. This means that if a download batch is running concurrently with the scanning of an earlier batch (as can happen due to buffering in `CompactBlockProcessor.runSyncingAndEnhancingOnRange()`), the network I/O blocks on the Rust layer scanning.

Instead, we should flush the buffer once the entire batch is written to disk. There is currently no benefit to early flushing (as we don't start scanning until the entire batch is downloaded), and if the download fails and causes the synchronizer to stop, then the un-flushed buffer doesn't matter because on restart all downloaded blocks are cleared.

Alternatively, we could decouple the network I/O from the `RustBackend.writeBlockMetadata` call, so that blocks continue to be downloaded and written to disk during the Rust FFI call.

HonzaR:
- Closes #1364
- Filed another related improvement issue #1555
- The implemented changes still work well for the sandblasting period, too, where we sync batch of 100 blocks.
- Changelog update

Block synchronization tests results:
---
**Restore older wallet (122K blocks):**
_Before:_
10:04:18.586 - 09:39:49.527 = 00:24:29
_After:_
10:20:13.480 - 10:07:22.831 = 00:12:50
= **cca -48.5%**

**New wallet: (2.5K blocks):**
_Before:_
10:41:23.064 - 10:40:47.286 = 00:00:35.778
_After:_
10:44:26.194 - 10:44:07.951 = 00:00:18.000000
= **cca -50%** 

> **Note**

# Author

# Reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `onCriticalErrorHandler` in case any error occurs 
 - Block synchronization logic now works above batch of blocks with size 1000 blocks instead of just 100 blocks 
   except the Zcash sandblasting period in which batch size of 100 blocks is still used
+- The internal `FileCompactBlockRepository.BLOCKS_METADATA_BUFFER_SIZE` constant has been raised from 10 to 1000 to 
+  match the block synchronization batch size. The overall speed-up of the entire block synchronization logic is about 
+  50%.
 - Checkpoints update
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,14 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `fullyScannedHeight` as its lower bound.
 - The fetch UTXOs action reports `FetchUtxosException` to the wrapping `onProcessorErrorHandler` or 
   `onCriticalErrorHandler` in case any error occurs 
-- Block synchronization logic now works above batch of blocks with size 1000 blocks instead of just 100 blocks 
-  except the Zcash sandblasting period in which batch size of 100 blocks is still used
+- The internal `CompactBlockProcessor.SYNC_BATCH_SIZE` has changed. Block synchronization logic now works above 
+  batch of blocks with size 1000 blocks instead of just 100 blocks, except the Zcash sandblasting period in which 
+  batch size of 100 blocks is still used.
 - The internal `FileCompactBlockRepository.BLOCKS_METADATA_BUFFER_SIZE` constant has been raised from 10 to 1000 to 
-  match the block synchronization batch size. The overall speed-up of the entire block synchronization logic is about 
-  50%.
+  match the block synchronization batch size. 
+- The overall speed-up of the entire block synchronization logic, thanks to the both mentioned synchronization 
+  improvements above is about 50% out of the Zcash sandblasting period. There is still some improvement in the 
+  sandblasting period.
 - Checkpoints update
 
 ### Fixed
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/storage/block/FileCompactBlockRepository.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/storage/block/FileCompactBlockRepository.kt
@@ -125,10 +125,15 @@ internal class FileCompactBlockRepository(
          */
         const val BLOCK_FILENAME_SUFFIX = "-compactblock"
 
+        // TODO [#1555]: Keep BLOCKS_METADATA_BUFFER_SIZE the same as current SYNC_BATCH_SIZE
+        // TODO [#1555]: https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/issues/1555
+
         /**
          * The size of block meta data buffer
+         *
+         * Note: we should keep this size as close as possible to [CompactBlockProcessor.SYNC_BATCH_SIZE]
          */
-        const val BLOCKS_METADATA_BUFFER_SIZE = 10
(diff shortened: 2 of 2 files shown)


Milan-Cerovsky (approved):

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/storage/block/FileCompactBlockRepository.kt
```diff
@@ -125,10 +125,15 @@ internal class FileCompactBlockRepository(
...
+        // TODO [#1555]: Keep BLOCKS_METADATA_BUFFER_SIZE the same as current SYNC_BATCH_SIZE
+        // TODO [#1555]: https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/issues/1555
+
         /**
          * The size of block meta data buffer
+         *
+         * Note: we should keep this size as close as possible to [CompactBlockProcessor.SYNC_BATCH_SIZE]
          */
-        const val BLOCKS_METADATA_BUFFER_SIZE = 10
+        const val BLOCKS_METADATA_BUFFER_SIZE = 1000
```
str4d, line 136: I was concerned that this might cause problems during sandblasting (note that both tested ranges start after sandblasting), because #1555 is not handled, and we aren't actually scanning all sandblasting blocks. The change to `BLOCKS_METADATA_BUFFER_SIZE` means that this code will never trigger for the 100-block sandblasting ranges:

https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/blob/f4086a179c4724cb85a64909b32297b323aefbff/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/storage/block/FileCompactBlockRepository.kt#L46-L49
https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/blob/f4086a179c4724cb85a64909b32297b323aefbff/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/storage/block/FileCompactBlockRepository.kt#L161-L163

However, it turns out that we are fine, because there is a catch-all that writes out whatever is left in the metadata buffer:

https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/blob/f4086a179c4724cb85a64909b32297b323aefbff/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/storage/block/FileCompactBlockRepository.kt#L52-L55

What this means is that whenever `BLOCKS_METADATA_BUFFER_SIZE` is greater than or equal to the batch size being written, we always have the desired behaviour: write all blocks to disk, then flush the metadata to the Rust side in a single go.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `onCriticalErrorHandler` in case any error occurs 
 - Block synchronization logic now works above batch of blocks with size 1000 blocks instead of just 100 blocks 
   except the Zcash sandblasting period in which batch size of 100 blocks is still used
+- The internal `FileCompactBlockRepository.BLOCKS_METADATA_BUFFER_SIZE` constant has been raised from 10 to 1000 to 
+  match the block synchronization batch size. The overall speed-up of the entire block synchronization logic is about 
+  50%.
```
str4d, line 19: The benchmarks did not test scanning across sandblasting, so we don't know if the 50% number holds up there. I would expect it to still be significant though.

str4d (approved):
utACK 6ad1021bb061713f7dbf333378bca8e0951004fe

CHANGELOG.md
str4d: But I also suspect it will be less than 50% for sandblasting ranges, because I think the 50% we see here is actually partially due to #1548. It increased the batch size outside sandblasting ranges by 10x, which would both improve batch scanning speed and make the IO contention problem worse. My guess is that this PR, by fixing the IO contention problem, is unlocking more of the batch size increase improvement.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/storage/block/FileCompactBlockRepository.kt
HonzaR: Yes, you read this correctly. We're safe in changing BLOCKS_METADATA_BUFFER_SIZE size.

HonzaR:
I just did one more manual test now on the Pixel 8 Pro emulator for a wallet that covers the sandblasting period:

It's Alice's wallet from the SDK fixture values. 
- The after state contains both:
   - SYNC_BATCH=1000 (or 100 in sandblasting)
   - BLOCKS_METADATA_BUFFER_SIZE=1000 for the whole range. 
- The before state:
   - It does not include any of the above-mentioned improvement

2603531 - 1935001 = 668,530 (i.e. cca 115K blocks in the sandblasting period)
———
**Before**:
10:47:10.276 - 08:55:40.430 = 01:51:29
**After**:
10:16:18.025 - 08:55:55.730 = 01:20:22

The result is about 27.90% faster now. I read this as the significant improvement coming mostly out of the sandblasting period, which is expected.

Milan-Cerovsky (approved):

PR #1559, opened 2024-08-10
Integrate transaction status queue for transparent history

str4d:
This PR currently targets `tor-exchange-rate` because it builds on the Rust parts, and #1533 has not yet been merged.

> **Note**

# Author

# Reviewer


sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -1946,7 +1951,10 @@ class CompactBlockProcessor internal constructor(
                 }
                 when (val response = downloader.fetchTransaction(rawTransactionId)) {
                     is Response.Success -> {
-                        transactionDataResult = response.result.data
+                        val currentMinedHeight =
+                            runCatching { response.result.height.toBlockHeight(network) }
```
HonzaR, line 1955: This possibly won't work as intended. The related gRPC layer says about the height parameter: `height that the transaction was mined (or -1)`. So, the exception will be thrown even before this check in parsing the fetch response on line 1953, where the response data is parsed to `RawTransactionUnsafe` in which we have `BlockHeightUnsafe` that has this check included:

```
data class BlockHeightUnsafe(val value: Long) : Comparable<BlockHeightUnsafe> {
    init {
        require(UINT_RANGE.contains(value)) { "Height $value is outside of allowed range $UINT_RANGE" }
    }

    override fun compareTo(other: BlockHeightUnsafe): Int = value.compareTo(other.value)

    companion object {
        private val UINT_RANGE = 0.toLong()..UInt.MAX_VALUE.toLong()
    }
}
```
We'll check the possible -1 for height, but in the wrapping retry logic. We could explicitly check the height parameter for `<= 0` and fill in the height as null in RawTransactionUnsafe instead. What do you think?

str4d:
[Rebased on `feature-2.2.0`](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/c8563fd387d97ba2eb958519b4dcf1a406655523..6c12d04ed4449e5c7e4646dc9d0ff87c0350865b) to bring in latest `main` and squash the fixes for `librustzcash` updates into the first commit.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
str4d: Ah, I didn't realise that `BlockHeightUnsafe` had the same check. That means `BlockHeightUnsafe` is the incorrect type to use here, as there are three values that can be returned:
- 0 if the transaction is in the mempool.
- -1 (which is `0xffffffffffffffff` in the protobuf as it uses `uint64`) if the transaction is mined in a block that is not in the main chain.
- A non-negative height if the block is mined in the main chain.

So we'll need a new type that can specifically handle these semantics.

> The related gRPC layer says about the height parameter: `height that the transaction was mined (or -1)`.

This documentation was completely incorrect, as @nuttycom and I realised last week. The semantics are correctly documented, and bugs fixed, in https://github.com/zcash/lightwalletd/pull/495,

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/6c12d04ed4449e5c7e4646dc9d0ff87c0350865b..ab01835a972dc07620bc9c1c08a9bb63adf52ea8) to update the `librustzcash` revision, and fix `RawTransactionUnsafe` to handle the `height` parameter correctly.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -1946,7 +1952,15 @@ class CompactBlockProcessor internal constructor(
                 }
                 when (val response = downloader.fetchTransaction(rawTransactionId)) {
                     is Response.Success -> {
-                        transactionDataResult = response.result.data
+                        val currentMinedHeight = when (response.result) {
+                            is RawTransactionUnsafe.MainChain -> runCatching {
```
HonzaR, line 1956: @str4d Do we need this check? The only way how this fails is due to the height below the sapling activation height. Do we expect such heights here?

str4d: Eventually we'll want to support heights below Sapling activation. But this specific catch was kept from my earlier iteration of this PR, before you pointed out that very similar checks were present in `BlockHeightUnsafe`. Now that the height field is correctly handled, we maybe don't need this, but it shouldn't hurt.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/TransactionStatus.kt
```diff
@@ -2,8 +2,22 @@ package cash.z.ecc.android.sdk.internal.model
...
-}
\ No newline at end of file
+sealed class TransactionStatus {
+    abstract fun toPrimitiveValue(): Long
+
+    data class Mined(val height: BlockHeight) : TransactionStatus() {
+        override fun toPrimitiveValue() = height.value
+    }
+
+    data object NotInMainChain : TransactionStatus() {
```
str4d, line 12: Aha, `data object` is what I needed. When I initially tried using `sealed class`, I got compiler errors about the parent class having no data, which is why I went for the interface approach (there being no other obvious method to do this described in the Kotlin documentation).

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniTransactionDataRequest.kt
```diff
@@ -0,0 +1,38 @@
+package cash.z.ecc.android.sdk.internal.model
+
+import androidx.annotation.Keep
+import cash.z.ecc.android.sdk.internal.ext.isInUIntRange
+
+/**
+ * Serves as cross layer (Kotlin, Rust) communication class.
+ */
+@Keep
+interface JniTransactionDataRequest {
```
HonzaR, line 10: I'm unsure if we can pass such a construct from FFI via JNI to Kotlin. Should we rather switch to a standard class as we do in the case of the other `Jni` classes? When I run the current logic. I see this error:
```
CompactBlockProcessorException$EnhanceTransactionError$EnhanceTxDataRequestsError: Error while attempting to request transactions data from the Rust backend
Caused by: java.lang.ClassNotFoundException: Didn't find class "cash.z.ecc.android.sdk.internal.model.JniTransactionDataRequest.GetStatus"
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -1869,107 +1912,235 @@ class CompactBlockProcessor internal constructor(
...
+                                    RawTransaction.new(
+                                        rawTransactionUnsafe = rawTransactionUnsafe,
+                                        network = network
+                                    ),
+                                    backend = backend
+                                )
+                            }
+                            is RawTransactionUnsafe.Mempool,
+                            is RawTransactionUnsafe.OrphanedBlock -> {
+                                error("Unsupported transaction type: $rawTransactionUnsafe")
```
HonzaR, line 1953: This is the second thing I'm still unsure of. Should we treat these cases the same as we do in the case of the `enhanceTransaction` function from the line 2050, or is this fine?

HonzaR (commented):
I used both new FFI endpoints. I have two questions inline, though. @str4d, please.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniTransactionDataRequest.kt
HonzaR: Fixed by switching to `sealed class` on the Kotlin side and by changing the path to the nested classes on the Rust FFI side.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
HonzaR: Fixed by treating RawTransactionUnsafe the same in case of the SpendsFromAddress flow

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/TransactionStatus.kt
HonzaR: Yeah, I see. I changed that to `sealed class`.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniTransactionDataRequest.kt
str4d: Yep, I expected this part to fail in this way (I've never used subclasses of interfaces this way, but couldn't get `sealed class` to work myself).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -1869,107 +1915,225 @@ class CompactBlockProcessor internal constructor(
...
+
+        @Throws(EnhanceTxDownloadError::class)
+        private suspend fun getTransparentAddressTransactions(
+            transactionRequest: TransactionDataRequest.SpendsFromAddress,
+            downloader: CompactBlockDownloader,
+        ): List<RawTransactionUnsafe> {
+            val transactionsResult = arrayListOf<RawTransactionUnsafe>()
+            val traceScope = TraceScope("CompactBlockProcessor.getTransparentAddressTransactions")
+
+            if (transactionRequest.endHeight == null) {
```
HonzaR, line 1968: I'm not sure about this check. What should we do when a `TransactionRequest.SpendsFromAddress` with a null endHeight comes? We could use the current chain tip, the end of the current enhancing batch of blocks, or omit the request as we did now.

Milan-Cerovsky (approved):

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
nuttycom: I would not implement this as a function that returns a list; instead, I would implement it in a way that as each transaction is retrieved from the stream, it is sent to the back end for processing.

HonzaR: Ok, I'm going to change it. And when a `TransactionRequest.SpendsFromAddress` with a null `endHeight` comes, should we omit the one particular request, or should we do any of the suggested above?

nuttycom: At least for right now, the end height will always be present, so maybe this doesn't matter for right now, but what I would do if there is no end height is just process the full stream that is returned from the `GetTaddressTxids` call incrementally.

HonzaR: Yes, with the commit https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/pull/1559/commits/94b6579cdf436b680671000c5cafcedd57d24a6c  we switched from list to stream obtained from `GetTaddressTxids`, but if we got `TransactionDataRequest.SpendsFromAddress.endHeight == null` from `backend.transactionDataRequests()` we can't init the stream, so we handle it like [this](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/blob/94b6579cdf436b680671000c5cafcedd57d24a6c/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt#L1926), which should be fine for now.

Milan-Cerovsky (approved):

Issue #1632, opened 2024-11-20
Ensure that the SDK does not assume a default account anywhere


str4d:
- We have an `Account` type that is meant to encapsulate the choice in account, but we also have `Account.DEFAULT`. We need to ensure that the latter is not used anywhere inside the SDK's public APIs.
- We need to ensure that APIs returning data can do so for all accounts, either by returning all account data or having filter / `Account` arguments (as appropriate).

PR #1640, opened 2024-11-22
Switch from ZIP 32 account indices to UUID account identifiers

### LINKED ISSUE #1633: Migrate `Account` type to the new multi-seed-compatible account identifier
Currently `Account` internally uses a ZIP 32 account index, but this will not work with either multiple seeds in the wallet (e.g. both a "daily usage" seed and a Keystone HW seed) or imported viewing keys. Once `zcash_client_sqlite` has been updated to support a new multi-seed-compatible account identifier, we need to migrate the SDK to use it.

str4d:
This PR is opened against `main` as a draft solely so it can be pre-reviewed. Once #1632 is complete and merged, I'll rebase this branch and then retarget it against a feature branch (so we can review and merge it before the backing Rust changes are released).

Closes #1633.

# Author

# Reviewer


str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/compare/de385b259d4ef94b630e00023fa7eb927d90863f..192c5e1219c0f78702c3e83597f266b45f69d525) to rebase on `main` afer #1641 was merged.

HonzaR (commented):
The `encode_usk` function in `lib.rs` still works upon ZIP32 account index, while it needs to switch to the new `account_uuid` logic to align with the updated `JniUnifiedSpendingKey` class on the Kotlin side. @str4d can you please make this change as part of this PR? My current branch `1632-remove-default-account` is built on these changes.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniAccountBalance.kt
```diff
@@ -25,7 +25,7 @@ import androidx.annotation.Keep
 @Keep
 @Suppress("LongParameterList")
 class JniAccountBalance(
-    val account: Int,
+    val account: ByteArray,
```
daira, line 28:
```suggestion
    val accountUuid: ByteArray,
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/DerivationToolExt.kt
```diff
@@ -1,16 +1,15 @@
...
 import cash.z.ecc.android.sdk.model.UnifiedFullViewingKey
 import cash.z.ecc.android.sdk.model.UnifiedSpendingKey
 import cash.z.ecc.android.sdk.model.ZcashNetwork
 
 fun Derivation.deriveUnifiedAddress(
     seed: ByteArray,
     network: ZcashNetwork,
-    account: Account
-): String = deriveUnifiedAddress(seed, network.id, account.value)
+    accountIndex: Int
```
daira, line 11: This is different to Swift, where `Zip32AccountIndex` which wraps a `UInt32` is used.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeDerivationToolImpl.kt
```diff
@@ -21,14 +20,14 @@ internal class TypesafeDerivationToolImpl(private val derivation: Derivation) :
     override suspend fun deriveUnifiedSpendingKey(
         seed: ByteArray,
         network: ZcashNetwork,
-        account: Account
-    ): UnifiedSpendingKey = derivation.deriveUnifiedSpendingKey(seed, network, account)
+        accountIndex: Int
```
daira, line 23: Again, in Swift this uses the `Zip32AccountIndex` wrapper.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
```diff
@@ -1,16 +1,44 @@
...
+ * What this means is that account identifiers are not stable across "wallet recreation
+ * events". Examples of these include:
+ * - Restoring a wallet from a backed-up seed.
+ * - Importing the same viewing key into two different wallet instances.
  *
- * @param value A 0-based account index.  Must be >= 0.
+ * @param accountUuid The account identifier. Must be length 16.
  */
-data class Account(val value: Int) {
+data class Account(val accountUuid: ByteArray) {
```
daira, line 25: Why not call this `AccountUuid` and use it as the type passed to identify an account in other APIs (rather than `ByteArray`)? That would eliminate the need to explicitly require it to be 16 bytes in most places.

HonzaR:
We need to ensure that [this](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/pull/1645#discussion_r1869826778) is fulfilled as part of this PR.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt
```diff
@@ -98,13 +98,35 @@ class RustBackend private constructor(
     ): JniUnifiedSpendingKey {
         return withContext(SdkDispatchers.DATABASE_IO) {
             createAccount(
+                dbDataPath = dataDbFile.absolutePath,
```
nuttycom, line 101: I reordered the parameters here because the database file and the network id are really environment parameters, whereas the account name and key source etc. are the arguments to the logical function that's being executed.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniAccountBalance.kt
HonzaR: Resolved. Thanks for pointing this out.

backend-lib/src/main/rust/lib.rs
```diff
@@ -288,22 +288,22 @@ fn encode_account<'a, P: Parameters>(
...
     env.new_object(
         JNI_ACCOUNT,
-        "([BLjava/lang/String;Ljava/lang/String;Ljava/lang/String;[BI)V",
+        "(Ljava/lang/String;[BJLjava/lang/String;[BLjava/lang/String;)V",
         &[
-            (&env.byte_array_from_slice(account.id().expose_uuid().as_bytes())?).into(),
-            (&ufvk).into(),
             (&account_name).into(),
+            (&env.byte_array_from_slice(account.id().expose_uuid().as_bytes())?).into(),
+            hd_account_index,
```
nuttycom, line 303: Nit: the hd_account_index and seed fingerprint go together, as they're the two necessary bits of HD derivation information.

demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/Constants.kt
```diff
@@ -10,6 +10,8 @@ val ANDROID_STATE_FLOW_TIMEOUT = 5.seconds
  */
 const val MINIMAL_WEIGHT = 0.0001f
 
-// TODO [#1644]: Refactor Account ZIP32 index across SDK
-// TODO [#1644]: https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/issues/1644
+/**
+ * Until we support full multi-account feature in Demo app we use this constant as a single source of truth for
+ * account selection
+ */
```
nuttycom, line 16: This should no longer be necessary? Is it even still possible to query by ZIP 32 account index?

nuttycom: Never mind, saw that this is the demo app.

demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/ui/screen/home/viewmodel/WalletViewModel.kt
```diff
@@ -104,11 +105,13 @@ class WalletViewModel(application: Application) : AndroidViewModel(application)
                     withContext(Dispatchers.IO) {
                         Mnemonics.MnemonicCode(it.seedPhrase.joinToString()).toSeed()
                     }
+                val accountIndex = getCurrentAccount().hdAccountIndex?.toInt()
+                    ?: error("The account must be set at this point")
```
nuttycom, line 109: Ideally the spending key would only be derived right before it is used for spending; it would not be constructed ahead of time.

nuttycom: In support of that, instead of this being an error, I would have the whole spending key be a nullable value and set to null if the hd account index is not available. But maybe that will change with multi-account support instead?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/DerivationToolExt.kt
```diff
@@ -20,8 +19,15 @@ fun Derivation.deriveUnifiedAddress(
...
     seed: ByteArray,
     network: ZcashNetwork,
-    account: Account
-): UnifiedSpendingKey = UnifiedSpendingKey(deriveUnifiedSpendingKey(seed, network.id, account.value))
+    accountIndex: Int
+): UnifiedSpendingKey =
+    UnifiedSpendingKey(
+        JniUnifiedSpendingKey(
+            // fixme
+            accountUuid = byteArrayOf(),
```
daira, line 27: This needs to be finished.

daira (commented):
Unfinished but looks good so far. Please add a `Zip32AccountIndex` class to replace `Int` for ZIP 32 account indices.

backend-lib/src/main/rust/lib.rs
```diff
@@ -438,17 +421,128 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_createAcc
...
+/// Tells the wallet to track an account using a unified full viewing key.
+///
+/// Returns details about the imported account, including the unique account identifier for
+/// the newly-created wallet database entry. Unlike the other account creation APIs
+/// ([`Self::create_account`] and [`Self::import_account_hd`]), no spending key is returned
+/// because the wallet has no information about how the UFVK was derived.
+///
+/// Certain optimizations are possible for accounts which will never be used to spend funds. If
+/// `spending_key_available` is `false`, the wallet may choose to optimize for this case, in
+/// which case any attempt to spend funds from the account will result in an error.
```
daira, line 450: It's unclear whether this should apply to the hardware wallet case. Does `spending_key_available` mean:
1. That the spending key is unavailable to the SDK (but the SDK might still have to track other things necessary to make proofs, for example), or
2. That the account will never be spent from, and so no spend-related information needs to be tracked?

From the description in this paragraph I would assume 2, but from the name `spending_key_available` I would assume 1.

nuttycom:
This is really used to answer the “should the wallet track spending
metadata” question.

backend-lib/src/main/rust/lib.rs
```diff
@@ -438,17 +421,128 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_createAcc
...
+            .zip(hd_account_index)
+            .map(|(seed_fp, idx)| Zip32Derivation::new(seed_fp, idx));
+
+        let purpose = match purpose {
+            0 => Ok(AccountPurpose::Spending { derivation }),
+            1 => Ok(AccountPurpose::ViewOnly),
+            _ => Err(anyhow!(
+                "Account purpose must be either 0 (Spending) or 1 (ViewOnly)"
+            )),
+        }?;
```
daira, line 533: `seed_fingerprint`, `hd_account_index`, and `derivation` are only used if `purpose` is 0. Is the following suggestion equivalent, or are we relying on deriving those values for error checking?
```suggestion
        let purpose = match purpose {
            0 => {
                let seed_fingerprint =
                    <[u8; 32]>::try_from(&env.convert_byte_array(seed_fingerprint_bytes)?[..])
                    .ok()
                    .map(SeedFingerprint::from_bytes);
                let hd_account_index = zip32::AccountId::try_from(hd_account_index_raw).ok();

                let derivation = seed_fingerprint
                    .zip(hd_account_index)
                    .map(|(seed_fp, idx)| Zip32Derivation::new(seed_fp, idx));

                Ok(AccountPurpose::Spending { derivation })
            },
            1 => Ok(AccountPurpose::ViewOnly),
            _ => Err(anyhow!(
                "Account purpose must be either 0 (Spending) or 1 (ViewOnly)"
            )),
        }?;
```

backend-lib/src/main/rust/lib.rs
```diff
@@ -438,17 +421,128 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_createAcc
...
+///   birthday height.
+/// - `recover_until`: An optional height at which the wallet should exit "recovery mode". In
+///    order to avoid confusing shifts in wallet balance and spendability that may temporarily be
+///    visible to a user during the process of recovering from seed, wallets may optionally set a
+///    "recover until" height. The wallet is considered to be in "recovery mode" until there
+///    exist no unscanned ranges between the wallet's birthday height and the provided
+///   `recover_until` height, exclusive.
+/// - `key_source`: A string identifier or other metadata describing the source of the seed.
+///   This is treated as opaque metadata by the wallet backend; it is provided for use by
+///   applications which need to track additional identifying information for an account.
```
daira, line 469:
```suggestion
/// - `key_source`: A string identifier or other metadata describing the source of the seed.
///   This is treated as opaque metadata by the wallet backend; it is provided for use by
///   applications which need to track additional identifying information for an account.
/// - `ufvk_str`: The UFVK used to detect transactions involving the account.
/// - `treestate`: The tree state corresponding to the last block prior to the wallet's
///   birthday height.
/// - `recover_until`: An optional height at which the wallet should exit "recovery mode". In
///    order to avoid confusing shifts in wallet balance and spendability that may temporarily be
///    visible to a user during the process of recovering from seed, wallets may optionally set a
///    "recover until" height. The wallet is considered to be in "recovery mode" until there
///    exist no unscanned ranges between the wallet's birthday height and the provided
///   `recover_until` height, exclusive.
/// - `purpose`: 0 (Spending) if data required for spending should be tracked by the wallet,
///   or 1 (ViewOnly) if the account will never be used to spend funds.
/// - `seed_fingerprint_bytes`: the [seed fingerprint]. Only used if `purpose` is 0 (Spending).
/// - `hd_account_index_raw`: the ZIP 32 account index. Only used if `purpose` is 0 (Spending).
///
/// [seed fingerprint]: https://zips.z.cash/zip-0032#seed-fingerprints
```

daira (commented):
Reviewed with comments.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/derived/TxOutputsView.kt
```diff
@@ -73,9 +72,9 @@ internal class TxOutputsView(private val sqliteDatabase: SupportSQLiteDatabase)
                 if (!it.isNull(toAccountIndex)) {
                     // TODO [#1644]: Refactor Account ZIP32 index across SDK
                     // TODO [#1644]: https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/issues/1644
-                    TransactionRecipient.Account(0)
+                    TransactionRecipient.RecipientAccount(it.getInt(toAccountIndex))
```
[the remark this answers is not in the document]
HonzaR: Done

demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/Constants.kt
HonzaR: This constant is part of Demo app only, and we use it only for account 0 selection across the Demo app's use cases. whereas Zashi works with `getAccounts`.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt
HonzaR: Oh, I see now. Later, I sorted the parameters alphabetically, which is what we do across the SDK. Do you want me to change it as you described?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/DerivationToolExt.kt
HonzaR: We could add this change. Let's use the same wrapper, but with the only difference in the inner parameter type as `Long`, and check the value for its scope as we do, e.g., in the case of `BlockHeight` [parameters](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/blob/6f9ab065df6a5c277ad6174759c02c3ca2fd0c01/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniRewindResult.kt#L19). We can't use unsigned numbers across SDK as they are not supported by Java, which we need for many APIs used in protobug files.

demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/ui/screen/home/viewmodel/WalletViewModel.kt
HonzaR: The `spendingKey` variable is a Kotlin StateFlow with a default value of null, which changes once we start observing the stream. So yes, we could leave null instead of throwing the error. The change pushed.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
HonzaR: Yes, we can do this optimization.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/DerivationToolExt.kt
```diff
@@ -20,8 +20,16 @@ fun Derivation.deriveUnifiedAddress(
...
     network: ZcashNetwork,
-    account: Account
-): UnifiedSpendingKey = UnifiedSpendingKey(deriveUnifiedSpendingKey(seed, network.id, account.value))
+    accountIndex: Int
+): UnifiedSpendingKey =
+    UnifiedSpendingKey(
+        JniUnifiedSpendingKey(
+            // FIXME: How to construct JniUnifiedSpendingKey without accountUuid?
+            // FIXME: The tests fixture currently used to pass tests
+            accountUuid = AccountFixture.new().accountUuid,
```
HonzaR, line 29: This is a bit tricky now. As the `deriveUnifiedSpendingKey` started to return `ByteArray` instead of the previously used `JniUnifiedSpendingKey`. Now we can't easily construct `UnifiedSpendingKey` just from `accountIndex`. Any ideas please? @str4d @nuttycom

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
```diff
@@ -1,12 +1,113 @@
...
+ * @param keySource A string identifier or other metadata describing the source of the seed
+ * @param seedFingerprint The seed fingerprint. Must be length 16.
+ * @param hdAccountIndex ZIP 32 account index
  */
-data class Account(val value: Int) {
+data class Account internal constructor(
+    val accountUuid: ByteArray,
+    val ufvk: String?,
+    val accountName: String?,
+    val keySource: String?,
```
Milan-Cerovsky, line 36: keysource we use should be "Keystone" or "Zashi", right?

Milan-Cerovsky: account name probably too 🤔

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
```diff
@@ -1,12 +1,113 @@
...
+    override fun toString(): String {
+        return "Account(accountUuid=${accountUuid.contentToString()}," +
+            " ufvk length=${ufvk?.length}," +
+            " accountName=$accountName," +
+            " keySource=$keySource," +
+            " seedFingerprint size=${seedFingerprint?.size}," +
+            " hdAccountIndex=$hdAccountIndex)"
+    }
+
+    override fun equals(other: Any?): Boolean {
```
Milan-Cerovsky, line 83: we probably do not need to override equals and hasCode here

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
```diff
@@ -1,12 +1,113 @@
...
+            "Account UUID must be 16 bytes"
+        }
+        seedFingerprint?.let {
+            require(seedFingerprint.size == JNI_ACCOUNT_SEED_FP_BYTES_SIZE) {
+                "Seed fingerprint must be 32 bytes"
+            }
+        }
+    }
+
+    companion object {
```
Milan-Cerovsky, line 51: put companion object to the bottom of the class and u can even name it 'companion object Factory' for it to be evident what it is

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/AccountCreateSetup.kt
```diff
@@ -0,0 +1,35 @@
...
+ * @param keySource A string identifier or other metadata describing the source of the seed
+ * @param seed the wallet's seed phrase. This is required the first time a new wallet is set up. For
+ * subsequent calls, seed is only needed if [InitializerException.SeedRequired] is thrown.
+ */
+data class AccountCreateSetup(
+    val accountName: String,
+    val keySource: String?,
+    val seed: ByteArray,
+) {
+    override fun equals(other: Any?): Boolean {
```
Milan-Cerovsky, line 16: equals and hashCode are generated by data class here too so we probably do not need to override it

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
```diff
@@ -132,14 +135,28 @@ interface Synchronizer {
...
+     *
+     * @throws [InitializeException.ImportAccountException] in case of the operation failure
      */
-    suspend fun getFastestServers(
-        context: Context,
-        servers: List<LightWalletEndpoint>
-    ): Flow<FastestServersResult>
+    suspend fun importAccountByUfvk(
+        purpose: AccountPurpose,
+        recoverUntil: Long?,
```
Milan-Cerovsky, line 156: where do we get recoverUntil and treeState from ?

Milan-Cerovsky (changes requested):

Milan-Cerovsky:
@HonzaR please merge feature branch into this one

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
HonzaR: As we discussed with @LukasKorba, I vote for using all caps `KEYSTONE` and `ZASHI` in case the `keySource`. The `accountName` parameter could not be tied by any rules.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
HonzaR: The `equals` and `hashCode` are required due to the `ByteArray` parameters in this class.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/AccountCreateSetup.kt
HonzaR: The `equals` and `hashCode` are required due to the `ByteArray` parameter in this class.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
HonzaR: It's the same as in the case of the `createAccount` API. Check its usage, please.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
LukasKorba: I used `Keystone` and `Zashi` atm. If we want to set some rules here, we can go either with lowercased or uppcased approach. In iOS world, uppcased is very rare so I would rather use `keystone` and `zashi`.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
HonzaR: We can do this later after a discussion in a follow-up to make such a change across the SDK. We use this naming and ordering, if you agree?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
HonzaR: Okay, let's use the lowercase version, then.  @Milan-Cerovsky do you agree?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/DerivationToolExt.kt
HonzaR: I see now. I just got details about this from @LukasKorba that we should remove all account-related information from spending and viewing key objects. Then, we can update this easily.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
Milan-Cerovsky: yes lets go with it

LukasKorba: FYI, I updated the code to follow the lowercase rule

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
HonzaR: Done

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeDerivationToolImpl.kt
HonzaR: Done

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/DerivationToolExt.kt
HonzaR: Done

HonzaR (commented):
Self-reviewed. I left a few comments. Pushing changes to address few more.

CHANGELOG.md
```diff
@@ -7,18 +7,28 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- `Synchronizer.getAccounts()`
-- `Synchronizer.walletBalances: StateFlow<Map<Account, AccountBalance>?>` that is replacement for the removed 
+- `Synchronizer.importAccountByUfvk()` has been added
+- `Synchronizer.getAccounts()` returning all the created or imported accounts. See the documentation in `Account`.
+- `Synchronizer.walletBalances: StateFlow<Map<AccountUuid, AccountBalance>?>` that is replacement for the removed 
   `orchardBalances`, `saplingBalances`, and `transparentBalance`
+- `getTransactions(accountUuid: AccountUuid)` to get transactions belonging to the given account
```
daira, line 14:
```suggestion
- `getTransactions(accountUuid: AccountUuid)` to get transactions belonging to the given account
- `model.Zip32AccountIndex`
```
Please check for other additions.

HonzaR: Changelog updated.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -663,6 +692,48 @@ class SdkSynchronizer private constructor(
...
+                chainTip ?: network.saplingActivationHeight
+            )
+        val treeState: TreeState = loadedCheckpoint.treeState()
+
+        return runCatching {
+            backend.importAccountUfvk(
+                purpose = purpose,
+                recoverUntil = chainTip?.value,
+                setup = setup,
+                treeState = treeState.encoded,
```
str4d, line 726:
```suggestion
                recoverUntil = chainTip,
                setup = setup,
                treeState = treeState,
```

str4d (commented):
Reviewed 0b603e2927f111d65af40632c541f3e2338ebf5b. I did not review the changes to the tests or the demo app.

Changes requested (I can't mark as such because I opened the PR).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/AccountPurpose.kt
```diff
@@ -0,0 +1,25 @@
...
+sealed class AccountPurpose {
+    /**
+     * Constant value that uniquely identifies this enum across FFI
+     */
+    abstract val value: Int
+
+    /**
+     * For spending accounts, the wallet will track information needed to spend received notes
+     */
+    data object Spending : AccountPurpose() {
```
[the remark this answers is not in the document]
HonzaR: Yes, on it.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/AccountCreateSetup.kt
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @param accountName A human-readable name for the account
+ * @param keySource A string identifier or other metadata describing the source of the seed
+ * @param seed the wallet's seed phrase. This is required the first time a new wallet is set up. For
+ * subsequent calls, seed is only needed if [InitializerException.SeedRequired] is thrown.
+ */
+data class AccountCreateSetup(
+    val accountName: String,
+    val keySource: String?,
+    val seed: ByteArray,
```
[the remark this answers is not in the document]
HonzaR: Yes, I'm adding this change, too. We could make such a change to the other APIs, too, to eliminate the need to override the `equals` and `hashCode`, though. That is the primary purpose of the class, I believe. We could also rename it a bit, so it makes a bit more sense. Going further we could then use it in the public APIs that still work with `ByteArray`.

str4d (commented):
utACK 7a064ece9cda0064bee9d7b1da1ffd44d37acaf9. Any further changes can be made in subsequent PRs to the feature branch.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt
nuttycom: No, consistency is better (though alphabetical parameter lists seems like not a great convention to me.)

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/TransactionEncoder.kt
```diff
@@ -25,6 +25,7 @@ internal interface TransactionEncoder {
      */
     suspend fun createTransaction(
         usk: UnifiedSpendingKey,
+        account: Account,
```
nuttycom, line 28: We should probably look up the account by deriving the UFVK from the USK and then using that to determine the account ID here.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/TransactionEncoderImpl.kt
```diff
@@ -47,25 +47,39 @@ internal class TransactionEncoderImpl(
...
         amount: Zatoshi,
         recipient: TransactionRecipient,
         memo: ByteArray?
     ): EncodedTransaction {
-        require(recipient is TransactionRecipient.Address)
+        require(recipient is TransactionRecipient.RecipientAddress)
 
-        val transactionId = createSpend(usk, amount, recipient.addressValue, memo)
+        val transactionId =
+            createSpend(
```
nuttycom, line 58: It is actually `createSpend` that needs to change, and then that can propagate up.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/AccountUsk.kt
```diff
@@ -0,0 +1,74 @@
...
+    val accountUuid: AccountUuid,
+    /**
+     * The binary encoding of the [ZIP 316](https://zips.z.cash/zip-0316) Unified Spending
+     * Key for [accountUuid].
+     *
+     * This encoding **MUST NOT** be exposed to users. It is an internal encoding that is
+     * inherently unstable, and only intended to be passed between the SDK and the storage
+     * backend. Wallets **MUST NOT** allow this encoding to be exported or imported.
+     */
+    private val bytes: FirstClassByteArray
```
nuttycom, line 28: Why is this not `UnifiedSpendingKey`?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/TransactionEncoder.kt
nuttycom: To clarify, `account` should not be an additional parameter, because we have enough information to discover it inside of `createSpend`.

nuttycom (commented):
post-hoc ACK, but before release we should remove the `Account` parameter from `createSpend` etc. and look those up based on the USK.

backend-lib/src/main/rust/lib.rs
```diff
@@ -438,17 +421,132 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_createAcc
...
+/// - `recover_until`: An optional height at which the wallet should exit "recovery mode". In
+///    order to avoid confusing shifts in wallet balance and spendability that may temporarily be
+///    visible to a user during the process of recovering from seed, wallets may optionally set a
+///    "recover until" height. The wallet is considered to be in "recovery mode" until there
+///    exist no unscanned ranges between the wallet's birthday height and the provided
+///   `recover_until` height, exclusive.
+/// - `purpose`: 0 (Spending) if data required for spending should be tracked by the wallet,
+///   or 1 (ViewOnly) if the account will never be used to spend funds.
+/// - `seed_fingerprint_bytes`: the [seed fingerprint]. Only used if `purpose` is 0 (Spending).
+/// - `hd_account_index_raw`: the ZIP 32 account index. Only used if `purpose` is 0 (Spending).
```
daira, line 471: The suggestions to add "Only used if `purpose` is 0 (Spending)." depended on the suggestion [below](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/pull/1640/files#r1874197278) to only convert/check these parameters in that case. By taking the comment suggestion but not the code suggestion, this is now incorrectly documented, because `seed_fingerprint_bytes` has to point to a 32-byte array and `hd_account_index_raw` has to be in the range of a valid ZIP 32 account index, *regardless* of `purpose`.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeBackendImpl.kt
```diff
@@ -27,29 +30,67 @@ internal class TypesafeBackendImpl(private val backend: Backend) : TypesafeBacke
...
+                    AccountPurpose.ViewOnly ->
+                        backend.importAccountUfvk(
+                            accountName = setup.accountName,
+                            keySource = setup.keySource,
+                            purpose = setup.purpose.value,
+                            recoverUntil = recoverUntil?.value,
+                            treeState = treeState.encoded,
+                            ufvk = setup.ufvk.encoding,
+                            seedFingerprint = null,
+                            zip32AccountIndex = null,
```
daira, line 81: Bug: these are not valid values for `seedFingerprint` and `zip32AccountIndex`.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/TransactionEncoder.kt
daira: Filed #1652.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Account.kt
```diff
@@ -1,12 +1,117 @@
...
  *
- * @param value A 0-based account index.  Must be >= 0.
+ * @param accountUuid The account identifier.
+ * @param ufvk The account's Unified Full Viewing Key, if any.
+ * @param accountName A human-readable name for the account. This is visible to the wallet user, and
+ *        the wallet app may have obtained it from them.
+ * @param keySource A string identifier or other metadata describing the location of the spending
+ *        key corresponding to the account's UFVK. This is set internally by the wallet app based on
+ *        its private enumeration of spending methods it supports.
+ * @param seedFingerprint The seed fingerprint. Must be length 16.
```
daira, line 31: This does not match "Seed fingerprint must be 32 bytes" (which indeed is the value of `JNI_ACCOUNT_SEED_FP_BYTES_SIZE`).

HonzaR: Good catch. Thank you!

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeBackendImpl.kt
HonzaR: Discussion and findings moved to https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/issues/1651

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/AccountUsk.kt
HonzaR: Yes, we could do that. Good point.

daira: Filed as #1654.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Zip32AccountIndex.kt
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @param index The account ZIP 32 account identifier
+ */
+data class Zip32AccountIndex internal constructor(
+    val index: Long
+) {
+    init {
+        require(index.isInUIntRange()) {
+            "Account index $index is outside of allowed UInt range"
+        }
```
daira, line 16: The valid range of a ZIP 32 account index is $[0, 2^{31})$, not $[0, 2^{32})$.

daira: Filed #1655.

sdk-lib/src/test/java/cash/z/ecc/android/sdk/model/AccountTest.kt
```diff
@@ -1,13 +1,15 @@
...
 import org.junit.Test
+import java.util.UUID
 import kotlin.test.assertFailsWith
 
 class AccountTest {
     @Test
     fun out_of_bounds() {
         assertFailsWith(IllegalArgumentException::class) {
-            Account(-1)
+            AccountFixture.new(accountUuid = UUID.fromString("random"))
```
daira, line 12: Why is this invalid? I'm surprised it passes.

daira (commented):
Post-hoc ACK with comments. Follow-up tickets: #1651, #1652, #1654, #1655.

backend-lib/src/main/rust/lib.rs
nuttycom: Fixed in #1656

PR #1687, opened 2025-02-19
JNI bugfixes

### LINKED ISSUE #1657: Error while inserting UTXO: Network type mismatch
#### If so, please list the steps to reproduce below:
1. 
2. 
3. 

### Actual behavior + errors
```
                 java.lang.RuntimeException: Error while inserting UTXO: Data DB is corrupted: Network type mismatch: account UIVK is for mainnet but a testnet address was requested.
                 	at cash.z.ecc.android.sdk.internal.jni.RustBackend.putUtxo(Native Method)
                 	at cash.z.ecc.android.sdk.internal.jni.RustBackend.access$putUtxo(RustBackend.kt:27)
                 	at cash.z.ecc.android.sdk.internal.jni.RustBackend$Companion.putUtxo(Unknown Source:0)
                 	at cash.z.ecc.android.sdk.internal.jni.RustBackend$Companion.access$putUtxo(RustBackend.kt:499)
                 	at cash.z.ecc.android.sdk.internal.jni.RustBackend$putUtxo$2.invokeSuspend(RustBackend.kt:463)
                 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
                 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
                 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:644)
                 	at java.lang.Thread.run(Thread.java:1012)
```

- App Version: 
- Android Version:
- Device: (if applies)

### LINKED ISSUE #1684: RustDerivationTool: Fix inconsistency in `account_index` type across JNI
Currently the JNI uses `Long` on the Kotlin side, but `jint` on the Rust side. Example:
https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/blob/a3ac1d3dda29412977f82e6e3d67383890bd1556/backend-lib/src/main/rust/lib.rs#L2171-L2178
https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/blob/a3ac1d3dda29412977f82e6e3d67383890bd1556/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustDerivationTool.kt#L68-L71

I thought we'd fixed all of these when we migrated `RustBackend` to the new account APIs. We should check what is going on here and fix as necessary.

str4d:
This fixes multiple JNI type signature inconsistencies, and cleans up argument handling on the Rust side to be less error-prone.

fixes #1657, fixes #1684

> **Note**

# Author

# Reviewer

### DIFF
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -2,7 +2,7 @@ use std::convert::{Infallible, TryFrom, TryInto};
 use std::error::Error;
 use std::num::{NonZeroU32, NonZeroUsize};
 use std::panic;
-use std::path::Path;
+use std::path::PathBuf;
 use std::ptr;
 
 use anyhow::anyhow;
@@ -103,26 +103,13 @@ fn wallet_db<P: Parameters>(
     params: P,
     db_data: JString,
 ) -> anyhow::Result<WalletDb<rusqlite::Connection, P>> {
-    WalletDb::for_path(utils::java_string_to_rust(env, &db_data)?, params)
+    WalletDb::for_path(path_from_jni(env, db_data)?, params)
         .map_err(|e| anyhow!("Error opening wallet database connection: {}", e))
 }
 
 fn block_db(env: &mut JNIEnv, fsblockdb_root: JString) -> anyhow::Result<FsBlockDb> {
-    FsBlockDb::for_path(utils::java_string_to_rust(env, &fsblockdb_root)?)
+    FsBlockDb::for_path(path_from_jni(env, fsblockdb_root)?)
         .map_err(|e| anyhow!("Error opening block source database connection: {:?}", e))
-}
-
-fn zip32_account_index_from_jlong(account_index: jlong) -> anyhow::Result<zip32::AccountId> {
-    u32::try_from(account_index)
-        .map_err(|_| ())
-        .and_then(|id| zip32::AccountId::try_from(id).map_err(|_| ()))
-        .map_err(|_| anyhow!("Invalid account ID"))
-}
-
-fn account_id_from_jni(env: &JNIEnv, account_uuid: JByteArray) -> anyhow::Result<AccountUuid> {
-    Ok(AccountUuid::from_uuid(Uuid::from_slice(
-        &utils::java_bytes_to_rust(env, &account_uuid)?,
-    )?))
 }
 
 /// Initializes global Rust state, such as the logging infrastructure and threadpools.
@@ -355,17 +342,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_getAccoun
     let res = catch_unwind(&mut env, |env| {
         let network = parse_network(network_id as u32)?;
         let db_data = wallet_db(env, network, db_data)?;
-
-        let ufvk_string = utils::java_string_to_rust(env, &ufvk_string)?;
-        let ufvk = match UnifiedFullViewingKey::decode(&network, &ufvk_string) {
-            Ok(ufvk) => ufvk,
-            Err(e) => {
-                return Err(anyhow!(
-                    "Error while deriving viewing key from string input: {}",
-                    e,
-                ));
-            }
-        };
+        let ufvk = parse_ufvk(env, ufvk_string, &network)?;
 
         let account = db_data.get_account_for_ufvk(&ufvk)?;
 
@@ -396,7 +373,7 @@ fn encode_usk<'a>(
 }
 
 fn decode_usk(env: &JNIEnv, usk: JByteArray) -> anyhow::Result<UnifiedSpendingKey> {
-    let usk_bytes = parse_secret(env, &usk)?;
+    let usk_bytes = secret_from_jni(env, usk)?;
 
     // The remainder of the function is safe.
     UnifiedSpendingKey::from_bytes(Era::Orchard, usk_bytes.expose_secret()).map_err(|e| match e {
@@ -443,9 +420,8 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_createAcc
     let res = catch_unwind(&mut env, |env| {
         let network = parse_network(network_id as u32)?;
         let mut db_data = wallet_db(env, network, db_data)?;
-        let seed = parse_secret(env, &seed)?;
-        let treestate = TreeState::decode(utils::java_bytes_to_rust(env, &treestate)?.as_slice())
-            .map_err(|e| anyhow!("Invalid TreeState: {}", e))?;
+        let seed = secret_from_jni(env, seed)?;
+        let treestate = parse_treestate(env, treestate)?;
         let recover_until = recover_until.try_into().ok();
 
         let birthday =
@@ -530,16 +506,8 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_importAcc
     let res = catch_unwind(&mut env, |env| {
         let network = parse_network(network_id as u32)?;
         let mut db_data = wallet_db(env, network, db_data)?;
-        let ufvk_str = java_string_to_rust(env, &ufvk_str)?;
-        let ufvk = UnifiedFullViewingKey::decode(&network, &ufvk_str).map_err(|e| {
-            anyhow!(
-                "Value \"{}\" did not decode as a valid UFVK: {}",
-                ufvk_str,
-                e
-            )
-        })?;
-        let treestate = TreeState::decode(utils::java_bytes_to_rust(env, &treestate)?.as_slice())
-            .map_err(|e| anyhow!("Invalid TreeState: {}", e))?;
+        let ufvk = parse_ufvk(env, ufvk_str, &network)?;
+        let treestate = parse_treestate(env, treestate)?;
         let recover_until = recover_until.try_into().ok();
 
         let birthday =
@@ -605,7 +573,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_isSeedRel
     let res = catch_unwind(&mut env, |env| {
         let network = parse_network(network_id as u32)?;
         let db_data = wallet_db(env, network, db_data)?;
-        let seed = parse_secret(env, &seed)?;
+        let seed = secret_from_jni(env, seed)?;
 
         // Replicate the logic from `initWalletDb`.
         Ok(match db_data.seed_relevance_to_derived_accounts(&seed)? {
@@ -646,20 +614,6 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_getCurren
     unwrap_exc_or(&mut env, res, ptr::null_mut())
 }
 
-struct UnifiedAddressParser(UnifiedAddress);
-
-impl zcash_address::TryFromRawAddress for UnifiedAddressParser {
-    type Error = anyhow::Error;
-
-    fn try_from_raw_unified(
-        data: zcash_address::unified::Address,
-    ) -> Result<Self, zcash_address::ConversionError<Self::Error>> {
-        data.try_into()
-            .map(UnifiedAddressParser)
-            .map_err(|e| anyhow!("Invalid Unified Address: {}", e).into())
-    }
-}
-
 /// Returns the transparent receiver within the given Unified Address, if any.
 #[unsafe(no_mangle)]
 pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_getTransparentReceiverForUnifiedAddress<
@@ -672,16 +626,9 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_getTransp
     let res = catch_unwind(&mut env, |env| {
         let _span =
             tracing::info_span!("RustBackend.getTransparentReceiverForUnifiedAddress").entered();
-        let ua_str = utils::java_string_to_rust(env, &ua)?;
+        let (network, ua) = parse_ua(env, ua)?;
 
-        let (network, ua) = match ZcashAddress::try_from_encoded(&ua_str) {
-            Ok(addr) => addr
-                .convert::<(_, UnifiedAddressParser)>()
-                .map_err(|e| anyhow!("Not a Unified Address: {}", e)),
-            Err(e) => return Err(anyhow!("Invalid Zcash address: {}", e)),
-        }?;
-
-        if let Some(taddr) = ua.0.transparent() {
+        if let Some(taddr) = ua.transparent() {
             let taddr = match taddr {
                 TransparentAddress::PublicKeyHash(data) => {
                     ZcashAddress::from_transparent_p2pkh(network, *data)
@@ -716,16 +663,9 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_getSaplin
     let res = catch_unwind(&mut env, |env| {
         let _span =
             tracing::info_span!("RustBackend.getSaplingReceiverForUnifiedAddress").entered();
-        let ua_str = utils::java_string_to_rust(env, &ua)?;
+        let (network, ua) = parse_ua(env, ua)?;
 
-        let (network, ua) = match ZcashAddress::try_from_encoded(&ua_str) {
-            Ok(addr) => addr
-                .convert::<(_, UnifiedAddressParser)>()
-                .map_err(|e| anyhow!("Not a Unified Address: {}", e)),
-            Err(e) => return Err(anyhow!("Invalid Zcash address: {}", e)),
-        }?;
-
-        if let Some(addr) = ua.0.sapling() {
+        if let Some(addr) = ua.sapling() {
             let output = env
                 .new_string(ZcashAddress::from_sapling(network, addr.to_bytes()).encode())
                 .expect("Couldn't create Java string!");
@@ -959,14 +899,6 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_getTotalT
     unwrap_exc_or(&mut env, res, -1)
 }
 
-fn parse_protocol(code: i32) -> Option<ShieldedProtocol> {
-    match code {
-        2 => Some(ShieldedProtocol::Sapling),
-        3 => Some(ShieldedProtocol::Orchard),
-        _ => None,
-    }
-}
-
 #[unsafe(no_mangle)]
 pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_getMemoAsUtf8<'local>(
     mut env: JNIEnv<'local>,
@@ -982,9 +914,8 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_getMemoAs
         let network = parse_network(network_id as u32)?;
         let db_data = wallet_db(env, network, db_data)?;
 
-        let txid = parse_txid(env, &txid_bytes)?;
-        let protocol = parse_protocol(pool_type)
-            .ok_or(anyhow!("Shielded protocol not recognized: {}", pool_type))?;
+        let txid = parse_txid(env, txid_bytes)?;
+        let protocol = parse_protocol(pool_type)?;
         let output_index = u16::try_from(output_index)?;
 
         let memo = db_data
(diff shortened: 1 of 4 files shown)


backend-lib/src/main/rust/lib.rs
```diff
@@ -396,7 +396,7 @@ fn encode_usk<'a>(
 }
 
 fn decode_usk(env: &JNIEnv, usk: JByteArray) -> anyhow::Result<UnifiedSpendingKey> {
-    let usk_bytes = SecretVec::new(env.convert_byte_array(usk).unwrap());
+    let usk_bytes = parse_secret(env, &usk)?;
```
daira, line 399: Nit: `parse_secret` should be called `secret_from_jni` (like `account_id_from_jni`), because it isn't doing any parsing.

backend-lib/src/main/rust/lib.rs
```diff
@@ -443,8 +443,8 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_createAcc
     let res = catch_unwind(&mut env, |env| {
         let network = parse_network(network_id as u32)?;
         let mut db_data = wallet_db(env, network, db_data)?;
-        let seed = SecretVec::new(env.convert_byte_array(seed).unwrap());
-        let treestate = TreeState::decode(&env.convert_byte_array(treestate).unwrap()[..])
+        let seed = parse_secret(env, &seed)?;
+        let treestate = TreeState::decode(utils::java_bytes_to_rust(env, &treestate)?.as_slice())
             .map_err(|e| anyhow!("Invalid TreeState: {}", e))?;
```
daira, line 448: There are three identical uses of this pattern. Add a `treestate_from_jni` helper.

backend-lib/src/main/rust/lib.rs
```diff
@@ -236,7 +236,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_initDataD
         let mut db_data = wallet_db(env, network, db_data)
             .map_err(|e| anyhow!("Error while opening data DB: {}", e))?;
 
-        let seed = (!seed.is_null()).then(|| SecretVec::new(env.convert_byte_array(seed).unwrap()));
+        let seed = utils::java_nullable_bytes_to_rust(env, &seed)?.map(SecretVec::new);
```
daira, line 239: Nit: There are two uses of this pattern. It's marginal, but I would add a nullable counterpart to `parse_secret`.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1809,7 +1802,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeTr
         let network = parse_network(network_id as u32)?;
```
daira, line 1794: On line 1801 the wrong method name is used in the `info_span!`.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1855,21 +1848,19 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeTr
...
-        } else {
-            MemoBytes::from_bytes(&env.convert_byte_array(memo)?)
-                .map(Some)
-                .map_err(|e| anyhow!("Invalid MemoBytes: {}", e))?
-        };
+        let memo = utils::java_nullable_bytes_to_rust(env, &memo)?
+            .as_deref()
+            .map(MemoBytes::from_bytes)
+            .transpose()
+            .map_err(|e| anyhow!("Invalid MemoBytes: {}", e))?;
```
daira, line 1855: This pattern occurs twice (also in `RustBackend.proposeShielding`); consider adding a `nullable_memo_from_jni` helper.

backend-lib/src/main/rust/lib.rs
```diff
@@ -530,15 +530,15 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_importAcc
...
         let mut db_data = wallet_db(env, network, db_data)?;
-        let ufvk_str = java_string_to_rust(env, &ufvk_str);
+        let ufvk_str = java_string_to_rust(env, &ufvk_str)?;
         let ufvk = UnifiedFullViewingKey::decode(&network, &ufvk_str).map_err(|e| {
             anyhow!(
                 "Value \"{}\" did not decode as a valid UFVK: {}",
                 ufvk_str,
                 e
             )
         })?;
```
daira, line 540: There are three near-identical uses of this pattern (one uses `map_err` rather than `match`). Add a `ufvk_from_jni` helper.

backend-lib/src/main/rust/lib.rs
```diff
@@ -552,18 +552,15 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_importAcc
...
-                    <[u8; 32]>::try_from(&env.convert_byte_array(seed_fingerprint_bytes)?[..])
-                        .ok()
-                        .map(SeedFingerprint::from_bytes)
-                } else {
-                    None
-                };
+                let seed_fingerprint =
+                    utils::java_nullable_bytes_to_rust(env, &seed_fingerprint_bytes)?
+                        .and_then(|b| b.as_slice().try_into().ok())
+                        .map(SeedFingerprint::from_bytes);
```
daira, line 563: Although this happens to only be used once, it's doing the same kind of thing as the other conversion methods, so I might be inclined to add a helper anyway.

backend-lib/src/main/rust/lib.rs
```diff
@@ -831,7 +828,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_isValidSa
     let res = catch_unwind(&mut env, |env| {
         let _span = tracing::info_span!("RustBackend.isValidSaplingAddress").entered();
         let network = parse_network(network_id as u32)?;
-        let addr = utils::java_string_to_rust(env, &addr);
+        let addr = utils::java_string_to_rust(env, &addr)?;
 
         match Address::decode(&network, &addr) {
             Some(addr) => match addr {
```
daira, line 834: Consider this helper:
```rust
fn parse_address(env: &JNIEnv, addr: JString<'local>, network_id: jint) -> anyhow::Result<ZcashAddress> {
    let addr = utils::java_string_to_rust(env, &addr)?;
    Address::decode(&network, &addr).ok_or(anyhow!("Address is for the wrong network"))
}
```
Then lines 830-839 become just:
```rust
            match parse_address(network_id as u32, &addr)? {
                Address::Sapling(_) => Ok(JNI_TRUE),
                Address::Transparent(_) | Address::Unified(_) | Address::Tex(_) => Ok(JNI_FALSE),
            }
```
This can also be used in `isValid{Transparent,Unified,Tex}Address`, saving 15 non-blank lines overall I think.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2036,12 +2025,12 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_createPro
         let network = parse_network(network_id as u32)?;
```
daira, line 2017: On line 2024 the wrong method name is used in the `info_span!`.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2169,8 +2157,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_requiresS
     let res = catch_unwind(&mut env, |env| {
         let _span = tracing::info_span!("RustBackend.pcztRequiresSaplingProofs").entered();
```
daira, line 2150:
```suggestion
        let _span = tracing::info_span!("RustBackend.requiresSaplingProofs").entered();
```

daira, line 2150: This is the correct name, but the mangled function name is wrong on line 2150. How did this ever work? (I assume it didn't, and so we must have inadequate test coverage.)

backend-lib/src/main/rust/lib.rs
daira: I checked all other uses of `info_span!` and made comments for any inconsistencies.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2169,8 +2157,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_requiresS
...
 
-        let pczt = Pczt::parse(&env.convert_byte_array(pczt)?[..])
-            .map_err(|e| anyhow!("Invalid PCZT: {:?}", e))?;
+        let pczt = parse_pczt(env, &pczt)?;
 
         let prover = Prover::new(pczt);
```
daira, line 2155: Aside (just below): I'm perplexed that there is no `From<bool> for jboolean`. Doesn't matter I guess; this pattern only occurs once.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2540,7 +2526,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_model_TorClient_createTor
     tor_dir: JString<'local>,
 ) -> jlong {
     let res = catch_unwind(&mut env, |env| {
-        let tor_dir = utils::java_string_to_rust(env, &tor_dir);
+        let tor_dir = utils::java_string_to_rust(env, &tor_dir)?;
         let tor_dir = Path::new(&tor_dir);
```
daira, line 2522: It happens that all of the uses of `Path::new` in this file (7 of them) are from Java strings. So this helper:
```rust
fn path_from_jni(env: &JNIEnv, path_str: &JString) -> anyhow::Result<Path> {
    Ok(Path::new(utils::java_string_to_rust(env, path_str)?))
}
```
would let all paths be immediately typed as `Path` and eliminate all of the `Path::new`s.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2250,14 +2236,14 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_extractAn
...
-        let pczt_with_signatures = Pczt::parse(&env.convert_byte_array(pczt_with_signatures)?[..])
-            .map_err(|e| anyhow!("Invalid PCZT-with-proofs: {:?}", e))?;
+        let pczt_with_signatures = parse_pczt(env, &pczt_with_signatures)
+            .map_err(|e| anyhow!("Invalid PCZT-with-signatures: {:?}", e))?;
 
-        let spend_params = utils::java_string_to_rust(env, &spend_params);
-        let output_params = utils::java_string_to_rust(env, &output_params);
+        let spend_params = utils::java_string_to_rust(env, &spend_params)?;
+        let output_params = utils::java_string_to_rust(env, &output_params)?;
         let prover = LocalTxProver::new(Path::new(&spend_params), Path::new(&output_params));
```
daira, line 2239:
```suggestion
        let spend_params = path_from_jni(env, &spend_params)?;
        let output_params = path_from_jni(env, &output_params)?;
        let prover = LocalTxProver::new(&spend_params, &output_params);
```

backend-lib/src/main/rust/lib.rs
```diff
@@ -2210,8 +2196,8 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_addProofs
         assert!(!prover.requires_orchard_proof());
 
         if prover.requires_sapling_proofs() {
-            let spend_params = utils::java_string_to_rust(env, &spend_params);
-            let output_params = utils::java_string_to_rust(env, &output_params);
+            let spend_params = utils::java_string_to_rust(env, &spend_params)?;
+            let output_params = utils::java_string_to_rust(env, &output_params)?;
             let local_prover =
                 LocalTxProver::new(Path::new(&spend_params), Path::new(&output_params));
```
daira, line 2194:
```suggestion
            let spend_params = path_from_jni(env, &spend_params)?;
            let output_params = path_from_jni(env, &output_params)?;
            let prover = LocalTxProver::new(&spend_params, &output_params);
```

backend-lib/src/main/rust/lib.rs
```diff
@@ -2036,12 +2025,12 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_createPro
         let network = parse_network(network_id as u32)?;
         let mut db_data = wallet_db(env, network, db_data)?;
         let usk = decode_usk(env, usk)?;
-        let spend_params = utils::java_string_to_rust(env, &spend_params);
-        let output_params = utils::java_string_to_rust(env, &output_params);
+        let spend_params = utils::java_string_to_rust(env, &spend_params)?;
+        let output_params = utils::java_string_to_rust(env, &output_params)?;
 
         let prover = LocalTxProver::new(Path::new(&spend_params), Path::new(&output_params));
```
daira, line 2023:
```suggestion
        let spend_params = path_from_jni(env, &spend_params)?;
        let output_params = path_from_jni(env, &output_params)?;
        let prover = LocalTxProver::new(&spend_params, &output_params);
```

backend-lib/src/main/rust/lib.rs
```diff
@@ -2742,3 +2727,20 @@ fn parse_network(value: u32) -> anyhow::Result<Network> {
...
+
+fn parse_txid(env: &JNIEnv, txid_bytes: &JByteArray) -> anyhow::Result<TxId> {
+    let txid_bytes = utils::java_bytes_to_rust(env, &txid_bytes)?;
+    Ok(TxId::read(&txid_bytes[..])?)
+}
+
+fn parse_pczt(env: &JNIEnv, pczt: &JByteArray) -> anyhow::Result<Pczt> {
+    Pczt::parse(&utils::java_bytes_to_rust(env, &pczt)?)
+        .map_err(|e| anyhow!("Invalid PCZT: {:?}", e))
+}
```
daira: Move all the conversion helpers together.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1050,7 +1050,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_writeBloc
     _: JClass<'local>,
     db_cache: JString<'local>,
     block_meta: JObjectArray<'local>,
-) -> jboolean {
```
daira, line 1000: The same argument applies to `initBlockMetaDb`: the return value is always discarded.

daira (approved):
utACK with non-blocking suggestions. The errors in `info_span!` strings should be addressed but it would be fine to do so in another PR. Note that, as pointed out in #1688, there are easily fixable clippy lint failures.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2686,63 +2720,19 @@ fn parse_network(value: u32) -> anyhow::Result<Network> {
...
-    account_uuid: JByteArray<'local>,
-    network_id: jint,
-) -> jobjectArray {
-    let res = catch_unwind(&mut env, |env| {
-        let _span = tracing::info_span!("RustBackend.listTransparentReceivers").entered();
-        let network = parse_network(network_id as u32)?;
-        let zcash_network = network.network_type();
-        let db_data = wallet_db(env, network, db_data)?;
-        let account = account_id_from_jni(&env, account_uuid)?;
+fn parse_secret(env: &JNIEnv, secret_bytes: &JByteArray) -> anyhow::Result<SecretVec<u8>> {
```
daira, line 2723:
```suggestion
fn parse_secret(env: &JNIEnv, secret_bytes: JByteArray) -> anyhow::Result<SecretVec<u8>> {
```
We are inconsistent (e.g. between `account_id_from_jni` and here) about whether we pass `JByteArray` or `&JByteArray`. Since `JByteArray` is implemented as an (indirect) wrapper around a pointer with only zero-size extra information, there is no need to pass a reference &mdash; but in any case be consistent.

backend-lib/src/main/rust/utils.rs
```diff
@@ -33,14 +33,30 @@ pub fn catch_unwind<F: FnOnce(&mut JNIEnv) -> R + UnwindSafe, R>(
     })
 }
 
-pub(crate) fn java_string_to_rust(env: &mut JNIEnv, jstring: &JString) -> String {
-    env.get_string(jstring)
-        .expect("Couldn't get Java string!")
-        .into()
+pub(crate) fn java_bytes_to_rust(env: &JNIEnv, jbytes: &JByteArray) -> anyhow::Result<Vec<u8>> {
```
daira, line 36: This could be `jbytes: JByteArray` since it is effectively already a pointer, as above.

backend-lib/src/main/rust/utils.rs
```diff
@@ -33,14 +33,30 @@ pub fn catch_unwind<F: FnOnce(&mut JNIEnv) -> R + UnwindSafe, R>(
...
+    Ok(env.convert_byte_array(jbytes)?)
 }
 
-pub(crate) fn java_nullable_string_to_rust(env: &mut JNIEnv, jstring: &JString) -> Option<String> {
-    (!jstring.is_null()).then(|| java_string_to_rust(env, jstring))
+pub(crate) fn java_nullable_bytes_to_rust(
+    env: &JNIEnv,
+    jbytes: &JByteArray,
```
daira, line 42: This could be `jbytes: JByteArray` since it is effectively already a pointer (see the comment on `parse_secret`).

backend-lib/src/main/rust/utils.rs
```diff
@@ -33,14 +33,30 @@ pub fn catch_unwind<F: FnOnce(&mut JNIEnv) -> R + UnwindSafe, R>(
...
+) -> anyhow::Result<Option<Vec<u8>>> {
+    (!jbytes.is_null())
+        .then(|| java_bytes_to_rust(env, jbytes))
+        .transpose()
+}
+
+pub(crate) fn java_string_to_rust(env: &mut JNIEnv, jstring: &JString) -> anyhow::Result<String> {
```
daira, line 49: This could be `jstring: JString` since it is effectively already a pointer (see the comment on `parse_secret`).

backend-lib/src/main/rust/utils.rs
```diff
@@ -33,14 +33,30 @@ pub fn catch_unwind<F: FnOnce(&mut JNIEnv) -> R + UnwindSafe, R>(
...
+    Ok(env.get_string(jstring)?.into())
+}
+
+pub(crate) fn java_nullable_string_to_rust(
+    env: &mut JNIEnv,
+    jstring: &JString,
```
daira, line 55: This could be `jstring: JString` since it is effectively already a pointer (see the comment on `parse_secret`).

nuttycom (approved):
utACK c78afc7d989bb74e1e375e38a62e472988495fe3

backend-lib/src/main/rust/lib.rs
str4d: We cannot rename the JNI method, because the arguments between the extern function and the Backend trait method are identical. You'll see I use different names for several such methods, whereas when the Backend trait method has fewer arguments (e.g. due to us inserting a `db_data` path for the JNI) we use the same method name.

backend-lib/src/main/rust/lib.rs
str4d: We cannot use `Path::new` in the helper like this because it is a reference type, and this would lead to an early drop of a temporary variable. We are instead forced to reallocate in a `PathBuf`. But that's fine.

backend-lib/src/main/rust/utils.rs
str4d: Leaving as-is because the method it wraps takes a reference, and this method doesn't do anything special on top (the type-specific helper methods do by comparison; they effectively are converting type, so we want to move there).

backend-lib/src/main/rust/lib.rs
```diff
@@ -2720,19 +2646,76 @@ fn parse_network(value: u32) -> anyhow::Result<Network> {
...
+    type Error = anyhow::Error;
+
+    fn try_from_raw_unified(
+        data: zcash_address::unified::Address,
+    ) -> Result<Self, zcash_address::ConversionError<Self::Error>> {
+        data.try_into()
+            .map(UnifiedAddressParser)
+            .map_err(|e| anyhow!("Invalid Unified Address: {}", e).into())
+    }
+}
```
nuttycom, line 2710: Why not just use https://github.com/zcash/librustzcash/blob/main/zcash_keys/src/address.rs#L28? 

It's really weird to me to have to go through `TryFromRawAddress` for these straightforward operations.

nuttycom (approved):
utACK 55299225c004ad5a5cf9bcdcaab9e7a4db865395 with suggestion

backend-lib/src/main/rust/lib.rs
nuttycom: (I mean, why not use it directly?)

str4d: I _am_ using that; it's the `data.try_into()` call. The rest of the `TryFromRawAddress` machinery here is ensuring that if someone passes a string that isn't a UA (and is instead some other kind of address, or not an address at all), we get correct errors.

nuttycom: Ah. So, looking at the call sites of `parse_ua`, what is the reason to restrict these to unified addresses? If we parsed an arbitrary `ZcashAddress` at these call sites, we could still extract the transparent and/or Sapling receivers if present.

str4d: That might be worth considering, but that's outside the scope of this PR.

backend-lib/src/main/rust/lib.rs
daira: It might be better to systematically use a different name (e.g. add an `ffi` prefix), rather than sometimes use the same name and sometimes a different one.

PR #1874, opened 2026-02-24
feat: eip681 COR-1008

schell:
These changes introduce an EIP-681 transaction request parser (and an initial generator of sorts) to the Android SDK. 

Depends on:
* https://github.com/zcash/librustzcash/pull/2165

The impetus for these changes is to provide wallets with the ability accept and generate payment requests in the form of EIP-681.

> **Note**

# Author

# Reviewer

### DIFF
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustEip681Tool.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustEip681Tool.kt
@@ -1,0 +1,30 @@
+package cash.z.ecc.android.sdk.internal.jni
+
+import cash.z.ecc.android.sdk.internal.Eip681
+import cash.z.ecc.android.sdk.internal.model.JniEip681TransactionRequest
+
+class RustEip681Tool private constructor() : Eip681 {
+    @Throws(RuntimeException::class)
+    override fun parseTransactionRequest(uri: String): JniEip681TransactionRequest =
+        parseEip681TransactionRequest(uri)
+
+    @Throws(RuntimeException::class)
+    override fun transactionRequestToUri(request: JniEip681TransactionRequest): String =
+        eip681TransactionRequestToUri(request)
+
+    companion object {
+        suspend fun new(): Eip681 {
+            RustBackend.loadLibrary()
+
+            return RustEip681Tool()
+        }
+
+        @JvmStatic
+        private external fun parseEip681TransactionRequest(input: String): JniEip681TransactionRequest
+
+        @JvmStatic
+        private external fun eip681TransactionRequestToUri(
+            request: JniEip681TransactionRequest
+        ): String
+    }
+}
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniEip681TransactionRequest.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniEip681TransactionRequest.kt
@@ -24,7 +24,7 @@ sealed class JniEip681TransactionRequest {
      */
     @Keep
     @Suppress("LongParameterList")
-    class Native(
+    data class Native(
         val schemaPrefix: String,
         val hasPay: Boolean,
         val chainId: String?,
@@ -45,7 +45,7 @@ sealed class JniEip681TransactionRequest {
      * @param valueHex the transfer value in atomic units as a `0x`-prefixed hex string.
      */
     @Keep
-    class Erc20(
+    data class Erc20(
         val schemaPrefix: String,
         val hasPay: Boolean,
         val chainId: String?,
@@ -58,5 +58,5 @@ sealed class JniEip681TransactionRequest {
      * A valid EIP-681 request that is not a recognized transfer pattern.
      */
     @Keep
-    class Unrecognised : JniEip681TransactionRequest()
+    data object Unrecognised : JniEip681TransactionRequest()
 }
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniEip681TransactionRequest.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniEip681TransactionRequest.kt
@@ -23,6 +23,7 @@ sealed class JniEip681TransactionRequest {
      * @param gasPriceHex the gas price as a `0x`-prefixed hex string, or null if not specified.
      */
     @Keep
+    @Suppress("LongParameterList")
     class Native(
         val schemaPrefix: String,
         val hasPay: Boolean,
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniEip681TransactionRequest.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniEip681TransactionRequest.kt
@@ -1,0 +1,61 @@
+package cash.z.ecc.android.sdk.internal.model
+
+import androidx.annotation.Keep
+
+/**
+ * Serves as cross layer (Kotlin, Rust) communication class for parsed EIP-681 transaction requests.
+ *
+ * EIP-681 defines a standard URI format for Ethereum transaction requests, commonly used in
+ * QR codes and deep links. This sealed class represents the three recognized forms: native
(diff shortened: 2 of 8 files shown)


backend-lib/src/main/rust/eip681.rs
```diff
@@ -0,0 +1,263 @@
...
+        )
+        .map_err(|e| anyhow::anyhow!("Failed to construct NativeRequest: {}", e))
+    } else if env.is_instance_of(obj, &erc20_class)? {
+        let schema_prefix = get_string_field(env, obj, "schemaPrefix")?;
+        let has_pay = get_bool_field(env, obj, "hasPay")?;
+        let chain_id = chain_id_string_to_u64(get_nullable_string_field(env, obj, "chainId")?)?;
+        let token_contract = get_string_field(env, obj, "tokenContractAddress")?;
+        let recipient = get_string_field(env, obj, "recipientAddress")?;
+        let value_hex = get_string_field(env, obj, "valueHex")?;
+        let value = hex_string_to_u256(Some(value_hex))?.expect("valueHex is non-null for Erc20");
```
nuttycom: This probably shouldn't be a panic.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustEip681Tool.kt
```diff
@@ -0,0 +1,30 @@
...
+    override fun parseTransactionRequest(uri: String): JniEip681TransactionRequest =
+        parseEip681TransactionRequest(uri)
+
+    @Throws(RuntimeException::class)
+    override fun transactionRequestToUri(request: JniEip681TransactionRequest): String =
+        eip681TransactionRequestToUri(request)
+
+    companion object {
+        suspend fun new(): Eip681 {
+            RustBackend.loadLibrary()
```
noop-sk, line 17: The `suspend` modifier here is misleading — `loadLibrary()` and the constructor are both synchronous, so there is no async work being done. Either remove `suspend` (and let callers decide on dispatcher), or if library loading genuinely needs to be offloaded, wrap it in `withContext(Dispatchers.IO) { ... }` to make the suspension meaningful.

Also, `new` is an unusual name for a Kotlin factory. The conventional name here would be `create()`.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniEip681TransactionRequest.kt
```diff
@@ -0,0 +1,62 @@
...
+     * @param chainId the chain ID as a decimal string, or null if not specified in the URI.
+     * @param recipientAddress the recipient address (ERC-55 checksummed hex or ENS name).
+     * @param valueHex the transfer value as a `0x`-prefixed hex string, or null if not specified.
+     * @param gasLimitHex the gas limit as a `0x`-prefixed hex string, or null if not specified.
+     * @param gasPriceHex the gas price as a `0x`-prefixed hex string, or null if not specified.
+     */
+    @Keep
+    @Suppress("LongParameterList")
+    class Native(
+        val schemaPrefix: String,
```
noop-sk, line 28: Consider using `data class` here (and for `Erc20`) instead of plain `class`. Without it, `equals()`, `hashCode()`, and `toString()` are not implemented, which makes these classes harder to use in tests and comparisons. The `@Keep` annotation is fully compatible with `data class`.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniEip681TransactionRequest.kt
```diff
@@ -0,0 +1,62 @@
...
+        val recipientAddress: String,
+        val valueHex: String
+    ) : JniEip681TransactionRequest()
+
+    /**
+     * A valid EIP-681 request that is not a recognized transfer pattern.
+     */
+    @Keep
+    class Unrecognised : JniEip681TransactionRequest()
+}
```
noop-sk, line 62: Since `Unrecognised` carries no state, it should be a `data object` rather than a `class`:
```kotlin
@Keep
data object Unrecognised : JniEip681TransactionRequest()
```
This avoids creating a new instance on every call and gives correct `equals`/`hashCode` semantics for free.

backend-lib/src/main/rust/eip681.rs
```diff
@@ -0,0 +1,263 @@
...
+
+/// Parse a nullable `0x`-prefixed hex string into an `Option<U256>`.
+fn hex_string_to_u256(s: Option<String>) -> anyhow::Result<Option<U256>> {
+    match s {
+        Some(hex) => {
+            let stripped = hex
+                .strip_prefix("0x")
+                .or_else(|| hex.strip_prefix("0X"))
+                .unwrap_or(&hex);
+            Ok(Some(U256::from_str_radix(stripped, 16).map_err(|e| {
```
noop-sk, line 58: The doc comment says this function accepts a `0x`-prefixed hex string, but the `.unwrap_or(&hex)` fallback silently accepts un-prefixed hex strings too. This inconsistency could mask bugs on the Kotlin side where a caller passes a value without the prefix. Consider returning an error if neither `0x` nor `0X` prefix is found, to keep the contract strict and consistent with what `u256_option_to_jstring` produces on the encode path.

backend-lib/src/main/rust/eip681.rs
```diff
@@ -0,0 +1,263 @@
...
+        .map_err(|e| anyhow::anyhow!("Failed to construct NativeRequest: {}", e))
+    } else if env.is_instance_of(obj, &erc20_class)? {
+        let schema_prefix = get_string_field(env, obj, "schemaPrefix")?;
+        let has_pay = get_bool_field(env, obj, "hasPay")?;
+        let chain_id = chain_id_string_to_u64(get_nullable_string_field(env, obj, "chainId")?)?;
+        let token_contract = get_string_field(env, obj, "tokenContractAddress")?;
+        let recipient = get_string_field(env, obj, "recipientAddress")?;
+        let value_hex = get_string_field(env, obj, "valueHex")?;
+        let value = hex_string_to_u256(Some(value_hex))?.expect("valueHex is non-null for Erc20");
+
```
noop-sk, line 201: `.expect()` in a JNI function will trigger a Rust panic, which aborts the process rather than throwing a Java exception. Even though `valueHex` is non-nullable on the Kotlin side and this path should be unreachable in practice, it is safer to propagate an error through the `?` operator:
```rust
let value = hex_string_to_u256(Some(value_hex))?
    .ok_or_else(|| anyhow::anyhow!("valueHex must be non-null for Erc20"))?;
```
This way a logic bug produces a Java `RuntimeException` rather than a hard crash.

noop-sk (commented):
Overall the JNI bridge implementation is solid and follows existing patterns in the codebase. A few issues worth addressing before merge: missing tests, a spurious `suspend` modifier, and the sealed class variants should be `data class`/`data object`. Also a design question: what is the intended use-case for EIP-681 in the Zcash SDK? The PR description mentions a dependency on librustzcash#2165 but doesn't explain the higher-level motivation.

noop-sk (changes requested):

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustEip681Tool.kt
schell: This is an established pattern, so I'm just following the convention here: 
- RustDerivationTool.kt uses identical pattern: suspend fun new() with RustBackend.loadLibrary()
- RustBackend.kt uses identical pattern: suspend fun new(...) with loadLibrary()

schell:
@noop-sk - all comments addressed 🙇 . Now I'm just waiting for CI.

schell:
As far as I can tell, all 4 failing checks are due to pre-existing bugs on main that got inherited through my rebase. These failures also exist on `main` itself but aren't visible because the `pull-request.yml` workflow doesn't run on pushes to `main`. 

I'm starting work on a fix PR to rebase this on top of, but want to make sure that @noop-sk @Milan-Nerevar are not working on a similar PR?

nuttycom (approved):
utACK 2082b6c24f0fdc0ee3db2e5b0d8ec5f7d76c9fbc

noop-sk (approved):

PR #1919, opened 2026-04-17
MOB-1030: Fixed runtime crashes

noop-sk:
<!-- Write any additional comments here when opening the pull request -->

Root cause
                                                                                                                                                                                                                                        
  When the Rust layer returns a scan range where endHeight == startHeight (observed after a resync reset), ScanRange.new() attempts to subtract 1 from endHeight to convert the exclusive bound into a ClosedRange. This produces     
  BlockHeight(-1), which immediately throws IllegalArgumentException — BlockHeight validates that its value fits within the UInt32 range (0..4294967295).                                                                               
   
  Fix                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                      
  ScanRange.new() now returns null for ranges where endHeight <= startHeight and logs a warning for observability. The caller in TypesafeBackendImpl.suggestScanRanges() uses mapNotNull to skip these ranges silently, as they contain 
  no blocks.
                                                                                                                                                                                                                                        
  The condition uses <= rather than == to also cover inverted ranges (endHeight < startHeight). Rust should not produce these, but if it did, the original code would silently create an empty ClosedRange with no blocks scanned and no
   indication of the problem.
                                                                                                                                                                                                                                        
  Tests added                                                                                                                                                                                                                           
  - empty_scan_range_returns_null — zero heights (0, 0)
  - non_zero_empty_scan_range_returns_null — equal non-zero heights (100, 100)                                                                                                                                                          
  - inverted_scan_range_returns_null — inverted range (200, 100)    

> **Note**
>This code review checklist is intended to serve as a starting point for the author and reviewer, although it may not be appropriate for all types of changes (e.g. fixing a spelling typo in documentation).  For more in-depth discussion of how we think about code review, please see [Code Review Guidelines](../blob/main/docs/CODE_REVIEW_GUIDELINES.md).

# Author
<!-- NOTE: Do not modify these when initially opening the pull request.  This is a checklist template that you tick off AFTER the pull request is created. -->

- [ ] **Self-review** your own code in GitHub's web interface[^1]
- [ ] Add **automated tests** as appropriate
- [ ] Update the [**manual tests**](../blob/main/docs/testing/manual_testing)[^2] as appropriate
- [ ] Check the **code coverage**[^3] report for the automated tests
- [ ] Update **documentation** as appropriate (e.g [README.md](../blob/main/README.md), [Architecture.md](../blob/main/docs/Architecture.md), etc.)
- [ ] **Run the demo app** and try the changes
[...]

PR #1923, opened 2026-04-22
fix: MOB-1112 regressions in MOB-987 and MOB-1100 breaking CI on main

schell:
## Summary

Three pre-existing regressions on `main` that are currently breaking 4 of the 5 required CI checks on any new PR. These were introduced by MOB-987 and MOB-1100 and got merged despite failing CI (see 'Why this slipped through' below). This PR fixes all three so new PRs can pass CI cleanly.

## Checks currently failing on `main`

- `demo_app_release_build`
- `static_analysis_android_lint`
- `test_android_modules_unit`
- `test_android_modules_wtf`

All four will pass once this lands.

## AI Disclosure

This PR was co-authored by Claude Opus 4.7 (including this description)

### 1. MOB-987: Pass Locale to ZecSendExt.new in SendView

`demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/ui/screen/send/view/SendView.kt`

MOB-987 changed `ZecSendExt.new()` signature from `(Context, ...)` to `(Locale, ...)` but missed two call sites in `SendView.kt` (lines 259 and 286). They still passed `context`, producing compile errors:

```
e: SendView.kt:259:25 Argument type mismatch: actual type is 'Context', but 'Locale' was expected.
e: SendView.kt:286:25 Argument type mismatch: actual type is 'Context', but 'Locale' was expected.
```

This kills both `demo_app_release_build` and `static_analysis_android_lint`.

Fix: replace `context` with `Locale.getDefault()` at those two call sites, matching the pattern already applied elsewhere in the same file (line 158).

### 2. MOB-1100: Allow empty/inverted scan ranges from Rust, fix unit test mocking

`backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniScanRange.kt`
`build-conventions/src/main/kotlin/zcash-sdk.android-conventions.gradle.kts`

The three `ScanRangeTest` unit tests introduced by MOB-1100 fail at runtime:

- `empty_scan_range_returns_null` / `non_zero_empty_scan_range_returns_null`: throw `RuntimeException` because `Twig.warn` → `android.util.Log.w` throws "Method w in android.util.Log not mocked" in plain JVM unit tests.
- `inverted_scan_range_returns_null`: throws `IllegalArgumentException` from `JniScanRange`'s init block, before `ScanRange.new` can reach its null-return path.

Two coupled fixes:

1. **`JniScanRange`**: drop `require(endHeight >= startHeight)`. The Rust layer can legitimately return empty/inverted ranges (e.g. after a resync reset), and `ScanRange.new()` already handles them by returning null with a warning. Rejecting them in `JniScanRange` is what caused the crashes MOB-1100 was trying to fix in the first place.
2. **`android-conventions.gradle.kts`**: enable `testOptions.unitTests.isReturnDefaultValues` so `android.util.*` calls are no-ops in local unit tests rather than throwing.

### 3. MOB-987: Return null from Zatoshi.fromZecString for empty or invalid input

`sdk-incubator-lib/src/main/java/cash/z/ecc/android/sdk/model/ZecString.kt`

Three `ZecStringTest` (emulator.wtf) cases regressed after MOB-987 removed `ZecStringExt.filterConfirm()`:

- `empty_string`: empty input returned `Zatoshi(value=0)` instead of `null`, because `BigDecimal?.convertZecToZatoshi()` returns `Zatoshi(0L)` on null.
- `parse_bad_string` and `parse_invalid_numbers`: partial-parse inputs like `"asdf"`, `"+@#\$~^&*="`, `"1,2"`, `"1,23,"`, `"1,234,"` weren't rejected because `DecimalFormat.parse(String)` is lenient by default.

Fix by:
1. Short-circuiting empty input with an early `return null`.
2. Switching to `DecimalFormat.parse(String, ParsePosition)` and verifying `parsePosition.index == zecString.length`. If the whole string isn't consumed, return null.

## Why this slipped through

Looking at PR #1914 (MOB-987) and PR #1919 (MOB-1100) head commits, both showed these 4 CI checks as **failing** at merge time. They were merged anyway. Two systemic reasons:

1. **No branch protection**: `main` doesn't enforce "all checks must pass before merge", so failing CI didn't block the merge button.
2. **No post-merge verification**: `pull-request.yml` runs only on `pull_request` events, not on pushes to `main`. So once bad code lands, nothing re-verifies `main` and the team doesn't get notified the branch is broken.

As a follow-up (out of scope for this PR), consider enabling branch protection with required status checks on `main`.

## Testing

- Fix 1: compile-only fix; verifiable via `./gradlew :demo-app:compileZcashmainnetReleaseKotlin`.
- Fix 2: run `./gradlew :sdk-lib:testDebugUnitTest` locally; `ScanRangeTest` three tests should now pass.
- Fix 3: run the emulator.wtf tests via `./gradlew :sdk-incubator-lib:testDebugWithEmulatorWtf`. Alternatively these will be exercised by CI on this PR.

## Unblocks

This PR unblocks #1874 (`feat/eip681`), which will be rebased on top of this once merged.

### DIFF
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniScanRange.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniScanRange.kt
@@ -25,8 +25,9 @@ class JniScanRange(
         require(endHeight.isInUIntRange()) {
             "Height $endHeight is outside of allowed UInt range"
         }
-        require(endHeight >= startHeight) {
-            "End height $endHeight must be greater than start height $startHeight."
-        }
+        // Note: endHeight < startHeight is intentionally allowed here. The Rust layer can return
+        // empty or inverted ranges (e.g. after a resync reset), and ScanRange.new() handles this
+        // case by returning null. Rejecting such ranges at this layer would cause the crashes that
+        // MOB-1100 was intended to fix.
     }
 }
--- a/demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/ui/screen/send/view/SendView.kt
+++ b/demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/ui/screen/send/view/SendView.kt
@@ -256,7 +256,7 @@ private fun SendMainContent(
             onClick = {
                 val zecSendValidation =
                     ZecSendExt.new(
-                        context,
+                        Locale.getDefault(),
                         recipientAddressString,
                         amountZecString,
                         memoString,
@@ -283,7 +283,7 @@ private fun SendMainContent(
             onClick = {
                 val zecSendValidation =
                     ZecSendExt.new(
-                        context,
+                        Locale.getDefault(),
                         recipientAddressString,
                         amountZecString,
                         memoString,
--- a/sdk-incubator-lib/src/main/java/cash/z/ecc/android/sdk/model/ZecString.kt
+++ b/sdk-incubator-lib/src/main/java/cash/z/ecc/android/sdk/model/ZecString.kt
@@ -55,6 +55,8 @@ data class MonetarySeparators(
 
 private const val DECIMALS = 8
 
+private const val GROUP_SIZE = 3
+
 // TODO [#412]: https://github.com/zcash/zcash-android-wallet-sdk/issues/412
 // The SDK needs to fix the API for currency conversion
 fun Zatoshi.toZecString(locale: Locale) = convertZatoshiToZecString(locale, DECIMALS, DECIMALS)
@@ -73,8 +75,6 @@ const val FRACTION_DIGITS = 2
  * @return [zecString] parsed into Zatoshi or null if parsing failed.
  */
 fun Zatoshi.Companion.fromZecString(zecString: String, locale: Locale): Zatoshi? {
-    if (zecString.isEmpty()) return null
-
     val decimalFormat =
         currencyFormatter(
             locale = locale,
@@ -85,11 +85,22 @@ fun Zatoshi.Companion.fromZecString(zecString: String, locale: Locale): Zatoshi?
         }
 
     // Use ParsePosition so we can verify the entire string was consumed. DecimalFormat.parse may
-    // otherwise accept only a leading numeric portion (e.g. "1,2" or "1,23,"), which we need to reject.
-    // Completely invalid input such as "asdf" fails parsing at index 0 and results in null.
+    // otherwise accept only a leading numeric portion (e.g. "1.13 trailing" -> 1.13), which we need
+    // to reject. Completely invalid input such as "asdf" fails parsing at index 0 and results in
+    // null.
     val parsePosition = ParsePosition(0)
-    val parsed = decimalFormat.parse(zecString, parsePosition) as? BigDecimal
-    if (parsed == null || parsePosition.index != zecString.length) {
+    val parsed =
+        if (zecString.isEmpty()) {
+            null
+        } else {
+            decimalFormat.parse(zecString, parsePosition) as? BigDecimal
+        }
+
+    val separators = MonetarySeparators.current(locale)
+    if (parsed == null ||
+        parsePosition.index != zecString.length ||
+        !hasValidGrouping(zecString, separators)
+    ) {
         return null
     }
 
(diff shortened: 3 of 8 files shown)


backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/JniScanRange.kt
```diff
@@ -25,8 +25,9 @@ class JniScanRange(
         require(endHeight.isInUIntRange()) {
             "Height $endHeight is outside of allowed UInt range"
         }
-        require(endHeight >= startHeight) {
-            "End height $endHeight must be greater than start height $startHeight."
-        }
+        // Note: endHeight < startHeight is intentionally allowed here. The Rust layer can return
+        // empty or inverted ranges (e.g. after a resync reset), and ScanRange.new() handles this
+        // case by returning null. Rejecting such ranges at this layer would cause the crashes that
+        // MOB-1100 was intended to fix.
```
schell, line 31: ## What the require was trying to protect against
When it was added in 2023, the invariant "endHeight >= startHeight" was a data integrity check for the JNI boundary - i.e., "the Rust side shouldn't send us nonsense." It's a fail-fast assertion meant to catch Rust bugs early.

## Why removing this is reasonable
1. The real consumer already handles it. ScanRange.new() in sdk-lib is the actual code path that translates JniScanRange into ScanRange. It explicitly checks endHeight <= jni.startHeight and returns null with a warning log. The author of MOB-1100 documented this as expected behavior: "can occur after a resync reset when the Rust layer returns a zero-height range".
2. The require is actively harmful in production. When Rust legitimately returns an empty range (endHeight == startHeight), the require throws before ScanRange.new gets to filter it. This is precisely the crash MOB-1100 was trying to fix. So the require is doing the opposite of what it was intended to do — instead of catching a Rust bug, it's crashing the app on an expected edge case.
3. The >= case was already intended to be allowed. Notice the original require(endHeight >= startHeight) already permits endHeight == startHeight (an empty range). That slipped through the require in (year) 2023 but gets caught by ScanRange.new's <= check. So the layered handling was always the design — the require only rejects inverted ranges, which is a subtle subset of "bad data".

demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/ui/screen/send/view/SendView.kt
```diff
@@ -283,7 +283,7 @@ private fun SendMainContent(
             onClick = {
                 val zecSendValidation =
                     ZecSendExt.new(
-                        context,
+                        Locale.getDefault(),
```
schell, line 286: At these two call sites (lines 259 and 286), I replaced the first argument from context to Locale.getDefault():
```
// Before
ZecSendExt.new(
    context,                       // <- a Context
    recipientAddressString,
    amountZecString,
    memoString,
)
// After
ZecSendExt.new(
    Locale.getDefault(),           // <- a Locale
    recipientAddressString,
    amountZecString,
    memoString,
)
```
Same change applied in both onClick handlers (the "Get Proposal" button at line 255 and the "Send" button at line 282).

## Why this is needed
ZecSendExt.new() lives in sdk-incubator-lib/src/main/java/cash/z/ecc/android/sdk/model/ZecSendExt.kt. Its current signature is:
```
fun new(
    locale: Locale,
    destinationString: String,
    zecString: String,
    memoString: String,
): ZecSendValidation
```
The first parameter is `Locale`. `SendView.kt `was passing `Context,` which is a completely different type. The Kotlin compiler rejects this with:

e: SendView.kt:259:25 Argument type mismatch: actual type is 'Context', but 'Locale' was expected.

This kills demo_app_release_build and static_analysis_android_lint on every PR.

## How the mismatch happened
Before MOB-987 (commit 3c731a0d, merged 2026-04-16), the signature was:

```
fun new(
    context: Context,              // <-- used to take Context
    destinationString: String,
    zecString: String,
    memoString: String,
): ZecSendValidation
```

Internally it called `Zatoshi.fromZecString(context, zecString, locale)` - which itself used context for an ApplicationContext-based locale check via `ZecStringExt.filterConfirm(context, separators, zecString)`.
MOB-987 rewrote the whole formatting/parsing stack to be purely Locale-based (no more Android Context needed) and refactored ZecSendExt.new accordingly:
- Context parameter removed
- Locale parameter moved to first position (it used to be optional at the end)
The MOB-987 commit updated most call sites — for example, in the very same `SendView.kt` file at line 158, it changed:
```
// From:
accountBalance.sapling.available.toZecString()
// To:
accountBalance.sapling.available.toZecString(Locale.getDefault())
```
…but the author missed the two ZecSendExt.new(context, ...) call sites on lines 259 and 286. Since MOB-987's PR (#1914) was merged with demo_app_release_build and static_analysis_android_lint already failing, the oversight wasn't caught.

sdk-incubator-lib/src/main/java/cash/z/ecc/android/sdk/model/ZecString.kt
```diff
@@ -82,16 +84,17 @@ fun Zatoshi.Companion.fromZecString(zecString: String, locale: Locale): Zatoshi?
...
-        } catch (_: ParseException) {
-            null
-        }
+    // Use ParsePosition so we can verify the entire string was consumed. DecimalFormat.parse will
+    // otherwise accept partial input (e.g. "1,2", "1,23,", or "asdf" -> 0), which we need to reject.
+    val parsePosition = ParsePosition(0)
+    val parsed = decimalFormat.parse(zecString, parsePosition) as? BigDecimal
+    if (parsed == null || parsePosition.index != zecString.length) {
+        return null
+    }
```
schell, line 105: Parsing `"1.13 "` now returns `null` because `parsePosition.index == 4`, not `5`. Previously the lenient parse would have ignored trailing whitespace. If any existing caller relies on this, they'd regress. I consider this acceptable because (a) the tests assert this strict behavior and (b) input from a `TextField` is typically trimmed elsewhere.

Is that acceptable?

noop-sk (approved):

Issue #1921, opened 2026-04-21
Integrate and expose v1 implementation of shielded voting


true-jared:
This will enable wallets that use this SDK to interact with the Orchard-based Shielded Voting Protocol.

It is expected that this first version will have various limitations, such as requiring wallets to manually bake in details of each vote and release a corresponding update.

Corresponding Swift SDK implementation PR: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1687

PR #1933, opened 2026-05-06
[#1921] Expose Synchronizer snapshot accessors

greg0x:
## Summary

Refs zcash/zcash-android-wallet-sdk#1921.

Expose two narrow Synchronizer accessors needed by snapshot-height workflows:

- `fullyScannedHeight`, delegated from `CompactBlockProcessor.fullyScannedHeight`
- `getTreeState(height)`, returning protobuf-encoded tree state bytes from lightwalletd

This PR intentionally has no Rust changes, no voting dependency, and no voting backend/JNI surface.

Contribution-guide follow-ups included:

- ~~bumps `LIBRARY_VERSION` to `2.4.9` because this adds public API~~ removed, [per discussion](https://github.com/zcash/zcash-android-wallet-sdk/pull/1933#discussion_r3202096598)
- adds a `CHANGELOG.md` entry

## Testing

- `./gradlew checkProperties`
- `./gradlew ktlint`
- `./gradlew detektAll`
- `./gradlew :sdk-lib:compileReleaseKotlin`

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -22,7 +22,7 @@ ZCASH_ASCII_GPG_KEY=
 # Configures whether release is an unstable snapshot, therefore published to the snapshot repository.
 IS_SNAPSHOT=true
 
-LIBRARY_VERSION=2.4.9
+LIBRARY_VERSION=2.4.8
 
 # Kotlin compiler warnings can be considered errors, failing the build.
 ZCASH_IS_TREAT_WARNINGS_AS_ERRORS=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -22,7 +22,7 @@ ZCASH_ASCII_GPG_KEY=
 # Configures whether release is an unstable snapshot, therefore published to the snapshot repository.
 IS_SNAPSHOT=true
 
-LIBRARY_VERSION=2.4.8
+LIBRARY_VERSION=2.4.9
 
 # Kotlin compiler warnings can be considered errors, failing the build.
 ZCASH_IS_TREAT_WARNINGS_AS_ERRORS=true
@@ -139,4 +139,3 @@ BIP39_VERSION=1.0.9
 
 # This shouldn't be changed, as Android doesn't support targets beyond Java 8
 ANDROID_JVM_TARGET=1.8
-
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
@@ -696,10 +696,10 @@ class SdkSynchronizer private constructor(
             }
 
             is Response.Failure -> {
-                val message =
-                    "Failed to fetch tree state at height ${height.value}: ${response.toThrowable()}"
+                val throwable = response.toThrowable()
+                val message = "Failed to fetch tree state at height ${height.value}: $throwable"
                 Twig.error { message }
-                throw response.toThrowable()
+                throw throwable
             }
         }
 
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
@@ -692,7 +692,7 @@ class SdkSynchronizer private constructor(
                 )
         ) {
             is Response.Success -> {
-                TreeState.new(response.result).encoded
+                response.result.encoded
             }
 
             is Response.Failure -> {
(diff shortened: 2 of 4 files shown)


sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -680,6 +683,26 @@ class SdkSynchronizer private constructor(
...
+            is Response.Success -> {
+                TreeState.new(response.result).encoded
+            }
+
+            is Response.Failure -> {
+                val message =
+                    "Failed to fetch tree state at height ${height.value}: ${response.toThrowable()}"
+                Twig.error { message }
+                throw response.toThrowable()
+            }
```
noop-sk, line 703: `toThrowable()` is called twice — the object in the log message and the thrown one are different instances. Capture it once:

```kotlin
val throwable = response.toThrowable()
Twig.error { "Failed to fetch tree state at height ${height.value}: $throwable" }
throw throwable
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -680,6 +683,26 @@ class SdkSynchronizer private constructor(
...
+        when (
+            val response =
+                processor.downloader.getTreeState(
+                    height = BlockHeightUnsafe(height.value),
+                    serviceMode = sdkFlags ifTor ServiceMode.UniqueTor
+                )
+        ) {
+            is Response.Success -> {
+                TreeState.new(response.result).encoded
+            }
```
noop-sk, line 696: `TreeState.new()` only copies `unsafe.encoded` with no validation — see the `// Potential validation comes here` placeholder in `TreeState`. The wrapper is a no-op here; either use `response.result.encoded` directly, or add the actual validation that would justify the indirection.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
```diff
@@ -666,6 +675,14 @@ interface Synchronizer {
...
     suspend fun deleteAccount(accountUuid: AccountUuid): Boolean
 
+    /**
+     * Returns the commitment tree state at the given block height, encoded as a protobuf [ByteArray].
+     *
+     * This lets app-layer consumers generate witnesses or verify inclusion proofs against a
+     * specific Orchard note commitment tree snapshot without exposing the lightwalletd transport.
+     */
+    suspend fun getTreeState(height: BlockHeight): ByteArray
+
```
noop-sk, line 689: Question: returning raw `ByteArray` means callers need to know the protobuf schema out-of-band. Was returning `TreeState` (or `TreeStateUnsafe`) considered, or is there a reason to keep the return type opaque at the `Synchronizer` interface boundary?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -680,6 +683,26 @@ class SdkSynchronizer private constructor(
             refreshAccountsBus.emit(Unit)
         }
 
+    override suspend fun getTreeState(height: BlockHeight): ByteArray =
+        when (
+            val response =
+                processor.downloader.getTreeState(
+                    height = BlockHeightUnsafe(height.value),
```
noop-sk, line 690: Question: this always makes a live RPC to lightwalletd — there is no readiness guard. Should the KDoc note that callers ought to verify `fullyScannedHeight >= height` before calling, to ensure the returned tree state is consistent with the local wallet state?

gradle.properties
```diff
@@ -22,7 +22,7 @@ ZCASH_ASCII_GPG_KEY=
 # Configures whether release is an unstable snapshot, therefore published to the snapshot repository.
 IS_SNAPSHOT=true
 
-LIBRARY_VERSION=2.4.8
+LIBRARY_VERSION=2.4.9
```
noop-sk, line 26: Lets drop version change completely - we will adjust the version and release this in release branche.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
greg0x: b4cbdba9b63e

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
greg0x: 0c35d3bee22d

gradle.properties
greg0x: 2436c72d9315

sdk-lib/src/main/java/cash/z/ecc/android/sdk/Synchronizer.kt
greg0x: Good question. I kept it as bytes because this value is intended to be an opaque handoff, not something app code parses directly. In the larger shielded-vote flow, `getTreeState()` fetches the serialized lightwalletd `service.TreeState` protobuf and the app passes those bytes through to the voting backend; the Rust side decodes them when generating witnesses.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
greg0x: f0c1d5aa87ea

noop-sk (approved):
utACK

Issue #971, opened 2023-04-06
Refactor FFI to separate gradle module


ccjernigan:
## Is your feature request related to a problem? Please describe.
sdk-lib contains a lot of functionality, and it could be modularized further.  In particular, the Rust build is deeply integrated into the SDK.

## Describe the solution you'd like
Split the Rust FFI into a separate Gradle module.

To manage type safety while also splitting things apart, the Rust FFI will likely use primitives for its interface and we'll add extension functions in the sdk-lib to add type safety.

PR #1936, opened 2026-05-07
Fix backend Rust cache path

### LINKED ISSUE #1935: CI Rust cache points at sdk-lib after Rust FFI moved to backend-lib
## Problem

The Rust FFI crate moved from `sdk-lib` to `backend-lib` in `bc197971` (`[#971] Refactor Rust FFI to separate module`), but the CI Rust cache was not updated.

`backend-lib/build.gradle.kts` configures:

```kotlin
cargo {
    module = "."
}
```

so Cargo build outputs are written under `backend-lib/target`.

`.github/actions/setup/action.yml` still caches `sdk-lib/target` and keys the cache from paths that no longer exist:

```yaml
path: |
  sdk-lib/target
  ~/.cargo
key: ${{ runner.os }}-rust-${{ hashFiles(
  '.../sdk-lib/Cargo.lock',
  '.../sdk-lib/Cargo.toml',
  '.../sdk-lib/build.gradle.kts',
  '.../gradle.properties') }}
```

`sdk-lib/Cargo.lock` and `sdk-lib/Cargo.toml` no longer exist. `hashFiles` over missing paths returns an empty hash, so the key effectively collapses to `${{ runner.os }}-rust-` and never invalidates on Cargo input changes.

The Rust cache restore also has no `restore-keys` fallback, so once a real key is in use, any input change will produce a full miss rather than a partial reuse.

## Impact

`~/.cargo` is still cached, which avoids some registry/crate download work, but the expensive compiled artifacts under `backend-lib/target` are never persisted.

Every PR job that triggers `cargoBuild*` rebuilds all four Android ABIs (`arm`, `arm64`, `x86`, `x86_64`) from scratch. `backend-lib/build.gradle.kts` wires the JNI merge step to depend on all four `cargoBuild*` tasks unconditionally, and `javaPreCompileDebug`/`javaPreCompileRelease` depend on `cargoBuild`, so Rust compile work shows up in:

- `static_analysis_android_lint`
- `test_android_modules_unit`
- `test_android_modules_wtf`
- `demo_app_release_build`

On PR #1934 (a deps-changing PR) the Rust compile span was roughly 21-23 minutes inside each of the slow jobs. The cache miss is structural and exists on `main` as well; PRs that touch Rust deps just make it more visible.

## Proposed Fix

Replace the custom Rust cache block with [`Swatinem/rust-cache`](https://github.com/Swatinem/rust-cache), pointed at `backend-lib`. It handles the target dir, registry, key composition, restore-keys fallback, and stale-artifact pruning, so we don't re-derive these correctly by hand (which is how this bug originated).

If we keep a hand-rolled cache, the minimum corrections are:

- Cache `backend-lib/target` (not `sdk-lib/target`).
- Hash inputs that exist and are stable signals: `backend-lib/Cargo.lock`, `backend-lib/Cargo.toml`, `backend-lib/build.gradle.kts`, `backend-lib/build.rs`, `rust-toolchain.toml`, `.github/actions/setup/action.yml`.
- Add `restore-keys` fallback so unrelated input churn produces a partial reuse, not a cold start.
- Align the Rust cache save step's `if:` gate with the sibling Gradle binary/dependency saves, which gate on `github.event.pull_request.head.sha == ''`.

## Acceptance Criteria

- The Rust cache key changes when `backend-lib/Cargo.lock` or `backend-lib/Cargo.toml` changes.
- Unrelated input changes hit a `restore-keys` fallback and reuse the most recent compatible cache.
- The cache payload includes `backend-lib/target`.
- Warm reruns no longer recompile heavy backend dependencies (e.g. `orchard`, `zcash_proofs`, `zcash_client_backend`, `zcash_client_sqlite`) inside each slow PR job.

greg0x:
Closes #1935.

The CI Rust cache pointed at `sdk-lib/target` and hashed `sdk-lib/Cargo.{lock,toml}` — paths that haven't existed since the FFI moved to `backend-lib` in `bc197971` (`[#971] Refactor Rust FFI to separate module`). `hashFiles` over missing inputs returns empty, so the key collapsed to `${{ runner.os }}-rust-` and never invalidated, while the saved payload was empty. The result was that each PR job re-ran the full 21–23 min Rust compile from scratch on every run.

This points the cache at `backend-lib/target`, hashes inputs that actually exist (`backend-lib/Cargo.lock`, `backend-lib/Cargo.toml`, `backend-lib/build.gradle.kts`, `backend-lib/build.rs`, `rust-toolchain.toml`, `gradle.properties`, `.github/actions/setup/action.yml`), and adds a `restore-keys` fallback so input churn produces partial reuse rather than a cold start.

The split `actions/cache/restore@` + `actions/cache/save@` pair is collapsed into a single `actions/cache@`. The bespoke save step had no `if:` gate (unlike its Gradle siblings), and re-deriving cache key composition by hand is what produced this bug — keeping the surface small reduces the chance of similar drift in the future.

## Test plan

- [ ] CI passes on this PR.
- [ ] First run after merge repopulates the cache; subsequent runs of `static_analysis_android_lint`, `test_android_modules_unit`, `test_android_modules_wtf`, and `demo_app_release_build` show meaningfully shorter Rust compile spans against the warm cache.
- [ ] A `backend-lib/Cargo.lock` change invalidates the exact key but hits the `restore-keys` fallback for partial reuse.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -64,16 +64,18 @@ runs:
         key: ${{ runner.os }}-gradle-build-${{ github.sha }}
         restore-keys: |
           ${{ runner.os }}-gradle-build-
-    - name: Rust Cache Restore
+    - name: Rust Cache
       id: rust-cache-restore
-      uses: actions/cache/restore@668228422ae6a00e4ad889ee87cd7109ec5666a7
+      uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7
       env:
         SEGMENT_DOWNLOAD_TIMEOUT_MINS: "3"
       with:
         path: |
-          sdk-lib/target
+          backend-lib/target
           ~/.cargo
-        key: ${{ runner.os }}-rust-${{ hashFiles(format('{0}{1}', github.workspace, '/sdk-lib/Cargo.lock'), format('{0}{1}', github.workspace, '/sdk-lib/Cargo.toml'), format('{0}{1}', github.workspace, '/sdk-lib/build.gradle.kts'), format('{0}{1}', github.workspace, '/gradle.properties')) }}
+        key: ${{ runner.os }}-rust-${{ hashFiles(format('{0}{1}', github.workspace, '/backend-lib/Cargo.lock'), format('{0}{1}', github.workspace, '/backend-lib/Cargo.toml'), format('{0}{1}', github.workspace, '/backend-lib/build.gradle.kts'), format('{0}{1}', github.workspace, '/backend-lib/build.rs'), format('{0}{1}', github.workspace, '/rust-toolchain.toml'), format('{0}{1}', github.workspace, '/gradle.properties'), format('{0}{1}', github.workspace, '/.github/actions/setup/action.yml')) }}
+        restore-keys: |
+          ${{ runner.os }}-rust-
 
     - name: Gradle Binary Download
       if: steps.gradle-binary-cache-restore.outputs.cache-hit != 'true'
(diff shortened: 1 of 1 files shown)


y4ssi (approved):
utACK

PR #1937, opened 2026-05-08
[#1941] Add endpoint broadcaster API

### LINKED ISSUE #1941: Add endpoint broadcaster API for caller-managed transaction submission
### Motivation

The current SDK transaction flow creates and submits transactions through the synchronizer’s configured lightwalletd endpoint. Apps that want to broadcast the same transaction to one or more specific endpoints do not have a clean API for separating transaction creation from submission.

This makes multi-server broadcast awkward and can accidentally route submission through the synchronizer’s primary endpoint instead of the caller-selected endpoint.

### Proposed Solution

* Add a `Broadcaster` API that can create and store transactions without immediately submitting them.
* Add endpoint-specific submission so callers can submit created transaction bytes to a selected lightwalletd server.
* Expose the broadcaster from SDK-backed synchronizers while keeping a throwing default for unsupported synchronizer implementations.
* Preserve legacy synchronizer behavior for existing create-and-submit APIs.
* Mark broadcaster-managed transactions as caller-managed so the synchronizer does not automatically resubmit them through its own endpoint.

czarcas7ic:
Closes: https://github.com/zcash/zcash-android-wallet-sdk/issues/1941
Resolves: MOB-1146

iOS counterparts: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1678 and https://github.com/zcash/zcash-swift-wallet-sdk/pull/1737

Adds a public broadcaster API on Synchronizer that separates transaction creation from network submission. It can create/store proposed and PCZT transactions without broadcasting, submit created transactions to a caller-selected lightwalletd endpoint, and keeps the legacy submit APIs behavior-compatible by delegating through the broadcaster to the current endpoint.

Testing:

This tests the entire flow that includes PRs
- https://github.com/zcash/zcash-android-wallet-sdk/pull/1937
- https://github.com/zodl-inc/zodl-android/pull/2222
- https://github.com/zodl-inc/zodl-android/pull/2223

https://github.com/user-attachments/assets/7935ca54-ed2e-4aec-8bae-dfd893271350


sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcaster.kt
```diff
@@ -0,0 +1,210 @@
...
+internal suspend fun Broadcaster.createAndSubmitTransactionFromPczt(
+    pcztWithProofs: Pczt,
+    pcztWithSignatures: Pczt,
+    endpoint: LightWalletEndpoint
+): Flow<TransactionSubmitResult> =
+    createTransactionFromPczt(pcztWithProofs, pcztWithSignatures)
+        .asFlow()
+        .map { transaction -> submit(transaction, endpoint) }
+
+internal class EndpointTransactionSubmitter(
```
nullcopy, line 87: `EndpointTransactionSubmitter.submit` and `OutboundTransactionManagerImpl.submit` have near-identical bodies. Might be worth extracting a common helper for the two? Not blocking

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/AutomaticResubmissionGuard.kt
```diff
@@ -0,0 +1,72 @@
...
+
+internal class AutomaticResubmissionGuard(
+    private val preferenceProvider: PreferenceProvider? = null
+) {
+    private val mutex = Mutex()
+    private val callerManagedTransactionIds = mutableSetOf<String>()
+    private var loadedFromPreferences = false
+
+    suspend fun excludeFromAutomaticResubmission(transaction: CreatedTransaction) {
+        excludeFromAutomaticResubmission(transaction.txId)
```
nullcopy: This exclusion list is append only and will grow forever, won't it? Or does this get reset periodically?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcaster.kt
```diff
@@ -0,0 +1,210 @@
...
+internal interface TransactionSubmitter {
+    suspend fun submit(
+        transaction: CreatedTransaction,
+        endpoint: LightWalletEndpoint
+    ): TransactionSubmitResult
+}
+
+// These helpers compose the public Broadcaster API, so they preserve
+// caller-managed exclusion semantics.
+internal suspend fun Broadcaster.createAndSubmitProposedTransactions(
```
nullcopy: This could easily be confused with `SdkBroadcaster.createAndSubmitProposedTransactions`. Can this be moved into the test source set, since its only used there anyways?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcaster.kt
```diff
@@ -0,0 +1,210 @@
...
+    proposal: Proposal,
+    usk: UnifiedSpendingKey,
+    endpoint: LightWalletEndpoint
+): Flow<TransactionSubmitResult> =
+    createProposedTransactions(proposal, usk)
+        .createSubmitResultFlow { transaction ->
+            submit(transaction, endpoint)
+        }
+
+internal suspend fun Broadcaster.createAndSubmitTransactionFromPczt(
```
nullcopy: Same comment as [above](https://github.com/zcash/zcash-android-wallet-sdk/pull/1937/changes#r3222481930)

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
```diff
@@ -157,6 +158,8 @@ class SdkSynchronizer private constructor(
     private val torClient: TorClient?,
     private val walletClient: CombinedWalletClient,
     private val walletClientFactory: WalletClientFactory,
+    private val currentEndpoint: LightWalletEndpoint,
```
nullcopy: `currentEndpoint` field name is misleading, IMO. It is captured once in Builder.build() and never updated, but the name implies it tracks a dynamic value. Maybe call it `lightWalletEndpoint` or `defaultSubmitEndpoint` or something?

sdk-lib/src/test/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcasterTest.kt
```diff
@@ -0,0 +1,301 @@
...
+                raw = raw,
+                expiryHeight = expiryHeight
+            )
+
+        private fun fakeProposal(): Proposal = uninitializedInstance(Proposal::class.java)
+
+        private fun fakeUsk(): UnifiedSpendingKey = uninitializedInstance(UnifiedSpendingKey::class.java)
+
+        private fun <T> uninitializedInstance(clazz: Class<T>): T {
+            val unsafeClass = Class.forName("sun.misc.Unsafe")
```
nullcopy: Good feedback from Claude:
```
Tests use sun.misc.Unsafe to allocate Proposal / UnifiedSpendingKey

  SdkBroadcasterTest.uninitializedInstance reflects into sun.misc.Unsafe to allocate uninitialized instances. It works today only because FakeOutboundTransactionManager never reads those
  values, but:
  - it's JVM-version-fragile (Unsafe deprecation is in motion);
  - it obscures intent for reviewers;
  - the rest of the SDK uses Mockito for these cases.

  Since the values are never dereferenced, the cleaner option is to narrow FakeOutboundTransactionManager.createProposedTransactions / FakeBroadcaster.createProposedTransactions to accept
  proposal: Proposal? = null, or just pass Mockito mocks. The reflection hatch should go.
```

nullcopy (changes requested):
Reviewed 6b45f7cb4517a8e0318c500378f801e366c72735

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcaster.kt
czarcas7ic: https://github.com/zcash/zcash-android-wallet-sdk/pull/1937/commits/40e60135d4acefea10ca467d9a6ad3f38963d09d

I pulled the shared submit call and result mapping into a common helper.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcaster.kt
czarcas7ic: https://github.com/zcash/zcash-android-wallet-sdk/pull/1937/commits/97f030842a228fb9668291b34f701feaec70023d

Moved test helper to test file here

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcaster.kt
czarcas7ic: https://github.com/zcash/zcash-android-wallet-sdk/pull/1937/commits/97f030842a228fb9668291b34f701feaec70023d

sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt
czarcas7ic: https://github.com/zcash/zcash-android-wallet-sdk/pull/1937/commits/5816d7217eab2555c3ff528a7a9598e26a9446f8

Changed name to `defaultSubmitEndpoint`

sdk-lib/src/test/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcasterTest.kt
czarcas7ic: https://github.com/zcash/zcash-android-wallet-sdk/pull/1937/commits/909b41e495fa99ee9038d189301ef4172b943d51

Used Mockito mock here

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/AutomaticResubmissionGuard.kt
czarcas7ic: https://github.com/zcash/zcash-android-wallet-sdk/pull/1937/commits/3c2759615dc834709231180fb8bf236d19fc3700
https://github.com/zcash/zcash-android-wallet-sdk/pull/1937/commits/db519ba7f8d59b0907c0c5308490532596f32aea

Thanks for pointing this out, this issue is also in the Swift SDK version which I am addressing in a separate swift SDK PR https://github.com/zcash/zcash-swift-wallet-sdk/pull/1734

czarcas7ic: Hm, I need to think about this one a little more. We need to change the contract of the caller to have its own retry logic we we completely get rid of the resubmission logic, or do some other trick.

czarcas7ic: Okay, I am going to lay out my understanding and see what you think about the way forward.

In today’s flow, the SDK owns transaction creation and broadcast the whole way through, so it makes sense that the SDK also owns automatic retry/resubmission.

In the new Broadcaster flow, the SDK creates and stores the transaction, but the caller owns where and how it gets broadcast. In our case, automatic mode fans out across many RPCs, while manual mode submits only to the user-selected RPC. Given that, retry should also live with the caller, not inside the SDK. Otherwise the SDK can silently retry through a different/current endpoint and bypass the caller’s broadcast policy.

So I think the right next steps are:

1. Implement the same behavior on the Swift SDK side, like https://github.com/zcash/zcash-swift-wallet-sdk/pull/1734, so Broadcaster created transactions are excluded from SDK automatic resubmission there too (we missed this when we merged the Broadcaster there).
2. Update both zodl-ios PR 34 and zodl-android PR 3 to add an app owned Broadcaster retry queue. Those PRs are not reviewed or merged yet, and both are already blocked on the automatic/manual server-mode PRs landing first.

@nullcopy if this tracks with you, I can implement it. Let me know if you have other ideas.

nullcopy: That sounds like a reasonable approach to me. I'll leave it to @LukasKorba and @noop-sk to sign off on ios/android app decisions, but I have no objections

czarcas7ic: After discussing with @LukasKorba some more, we came to the conclusion that we could instead keep retry ownership in the SDK, but make the caller pass the submit plan.

So Zodl still decides where a tx is allowed to go. Manual mode passes one server. Automatic mode passes the full server list. The SDK stores that plan per tx and uses the same plan for resubmission.

That avoids the separate “do not retry this txid” list, keeps retry behavior in one place, and prevents the SDK from guessing the wrong endpoint later.

EDIT: This was implemented here https://github.com/zcash/zcash-android-wallet-sdk/pull/1937/commits/9097bd016569c6f677d1b30a0682552130a79d22

EDIT 2: The Swift SDK side update is R4R here https://github.com/zcash/zcash-swift-wallet-sdk/pull/1737

nullcopy (approved):
utACK 156fa4efa816feb6764bb803a4cfeeb85fc07f89

PR #1934, opened 2026-05-07
[#1924] Add zcash_voting dependency foundation

greg0x:
Refs #1924.
Resolves MOB-1109

Associated issue: zodl-inc/zodl-android#2193.

Android analogue of zcash/zcash-swift-wallet-sdk#1699.

## Summary

- add released `zcash_voting 0.5.3` to `backend-lib` with default features disabled
- pin `orchard` to `=0.13.1` with `unstable-voting-circuits`
- add a narrow internal JNI linkage symbol, `VotingRustBackend.computeShareNullifier`, to prove dispatch into `zcash_voting`
- place the Kotlin JNI holder in `sdk-lib`, matching the umbrella `shielded-vote` implementation
- align `computeShareNullifier` with the umbrella signature: `(voteCommitment, shareIndex, blind)`
- update `Cargo.lock` and `CHANGELOG.md`

## Non-goals

- no public SDK voting API
- no voting database lifecycle
- no wallet database access
- no PIR or vote-tree-sync feature graph
- no app-facing Kotlin voting backend surface

## Testing

- `cargo fmt --manifest-path backend-lib/Cargo.toml --check`
- `cargo check --manifest-path backend-lib/Cargo.toml --locked`
- `./gradlew :backend-lib:compileReleaseKotlin :sdk-lib:compileReleaseKotlin`
- `git diff --check`

PR #1924, opened 2026-04-28
Draft: Add Android coinholder polling backend

greg0x:
## Summary
- add the Android SDK voting backend for coinholder polling
- expose voting recovery, Keystone, and share-tracking primitives needed by the app flow
- harden Android voting state transitions for round prep, resume, and multi-proposal submission parity

## Issues
- [MOB-1109 Integrate and expose v1 implementation of shielded voting](https://linear.app/zodl/issue/MOB-1109/integrate-and-expose-v1-implementation-of-shielded-voting)
- https://github.com/zodl-inc/zodl-android/issues/2193

## Companion PRs
- companion app PR: `zodl-inc/zodl-android#2206`

## Recording
- https://cleanshot.com/share/rqGnp2Mg

## Notes
- this branch includes the original MOB-1108 SDK backend work plus the additional parity and recovery fixes needed to get live Android voting through end-to-end testing

## Testing
- `./gradlew :sdk-lib:assembleDebug`

PR #1938, opened 2026-05-08
[#2193] Add Android voting JNI round lifecycle

greg0x:
Depends on zcash/zcash-android-wallet-sdk#1934 (first slice — `zcash_voting` dependency foundation).
Resolves MOB-1109

## Why

Second slice of the Android shielded-voting integration. #1934 wires the `zcash_voting` crate into backend-lib; this PR adds the voting DB handle and round lifecycle so the round-keyed surface (init / state / list / get-votes / clear / delete-skipped) compiles end-to-end from Rust through JNI to Kotlin.

Foundation for the Coinholder Polling flow tracked in zodl-inc/zodl-android#2193 — no user-visible behavior on its own.

## What's in

Rust / JNI (`backend-lib/src/main/rust/voting/`):
- `db.rs` — open/close native handle, `setWalletId`
- `rounds.rs` — `initRound`, `getRoundState`, `listRoundsJson`, `getVotesJson`, `clearRound`, `deleteSkippedBundles`
- `share_tracking.rs` — existing `computeShareNullifier`, lifted out of the monolithic module
- `helpers.rs` — shared JNI helpers: fixed-width byte validation (`PROTOCOL_FIELD_BYTES = 32`), `jint`/`jlong` widening, `RoundState` → `FfiRoundState` construction
- `json.rs` — round summary / vote record serialization, `RoundPhase` ↔ `u32` mapping
- `voting.rs` — reduced to module roots + shared imports

Kotlin (`sdk-lib/.../internal/`):
- `jni/VotingRustBackend.kt` — external declarations for the new surface
- `TypesafeVotingBackend.kt` + impl — `suspend` wrappers, IO dispatcher, error checks, JSON → `VoteRecord` parsing
- `model/voting/FfiVotingModels.kt` — `FfiRoundState`, `FfiRoundPhase`

## Not included

- hotkey generation
- bundle setup/count
- proving cache warmup
- delegation proving/submission
- vote commitment or share payloads
- recovery/share tracking
- public Synchronizer voting API

## Review focus

Primary review track: core-dev. Please look at:
- native handle lifecycle (`Box::into_raw` / `Box::from_raw` across `openVotingDb`/`closeVotingDb`, `Arc<VotingDb>` usage)
- JNI boundary error semantics (panic-safe via `catch_unwind`, exception conversion)
- byte-array validation (`java_bytes_exact` on 32-byte protocol fields)
- `RoundState` / `RoundSummary` JSON and FFI mapping
- Kotlin surface is intentionally `internal` and minimal — just enough to compile and exercise the JNI

Swift counterparts:
- zcash/zcash-swift-wallet-sdk#1716
- zcash/zcash-swift-wallet-sdk#1722
- zcash/zcash-swift-wallet-sdk#1719

## Validation

- `./gradlew checkProperties`
- `git diff --check`

# Author

# Reviewer


p0mvn (dismissed):

noop-sk (commented):
PR added into code

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -0,0 +1,91 @@
...
+        None => JObject::null(),
+    };
+    let long_class = env.find_class("java/lang/Long")?;
+    let weight_obj: JObject<'local> = match state.delegated_weight {
+        Some(w) => env.new_object(&long_class, "(J)V", &[JValue::Long(w as i64)])?,
+        None => JObject::null(),
+    };
+    let obj = env.new_object(
+        &class,
+        "(Ljava/lang/String;IJLjava/lang/String;Ljava/lang/Long;Z)V",
```
noop-sk, line 123: **Brittle JNI constructor signature**

This hardcoded descriptor will silently throw `NoSuchMethodError` at runtime if `FfiRoundState`'s field order or types ever change. Consider adding a comment that locks it to the Kotlin data class definition:

```rust
// Matches FfiRoundState(roundId: String, phase: Int, snapshotHeight: Long,
//                       hotkeyAddress: String?, delegatedWeight: Long?, proofGenerated: Boolean)
```

Or cover it with an instrumented test that exercises object construction end-to-end on a device.

backend-lib/src/main/rust/voting/rounds.rs
```diff
@@ -0,0 +1,144 @@
...
+    mut env: JNIEnv<'local>,
+    _: JClass<'local>,
+    db_handle: jlong,
+    round_id: JString<'local>,
+) -> jobject {
+    let res = catch_unwind(&mut env, |env| {
+        let db = db_from_handle(db_handle)?;
+        match db.get_round_state(&java_string_to_rust(env, &round_id)?) {
+            Ok(state) => make_ffi_round_state(env, state),
+            Err(VotingError::InvalidInput { .. }) => Ok(JObject::null().into_raw()),
```
noop-sk: **Ambiguous null mapping**

Is `VotingError::InvalidInput` here specifically "round not found", or can it also fire for malformed input (empty `round_id`, etc.)? If the latter, callers expecting `null` for "not found" could silently swallow real errors. Worth either documenting or matching against a more specific variant (e.g. `NotFound`) once the crate exposes one.

backend-lib/src/main/rust/voting/json.rs
```diff
@@ -0,0 +1,63 @@
+use super::*;
+
+const PHASE_INITIALIZED: u32 = 0;
```
noop-sk: **Phase constant duplication**

These constants are duplicated in Kotlin's `FFI_ROUND_PHASE_*` in `FfiVotingModels.kt` and must stay in sync manually. A comment cross-referencing the counterpart would help:

```rust
// Must match FFI_ROUND_PHASE_* constants in FfiVotingModels.kt
```

Alternatively, a round-trip smoke test that writes a phase from Rust and asserts it matches `FfiRoundPhase.INITIALIZED.value` on the Kotlin side would catch divergence.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
```diff
@@ -0,0 +1,48 @@
...
+    val value: Int
+) {
+    INITIALIZED(FFI_ROUND_PHASE_INITIALIZED),
+    HOTKEY_GENERATED(FFI_ROUND_PHASE_HOTKEY_GENERATED),
+    DELEGATION_CONSTRUCTED(FFI_ROUND_PHASE_DELEGATION_CONSTRUCTED),
+    DELEGATION_PROVED(FFI_ROUND_PHASE_DELEGATION_PROVED),
+    VOTE_READY(FFI_ROUND_PHASE_VOTE_READY);
+
+    companion object {
+        fun fromInt(value: Int) =
```
noop-sk, line 35: **Forward compatibility risk**

This throws if Rust adds a new `RoundPhase` variant before Kotlin is updated — a library update alone can crash the app. For an internal `sdk-lib` boundary it is defensible, but consider an `UNKNOWN(-1)` fallback or returning `null` until the phase set is stable.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/FfiVotingModels.kt
```diff
@@ -0,0 +1,48 @@
...
+    VOTE_READY(FFI_ROUND_PHASE_VOTE_READY);
+
+    companion object {
+        fun fromInt(value: Int) =
+            entries.firstOrNull { it.value == value }
+                ?: error("Unknown round phase: $value")
+    }
+}
+
+@Keep
```
noop-sk: **Misleading `@Keep`**

`FfiRoundSummary` is never constructed via JNI reflection — it is built from JSON in `TypesafeVotingDbImpl.listRounds()`. The `@Keep` annotation implies JNI construction that does not exist; remove it or replace with a clarifying comment.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
```diff
@@ -8,6 +17,94 @@ class VotingRustBackend private constructor() {
...
+            withHandle { handle -> clearRoundNative(handle, roundId) }
+
+        @Throws(RuntimeException::class)
+        suspend fun deleteSkippedBundles(
+            roundId: String,
+            keepCount: Int
+        ): Long =
+            withHandle { handle ->
+                deleteSkippedBundlesNative(handle, roundId, keepCount).also { deletedRows ->
+                    check(deletedRows >= 0) {
```
noop-sk: **Redundant check**

The native side returns `-1` only via `unwrap_exc_or` after a JNI exception is already pending, meaning `withContext` will have thrown before reaching this line. The guard is harmless but misleading — either remove it or add a comment explaining it is a last-resort defensive check for a path that cannot be reached under normal JNI error semantics.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    suspend fun clearRound(roundId: String)
+
+    suspend fun deleteSkippedBundles(
+        roundId: String,
+        keepCount: Int
+    ): Long
+}
+
+data class VoteRecord(
```
noop-sk: **Wrong package for a domain model**

`VoteRecord` is a domain model sitting in the interface file. It belongs in `model/voting/` alongside `FfiVotingModels.kt`, consistent with the rest of the SDK.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
```diff
@@ -0,0 +1,76 @@
...
+
+import cash.z.ecc.android.sdk.internal.jni.VotingRustBackend
+import cash.z.ecc.android.sdk.internal.model.voting.FfiRoundState
+import cash.z.ecc.android.sdk.internal.model.voting.FfiRoundSummary
+import org.json.JSONArray
+
+@Suppress("TooManyFunctions", "LongParameterList")
+class TypesafeVotingBackendImpl : TypesafeVotingBackend {
+    override suspend fun openVotingDb(dbPath: String, walletId: String): TypesafeVotingDb =
+        TypesafeVotingDbImpl(VotingRustBackend.new().openVotingDb(dbPath, walletId))
```
noop-sk: **Unnecessary `VotingRustBackend` instantiation per call**

`VotingRustBackend.new()` calls `RustBackend.loadLibrary()` on every `openVotingDb` invocation. While idempotent once loaded, it is unnecessary overhead. `VotingRustBackend` is stateless — consider making it a singleton or injecting it into `TypesafeVotingBackendImpl`.

backend-lib/src/main/rust/voting/db.rs
```diff
@@ -0,0 +1,78 @@
+use super::*;
+
+static NEXT_DB_HANDLE: AtomicI64 = AtomicI64::new(1);
+static DB_REGISTRY: OnceLock<Mutex<HashMap<jlong, Arc<VotingDb>>>> = OnceLock::new();
```
noop-sk, line 4: @nuttycom — could you sanity-check the registry pattern here? `Arc<VotingDb>` is cloned out of the `Mutex<HashMap>` on every JNI call, so concurrent calls from different threads each hold a live `Arc` and can call `VotingDb` methods in parallel. Is `VotingDb` internally synchronized (own `Mutex`, SQLite connection-level locking, etc.), or does correct usage require single-threaded access that the Kotlin-side `Mutex` is meant to guarantee?

backend-lib/src/main/rust/voting/db.rs
```diff
@@ -0,0 +1,78 @@
...
+
+fn registry() -> &'static Mutex<HashMap<jlong, Arc<VotingDb>>> {
+    DB_REGISTRY.get_or_init(|| Mutex::new(HashMap::new()))
+}
+
+fn next_handle() -> anyhow::Result<jlong> {
+    NEXT_DB_HANDLE
+        .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |id| {
```
noop-sk, line 12: @nuttycom — `Ordering::Relaxed` on both the success and failure orderings of `fetch_update` is correct for a monotone uniqueness counter on its own, since the registry `Mutex` provides the necessary acquire/release barriers around the actual map insert. Just flagging for a second pair of eyes given the subtle interaction between the two synchronization primitives.

backend-lib/src/main/rust/voting/rounds.rs
```diff
@@ -0,0 +1,144 @@
...
+    _: JClass<'local>,
+    db_handle: jlong,
+    round_id: JString<'local>,
+    snapshot_height: jlong,
+    ea_pk: JByteArray<'local>,
+    nc_root: JByteArray<'local>,
+    nullifier_imt_root: JByteArray<'local>,
+    session_json: JString<'local>,
+) {
+    let res = catch_unwind(&mut env, |env| {
```
noop-sk, line 19: @nuttycom — `catch_unwind` is used consistently across all JNI entry points in this file (and in `db.rs`, `share_tracking.rs`). Worth confirming that the `zcash_voting` crate's storage layer does not use types that are `!UnwindSafe` in ways that could leave internal state corrupt after a caught panic — SQLite connections in particular can have tricky unwind behaviour depending on the driver.

backend-lib/src/main/rust/voting/helpers.rs
greg0x: 662a8713

backend-lib/src/main/rust/voting/rounds.rs
greg0x: 2098c324

backend-lib/src/main/rust/voting/json.rs
greg0x: 4447de69

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/FfiVotingModels.kt
greg0x: 7cd088b7

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
greg0x: fe73e0fe

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
greg0x: e7307118

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
greg0x: 3c8bd285

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
greg0x: This is an internal api contract between rust and kotlin. For this it's better to fail as soon as possible so the issue comes out during dev,  failing silently.

The phase set can also be considered stable.

p.s.: I wish we could enjoy the comfort of UniFFI which solves this problem at generation time.

backend-lib/src/main/rust/voting/db.rs
greg0x: Safe. VotingDb wraps the connection in `std::sync::Mutex` ([storage/mod.rs:74-79](https://github.com/valargroup/zcash_voting/blob/b9acd1c197c6ef3425287b269ac7ac82d9a01e62/zcash_voting/src/storage/mod.rs#L74-L79)), so concurrent `Arc<VotingDb>` clones serialize inside zcash_voting; multi-step ops like `get_round_state` hold the guard for the whole call.

backend-lib/src/main/rust/voting/rounds.rs
greg0x: `zcash_voting` uses `rusqlite` for SQLite, and `VotingDb` keeps the connection behind `std::sync::Mutex`. Normal SQLite/storage failures return `VotingError`. If an unexpected panic occurs while the
connection guard is held, the mutex is poisoned, so later calls fail through the JNI exception path rather than silently continuing over that DB handle.

Longer term, `zcash_voting` should probably convert poisoned mutex / unset wallet id cases into explicit `VotingError::Internal` values instead  of relying on panics.

backend-lib/src/main/rust/voting/db.rs
greg0x: Confirm, that the above is correct.

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -0,0 +1,107 @@
...
+        ],
+    )?;
+    Ok(obj.into_raw())
+}
+
+pub(super) fn round_exists(db: &VotingDb, round_id: &str) -> anyhow::Result<bool> {
+    let conn = db.conn();
+    let wallet_id = db.wallet_id();
+    match conn.query_row(
+        "SELECT 1 FROM rounds WHERE round_id = ?1 AND wallet_id = ?2 LIMIT 1",
```
nullcopy: Please don't reach into SQL here (brittle). Please expose a `has_round` helper like that in `zcash_voting::storage` or something

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/FfiVotingModels.kt
```diff
@@ -0,0 +1,55 @@
+package cash.z.ecc.android.sdk.internal.model.voting
+
+import androidx.annotation.Keep
+
+// Must match PHASE_* constants in backend-lib/src/main/rust/voting/json.rs.
+internal const val FFI_ROUND_PHASE_INITIALIZED = 0
```
nullcopy: The `FFI_*/Ffi*` naming throughout this PR breaks from the `JNI_*/Jni*` convention in the rest of the codebase

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
```diff
@@ -8,6 +17,88 @@ class VotingRustBackend private constructor() {
...
+                snapshotHeight,
+                eaPK,
+                ncRoot,
+                nullifierIMTRoot,
+                sessionJson
+            )
+        }
+
+        @Throws(RuntimeException::class)
+        suspend fun getRoundState(roundId: String): FfiRoundState? =
```
nullcopy: Good feedback from Claude:
```
Mixed JSON / native return strategy for collections

  getRoundState returns a constructed FfiRoundState object across JNI, but listRoundsJson and getVotesJson return JSON strings that the Typesafe layer parses. That's two distinct
  serialization mechanisms, with the "Json" suffix leaking into the public Kotlin surface of VotingRustBackend.

  Two consistent options:
  - Use JSON for both (and rename getRoundState → getRoundStateJson). Simple, but you lose the typed FFI object.
  - Use constructed objects for both, using utils::rust_vec_to_java (backend-lib/src/main/rust/utils.rs:70) which already exists for exactly this pattern. Costs a bit more JNI plumbing per
   type but removes the JSON dependency and the Math.toIntExact(getLong(...)) re-validation step in TypesafeVotingBackendImpl.kt.

  Pick one. The hybrid is the worst of both for maintenance.
```

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
nullcopy: > p.s.: I wish we could enjoy the comfort of UniFFI which solves this problem at generation time.

@noop-sk Is this something we should track as a future improvement?

nullcopy (changes requested):
Reviewed 3c8bd285ce0c1182bfd80bfae3fe7ff6692789db

backend-lib/src/main/rust/voting/db.rs
nullcopy: multi-threaded usage is safe :+1:

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/FfiVotingModels.kt
greg0x: 450fe29f397f

backend-lib/src/main/rust/voting/helpers.rs
greg0x: dad02c384595

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
greg0x: c07f8e79491f

nullcopy (approved):
utACK 450fe29f397fae05c107e6f130357de6fbfce485

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
noop-sk: agree

PR #1942, opened 2026-05-10
[#1924] Add Android voting delegation prep and signing

greg0x:
Depends on zcash/zcash-android-wallet-sdk#1938 (second slice — voting DB handle and round lifecycle).
Resolves MOB-1109

## Why

Third slice of the Android shielded-voting integration. #1938 added the voting DB handle and round lifecycle; this PR adds the prep and signing surface needed before delegation proof generation: bundle setup with per-bundle weights, hotkey derivation, governance PCZT construction, sighash extraction, and spend-auth signature extraction. Proving-cache warmup is included so callers can amortize key loading before the first proof.

Vertical SDK slice — each new JNI symbol has a matching Rust handler, Kotlin `external fun`, FFI model, and typesafe wrapper. No app-side wiring yet — that follows once the proof-generation slice lands.

## What's in

Rust / JNI (`backend-lib/src/main/rust/voting/`):

- `notes.rs` — `computeBundleSetup`, `setupBundles`, `generateHotkey`
- `rounds.rs` — `getBundleCount` (added to existing module)
- `delegation.rs` — `buildGovernancePcztJson`, `extractPcztSighash`, `extractSpendAuthSig`
- `util.rs` — `warmProvingCaches`
- `helpers.rs` — width-checked casts (`u32_to_jint`, `u64_to_jlong`, `usize_to_jint`), `network_from_id`, `orchard_fvk_bytes`, `hotkey_orchard_raw_address_from_wallet_seed`, `coin_type_for`, `NU6_BRANCH_ID`, FFI builders for `FfiVotingHotkey` and `FfiBundleSetupResult`, round-phase guards (`update_round_phase_forward`, `require_round_phase_for_delegation_construction`), bundle helpers (`bundle_setup_from_notes`, `bundled_notes_for_index`)
- `json.rs` — `JsonNoteInfo`, `JsonGovernancePczt` (trimmed envelope), hex enc/dec, `json_from_jstring`

Bundle note identity persistence and validation are owned by `zcash_voting 0.5.8` (`VotingDb::setup_bundles` + `build_governance_pczt`) — the Android side does not maintain a sidecar table.

Kotlin (`sdk-lib/.../internal/`):

- `jni/VotingRustBackend.kt` — `external fun` declarations for the new surface
- `TypesafeVotingBackend.kt` + impl — `suspend` wrappers, `native { … }` helper that loads the library and dispatches on `Dispatchers.IO`, JSON → `GovernancePcztResult` parsing using the SDK's existing `String.fromHex` extension
- `model/voting/FfiVotingModels.kt` — `FfiBundleSetupResult`, `FfiVotingHotkey` (`publicKey` + `address` only — secret key stays in Rust), `HotkeyPublicKey` (size-validated constructor)
- `jni/JniConstants.kt` — `JNI_HOTKEY_PUBLIC_KEY_BYTES_SIZE`
- `ext/BlockExt.kt` — `String.fromHex()` made strict (rejects odd length and non-hex characters) now that it's on the JNI parse path

## Not included

- PIR precompute
- delegation proof generation
- delegation submission (Keystone or otherwise)
- vote tree sync / VAN witnesses / vote commitment
- recovery and share-tracking state
- public Synchronizer voting API

## Review focus

Primary review track: core-dev. Please look at:

- **Governance PCZT JSON envelope** — only `pczt_bytes` / `rk` / `action_index` / `pczt_sighash` cross JNI. The rest of the delegation-data fields (`alpha`, `rseed_*`, `padded_note_secrets`, `gov_nullifiers`, …) are persisted by `zcash_voting`'s `store_delegation_data` and read back in the proof-generation slice.
- **Byte-boundary validation** — `java_bytes`, `java_bytes_at_least`, `java_bytes32`, fixed-width `require_len` / `require_min_len`, and width-checked `u32`/`u64` ↔ `jint`/`jlong` casts.
- **Hotkey derivation and surface** — `hotkey_orchard_raw_address_from_wallet_seed` derives the Orchard raw address from the wallet seed at the supplied `account_index`, using `Scope::External` at index 0. The 32-byte hotkey secret key is intentionally **not** surfaced across JNI — it stays inside `VotingDb` and is used internally during PCZT construction / proof generation.
- **Round-phase guards** —
  - `update_round_phase_forward` uses an atomic conditional UPDATE (`WHERE phase < ?1`), so two concurrent callers cannot both advance from the same starting phase; idempotent re-advances return `Ok`, regressions return `Err`.
  - `buildGovernancePczt` is gated by `require_round_phase_for_delegation_construction`: requires `HotkeyGenerated`, rejects if the round has already advanced past `DelegationConstructed`.
- **`setupBundles` parity / retry contract** — the chunker output is recomputed in Rust and compared against `VotingDb::setup_bundles`' return values to catch divergence; on mismatch the JNI returns `Err`. The DB write has already happened at that point, so callers must `clearRound` and retry; this is a hard-stop bug indicator, not a recoverable error.
- **`extractSpendAuthSig` failure mode** — returns a descriptive `Err` if the action at `action_index` has no `spend_auth_sig` (e.g. PCZT not yet signed) or `action_index` is out of bounds; never silently returns zeroed bytes.
- **`native { }` helper in `TypesafeVotingBackendImpl.kt`** — calls `RustBackend.loadLibrary()` before every dispatched block; idempotent and matches the pattern in `RustDerivationTool` / `RustEip681Tool` / `TorClient`.
- Kotlin surface is intentionally `internal` and minimal.

Swift counterpart:

- zcash/zcash-swift-wallet-sdk#1723 (delegation workflow FFI)

## Validation

- `cargo check --manifest-path backend-lib/Cargo.toml --locked` (pinned to `zcash_voting 0.5.8`)
- `./gradlew checkProperties`
- `git diff --check`

# Author

# Reviewer


backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -72,20 +209,355 @@ pub(super) fn make_ffi_round_state<'local>(
...
+    let hash = state.finalize();
+    let mut out = [0u8; NOTE_IDENTITY_HASH_BYTES];
+    out.copy_from_slice(hash.as_bytes());
+    out
+}
+
+pub(super) fn init_voting_android_tables(db: &VotingDb) -> anyhow::Result<()> {
+    let conn = db.conn();
+    conn.execute_batch(
+        "CREATE TABLE IF NOT EXISTS android_voting_bundle_note_identity (
```
p0mvn: Is this a general pattern in this repository to prefix tables with `android_`?

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -72,20 +209,355 @@ pub(super) fn make_ffi_round_state<'local>(
...
+        &[
+            JValue::Int(u32_to_jint(count, "bundle_count")?),
+            JValue::Long(u64_to_jlong(weight, "eligible_weight")?),
+            JValue::Object(&weights_array_obj),
+        ],
+    )?;
+    Ok(obj.into_raw())
+}
+
+pub(super) fn bundle_setup_from_notes(notes: &[NoteInfo]) -> anyhow::Result<(u32, u64, Vec<u64>)> {
```
p0mvn, line 411: Should we have any function specs to explain what these do?

backend-lib/Cargo.toml
```diff
@@ -54,8 +54,11 @@ anyhow = "1"
 jni = { version = "0.21", default-features = false }
 uuid = "1"
 bitflags = "2"
+blake2b_simd = "1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
+# Explicit dependency for FFI-layer voting JSON encoding/decoding of binary fields.
+hex = "0.4"
```
p0mvn: I don't think we should be leaving an explicit comment since that isn't a pattern in this file

backend-lib/src/main/rust/voting/delegation.rs
```diff
@@ -0,0 +1,184 @@
...
+                &hotkey_raw_address,
+                nu6_branch_id(),
+                network.coin_type(),
+                &seed_fingerprint,
+                account_index,
+                &round_name,
+                address_index,
+            )
+            .map_err(|e| anyhow!("build_governance_pczt: {}", e))?;
+        update_round_phase_forward(&db, &round_id, RoundPhase::DelegationConstructed)?;
```
czarcas7ic, line 76: Can we require the round to already be HotkeyGenerated before this advances it to DelegationConstructed? Currently, a caller can skip generateHotkey, and then the hotkey step can no longer be completed for that round.

backend-lib/src/main/rust/voting/notes.rs
```diff
@@ -0,0 +1,84 @@
...
+        if count != expected_count || weight != expected_weight {
+            return Err(anyhow!(
+                "setup_bundles result mismatch: db=({}, {}) chunk=({}, {})",
+                count,
+                weight,
+                expected_count,
+                expected_weight
+            ));
+        }
+        store_bundle_note_identities(&db, &round_id, &notes)?;
```
czarcas7ic: Can we make the bundle setup and identity writes atomic? If setup_bundles succeeds but store_bundle_note_identities fails or the app is killed before it runs, the round is left with bundles that later cannot pass the identity check.

backend-lib/src/main/rust/voting/notes.rs
```diff
@@ -0,0 +1,84 @@
...
+        let notes: Vec<NoteInfo> = json_notes
+            .into_iter()
+            .map(NoteInfo::try_from)
+            .collect::<anyhow::Result<_>>()?;
+        let (expected_count, expected_weight, bundle_weights) = bundle_setup_from_notes(&notes)?;
+        let round_id = java_string_to_rust(env, &round_id)?;
+        let (count, weight) = db
+            .setup_bundles(&round_id, &notes)
+            .map_err(|e| anyhow!("setup_bundles: {}", e))?;
+        if count != expected_count || weight != expected_weight {
```
noop-sk, line 49: **Partial DB write on parity mismatch**

`db.setup_bundles()` has already committed its write when this mismatch check fires. Returning `Err` here leaves the DB in a state where bundles are written but the caller receives an error — meaning a retry would attempt to set up the same round again. The PR description acknowledges this ("treat the mismatch as a bug, not a recoverable error"), but callers need to know they must `clearRound` before retrying. Consider making that contract explicit in a comment or error message.

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -1,15 +1,45 @@
...
-pub(super) const SHARE_NULLIFIER_BYTES: usize = 32;
+pub(super) const SHARE_NULLIFIER_BYTES: usize = PROTOCOL_FIELD_BYTES;
+pub(super) const HOTKEY_SECRET_KEY_BYTES: usize = PROTOCOL_FIELD_BYTES;
+pub(super) const HOTKEY_PUBLIC_KEY_BYTES: usize = PROTOCOL_FIELD_BYTES;
+pub(super) const NETWORK_ID_TESTNET: jint = 0;
+pub(super) const NETWORK_ID_MAINNET: jint = 1;
+const NOTE_IDENTITY_HASH_BYTES: usize = PROTOCOL_FIELD_BYTES;
+const NOTE_IDENTITY_DOMAIN: &[u8] = b"zcash-android-voting-note-v1";
 
 pub(super) fn jint_to_u32(value: jint, field: &str) -> anyhow::Result<u32> {
```
noop-sk, line 45: **Cross-platform domain string concern**

`NOTE_IDENTITY_DOMAIN` is defined here in the Android SDK codebase, not in `zcash_voting` or a shared crate. If the note identity hash is ever compared or validated cross-platform (iOS SDK, server-side), the domain string must match exactly. Is this hash Android-only, or does it need to be consistent with other clients? If the latter, it belongs in `zcash_voting`.

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -72,20 +209,355 @@ pub(super) fn make_ffi_round_state<'local>(
...
+    update_hash_with_len_prefixed_bytes(&mut state, &note.rseed);
+    update_hash_with_len_prefixed_bytes(&mut state, note.ufvk_str.as_bytes());
+
+    let hash = state.finalize();
+    let mut out = [0u8; NOTE_IDENTITY_HASH_BYTES];
+    out.copy_from_slice(hash.as_bytes());
+    out
+}
+
+pub(super) fn init_voting_android_tables(db: &VotingDb) -> anyhow::Result<()> {
```
noop-sk: **Android-specific table with FK into `zcash_voting` schema**

`android_voting_bundle_note_identity` references `bundles(round_id, wallet_id, bundle_index)` — a table owned by the `zcash_voting` crate. If the crate renames or restructures `bundles`, this FK silently breaks at runtime (SQLite FK violations are not enforced unless `PRAGMA foreign_keys = ON`). Is FK enforcement enabled on the `VotingDb` connection? If not, this reference provides no actual safety.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/ext/BlockExt.kt
```diff
@@ -2,24 +2,35 @@ package cash.z.ecc.android.sdk.ext
...
 }
 
-// Not used within the SDK, but is used by the Wallet app
-@Suppress("unused", "MagicNumber")
+@Suppress("MagicNumber")
 fun String.fromHex(): ByteArray {
+    require(length % HEX_CHARS_PER_BYTE == 0) {
+        "Hex string must have an even length, got $length"
+    }
+
```
noop-sk, line 21: **Breaking change for existing callers**

The old `fromHex` was silent on invalid input (produced garbage bytes). This PR makes it strict — odd-length or non-hex input now throws `IllegalArgumentException` via `require`. The `@Suppress("unused")` removal confirms the wallet app uses this. Any call site passing unchecked input will now crash at runtime instead of producing wrong data. Intended, but worth a callout for the wallet app side — callers that previously relied on silent coercion need auditing.

backend-lib/src/main/rust/voting/delegation.rs
```diff
@@ -0,0 +1,240 @@
...
+        let seed_bytes =
+            java_secret_bytes_at_least(env, &wallet_seed, "walletSeed", PROTOCOL_FIELD_BYTES)?;
+        let derived_fvk_bytes =
+            orchard_fvk_bytes_from_wallet_seed(seed_bytes.expose_secret(), network, account_index)?;
+        if derived_fvk_bytes != fvk_bytes {
+            return Err(anyhow!(
+                "ufvk does not match walletSeed for network_id={network_id} account_index={account_index}"
+            ));
+        }
+        let hotkey_raw_address = hotkey_orchard_raw_address_from_wallet_seed(
```
noop-sk, line 43: @nuttycom — `hotkey_orchard_raw_address_from_wallet_seed` always uses `Scope::External`. The PR description says "using `Scope::External` at index 0", but `address_index` is a caller-supplied parameter (not hardcoded to 0). Is `Scope::External` the correct scope for all hotkey address indices, or should this be `Scope::Internal` for any index range? Want to confirm the derivation path is intentional.

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -72,20 +209,355 @@ pub(super) fn make_ffi_round_state<'local>(
...
+            JValue::Long(u64_to_jlong(state.snapshot_height, "snapshot_height")?),
             JValue::Object(&hotkey_obj),
             JValue::Object(&weight_obj),
             JValue::Bool(state.proof_generated as jboolean),
         ],
     )?;
     Ok(obj.into_raw())
 }
+
+pub(super) fn make_ffi_voting_hotkey<'local>(
```
noop-sk: @nuttycom — `hotkey.secret_key` is wrapped in `SecretVec` for the length check, then immediately `expose_secret()` is passed to `env.byte_array_from_slice()`. The secret key bytes end up as a plain Kotlin `ByteArray` on the JVM heap with no zeroization mechanism. Is this acceptable for the hotkey secret key, or should it be treated with the same care as the wallet seed (i.e. consumed immediately and never surfaced to app code)?

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -72,20 +209,355 @@ pub(super) fn make_ffi_round_state<'local>(
...
+        rusqlite::params![round_id, wallet_id],
+        |_| Ok(()),
+    ) {
+        Ok(()) => Ok(true),
+        Err(rusqlite::Error::QueryReturnedNoRows) => Ok(false),
+        Err(e) => Err(anyhow!("round_exists query failed: {}", e)),
+    }
+}
+
+pub(super) fn update_round_phase_forward(
```
noop-sk, line 450: @nuttycom — `update_round_phase_forward` does a read → check → write without holding a DB transaction. Two concurrent `buildGovernancePczt` calls on the same round could both see `Initialized`, both proceed past the check, and both attempt to write `DelegationConstructed`. Is `VotingDb` serializing writes internally (e.g. connection-level mutex), or does the Kotlin `accessMutex` in `VotingDb` guarantee single-threaded access here?

backend-lib/Cargo.toml
greg0x: 514a1cc3

backend-lib/src/main/rust/voting/delegation.rs
greg0x: 51786ec1

backend-lib/src/main/rust/voting/notes.rs
greg0x: 3b738fae

backend-lib/src/main/rust/voting/helpers.rs
greg0x: 563de2ad

backend-lib/src/main/rust/voting/helpers.rs
greg0x: Good catch! I will move this into `zcash_voting` before merging this PR

greg0x: 304a897631c6 - this is the primary fix 

I will update the Cargo.lock when `zcash_voting@5.3.8` is out https://github.com/valargroup/zcash_voting/pull/46

greg0x: 6a1a45a7b5e2

backend-lib/src/main/rust/voting/notes.rs
greg0x: 82e00efa

backend-lib/src/main/rust/voting/delegation.rs
greg0x: ceb81b9a

We reconstruct this address in the vote-proof builder, and it uses [address_at(address_index, Scope::External](https://github.com/valargroup/voting-circuits/blob/1b0a0891a185b81c2a7b170cdf69cad544804f52/voting-circuits/src/vote_proof/builder.rs#L463). The Android SDK must derive the delegation PCZT address with the same (address_index, Scope::External) pair, otherwise ZKP 2 reconstructs a different address than the one committed during delegation.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/ext/BlockExt.kt
greg0x: 313157f7

backend-lib/src/main/rust/voting/helpers.rs
greg0x: Moved this concern into `zcash_voting`

https://github.com/zcash/zcash-android-wallet-sdk/commit/304a897631c6d1726a9ba22964836069b99916ab

https://github.com/valargroup/zcash_voting/compare/zcash_voting-v0.5.7...zcash_voting-v0.5.8

backend-lib/src/main/rust/voting/helpers.rs
greg0x: This has also been dropped with moving the table into `zcash_voting` 

https://github.com/valargroup/zcash_voting/blob/b9acd1c197c6ef3425287b269ac7ac82d9a01e62/zcash_voting/src/storage/queries.rs#L10

backend-lib/src/main/rust/voting/helpers.rs
greg0x: 2cc58225

Nice catch

backend-lib/src/main/rust/voting/helpers.rs
greg0x: d808d3fb5c96

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
```diff
@@ -1,6 +1,45 @@
 package cash.z.ecc.android.sdk.internal.model.voting
 
 import androidx.annotation.Keep
+import cash.z.ecc.android.sdk.internal.jni.JNI_HOTKEY_PUBLIC_KEY_BYTES_SIZE
+
+@Keep
```
noop-sk: **Misleading `@Keep` on `HotkeyPublicKey`**

`HotkeyPublicKey` is constructed entirely in Kotlin — `JniVotingHotkey`'s secondary constructor calls `HotkeyPublicKey.new(pk)`. JNI never looks up this class by name or calls its constructor via reflection. `@Keep` conventionally signals "JNI constructs this class at runtime"; here it is used defensively against R8 shrinking, which is fine, but the signal is misleading. Because `JniVotingHotkey` (which is `@Keep`) holds a `publicKey: HotkeyPublicKey` field, R8 will not strip `HotkeyPublicKey` anyway — the `@Keep` is redundant. Either remove it or replace with a comment explaining the R8 concern.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
```diff
@@ -74,6 +110,16 @@ class VotingRustBackend private constructor() {
         suspend fun listRounds(): Array<JniRoundSummary> =
             withHandle { handle -> listRoundsNative(handle) }
 
+        @Throws(RuntimeException::class)
+        suspend fun getBundleCount(roundId: String): Int =
+            withHandle { handle ->
+                getBundleCountNative(handle, roundId).also { count ->
+                    check(count >= 0) {
```
noop-sk: **Redundant guard after native call**

`getBundleCountNative` on the Rust side calls `unwrap_exc_or(&mut env, res, -1)`: if the operation fails, a JVM exception is already pending and the -1 sentinel is returned. Kotlin will throw that exception before executing `check(count >= 0)`, so the check can never fire. No other `withHandle` wrapper in this class adds a post-native Kotlin guard. Remove the `.also { count -> check(...) }` to stay consistent.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
```diff
@@ -9,6 +48,16 @@ internal const val JNI_ROUND_PHASE_DELEGATION_CONSTRUCTED = 2
 internal const val JNI_ROUND_PHASE_DELEGATION_PROVED = 3
 internal const val JNI_ROUND_PHASE_VOTE_READY = 4
 
+@Keep
+data class JniBundleSetupResult(
+    val bundleCount: Int,
+    val eligibleWeight: Long,
+    val bundleWeights: List<Long> = emptyList()
```
noop-sk: **`bundleCount` and `bundleWeights.size` can diverge**

The primary constructor allows `JniBundleSetupResult(bundleCount = 2, eligibleWeight = 75_000_000L)` without supplying `bundleWeights`, leaving `bundleWeights.size == 0` while `bundleCount == 2`. The class has no `init` invariant guarding this. Since both constructors are `internal` only SDK-internal code is affected, but test helpers that construct this directly can produce inconsistent instances silently. Consider adding `init { require(bundleWeights.size == bundleCount) { ... } }` or removing the default.

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -213,3 +339,241 @@ impl TryFrom<VoteRecord> for JniVoteRecordPayload {
...
+    )?);
+    let public_key = require_len(
+        hotkey.public_key,
+        "hotkey_public_key",
+        HOTKEY_PUBLIC_KEY_BYTES,
+    )?;
+    let pk_obj: JObject<'local> = env.byte_array_from_slice(&public_key)?.into();
+    let addr_obj: JObject<'local> = env.new_string(&hotkey.address)?.into();
+    let obj = env.new_object(
+        &class,
```
noop-sk, line 373: **Hardcoded JNI constructor signatures**

`make_jni_voting_hotkey` passes `"([BLjava/lang/String;)V"` and `make_jni_bundle_setup_result` (further below) passes `"(IJ[J)V"` as literal strings to `env.new_object`. If the corresponding Kotlin constructors gain or reorder parameters — e.g., adding a field to `JniVotingHotkey` — the mismatch produces a silent `NoSuchMethodError` at runtime that only surfaces when those JNI paths are exercised. The existing JNI helpers in this codebase use the same pattern, but each new class multiplies the surface. Consider at least a cross-reference comment tying each signature string to its Kotlin counterpart.

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -213,3 +339,241 @@ impl TryFrom<VoteRecord> for JniVoteRecordPayload {
...
 }
+
+/// Builds the Kotlin hotkey JNI model after enforcing the expected key widths.
+/// The secret key is intentionally not surfaced across JNI.
+pub(super) fn make_jni_voting_hotkey<'local>(
+    env: &mut JNIEnv<'local>,
+    hotkey: voting::types::VotingHotkey,
+) -> anyhow::Result<jobject> {
+    let class = env.find_class("cash/z/ecc/android/sdk/internal/model/voting/JniVotingHotkey")?;
+    let _secret_key = SecretVec::new(require_len(
```
noop-sk: @nullcopy  — `let _secret_key = SecretVec::new(require_len(hotkey.secret_key, ...)?)` wraps the raw secret key bytes and holds them alive until `make_jni_voting_hotkey` returns (correct — `let _name` keeps the binding live unlike `let _ =`). Can you confirm that `SecretVec::drop` + `zeroize` runs before the JNI frame is torn down, and that the bytes aren't copied onto the stack by `require_len` before the `SecretVec` wrapper takes ownership?

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -213,3 +339,241 @@ impl TryFrom<VoteRecord> for JniVoteRecordPayload {
...
+    chunk_result
+        .bundles
+        .get(bundle_index)
+        .cloned()
+        .ok_or_else(|| anyhow!("bundle_index {bundle_index} is not present in note bundle set"))
+}
+
+/// Advances a round phase without allowing regressions; equal phases are
+/// treated as idempotent.
+pub(super) fn update_round_phase_forward(
```
noop-sk, line 450: @nullcopy — `update_round_phase_forward` does a conditional `UPDATE … WHERE phase < ?1` (atomic in SQLite), then reads back the current phase only if 0 rows were affected. For a single `VotingDb` connection this is safe, but `openVotingDb` can create multiple `Arc<VotingDb>` instances pointing at the same SQLite file — each with its own `accessMutex`. Two `VotingDb` handles from different `VotingRustBackend` instances could race through `build_governance_pczt` concurrently. Does `VotingDb` serialize writers internally, or does the caller need to guarantee a single open handle per file?

backend-lib/src/main/rust/voting/util.rs
```diff
@@ -0,0 +1,15 @@
+use super::*;
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_warmProvingCachesNative<
```
noop-sk, line 4: @nullcopy — `warm_proving_caches()` is called synchronously on `Dispatchers.IO` without any cancellation or progress signal. If two callers race (e.g., `warmProvingCaches` from one coroutine and `buildGovernancePczt` before warming finishes), is the underlying proving-key loading thread-safe? Is it safe to call `warm_proving_caches` concurrently or while `build_governance_pczt` is already running?

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
greg0x: 44512264

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
greg0x: 6ec41d94

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
greg0x: a78ba37d

backend-lib/src/main/rust/voting/helpers.rs
greg0x: 27f385b5

backend-lib/src/main/rust/voting/helpers.rs
greg0x: f553ca89

backend-lib/src/main/rust/voting/helpers.rs
greg0x: 981ca368

This code path should be theoretical, but better to harden the code now.

backend-lib/src/main/rust/voting/util.rs
greg0x: Yes. The proving-key caches are backed by process-lifetime Rust OnceLocks, so concurrent warm-up calls and first proof use racing with warm-up are synchronized by get_or_init. build_governance_pczt itself does not consume those proving-key caches. The Android wrapper is synchronous and has no cancellation/progress semantics for now, so callers should treat it as a best-effort background warm-up.

backend-lib/src/main/rust/voting/helpers.rs
nullcopy: The change looks good to me

nullcopy (approved):
utACK 981ca36865c216a5821f6f08c70c25609a8868db

PR #1950, opened 2026-05-12
[#1924] Add Android voting delegation proof and submission

greg0x:
Stacked on zcash/zcash-android-wallet-sdk#1942 (PR 4 — delegation setup + governance PCZT prep).
Resolves MOB-1109

> Draft: still iterating locally; opened early so reviewers can start tracking the delegation finish.

## Why

Fifth slice of the Android shielded-voting integration. PR 4 (#1942) lands the
delegation setup primitives, hotkey, and governance PCZT/sighash prep. This PR
finishes the delegation half of the workflow: enabling `client-pir`, persisting
witnesses, running PIR precompute, generating the delegation proof with progress
reporting, and producing the signed submission payload (including the
Keystone-supplied-signature variant).

No public `Synchronizer` voting API is added.

## What's in

- `client-pir` feature on `zcash_voting` is now enabled; lockfile delta is mostly
  the PIR transport stack (`pir-client`, `pir-types`, `valar-ypir`,
  `valar-spiral-rs`, `hyper-rustls`).
- New JNI surface for the delegation finish: witness storage, PIR precompute,
  proof + progress callback, signed submission, Keystone-signature submission.
- New `voting/progress.rs` module hosting `JniProgressReporter`; the only
  per-thread JNI attach pattern in the voting backend so far.

## On the `VotingDbBackend` seam

The typesafe Kotlin layer gains an internal `VotingDbBackend` interface and a
pass-through `RustVotingDbBackend` wrapper so `TypesafeVotingBackendImpl` is
unit-testable without JNI. It lands here because PR 5 is the first slice with
byte-array-typed result models that need size/count validation; PR 6–8 will
extend the same surface as more typed FFI arrives, so introducing it now
avoids a retroactive horizontal refactor later. Happy to split into a tiny
infra PR if a reviewer prefers.

## Not included

- Snapshot wallet notes / note witnesses (PR 6).
- Vote tree / VAN / vote commitment / cast-vote payloads (PR 7).
- Delegation tx hash store/lookup and recovery state (PR 8).

## Review focus

- `zcash_voting` `client-pir` feature transitive graph (`Cargo.lock` delta).
- PIR client construction (`connect_pir_client`) and SDK boundary for network
  errors.
- `progress.rs` threading and lifetime — `attach_current_thread` per callback,
  `GlobalRef` ownership, exception-clear after failed Java call.
- `verify_delegation_submission_sig` semantics for the Keystone variant: `rk`
  parsed as a `redpallas` `VerificationKey<SpendAuth>`, identity rejected, sig
  re-verified over the supplied sighash.
- `require_witnesses_match_bundle` invariants (count, root, position, duplicate
  commitments) before persisting to the voting DB.
- Re-entry guard on `VotingDb`: progress callbacks must not call back into the
  same handle while the access mutex is held.
- Typed `DelegationProofResult` / `DelegationSubmissionResult` byte/list size
  invariants enforced by the typesafe layer.

Note: `delegationProofResultFixtureNative` and `storeDelegationProofFixtureNative`
are JNI fixtures used by androidTest. Open to moving these behind a debug-only
artifact if reviewers prefer.

## Validation

- `./gradlew checkProperties`
- `git diff --check`

# Author

# Reviewer


sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
```diff
@@ -126,19 +381,160 @@ private class TypesafeVotingDbImpl(
...
         actionIndex = getCheckedInt("action_index")
     )
 
+internal fun JniDelegationPirPrecomputeResult.toDelegationPirPrecomputeResult() =
+    DelegationPirPrecomputeResult(
+        cachedCount = cachedCount,
+        fetchedCount = fetchedCount
+    )
+
+internal fun JniDelegationProofResult.toDelegationProofResult(): DelegationProofResult {
```
noop-sk, line 479: **`proof` not size-validated**

Every other field in this mapper has a `requireByteArraySize` call, but `proof` is passed through unchecked. Rust already validates fixed-length fields before constructing the JNI object, but the Kotlin validation layer here is supposed to be a second line of defence: if native returns a proof with an unexpected length (protocol change, truncated write) it would propagate silently into `DelegationProofResult` and only surface later, far from the source. Add `proof.requireByteArraySize("proof", EXPECTED_PROOF_BYTES)` — the expected size (96) is already observable in the fixture test.

backend-lib/src/main/rust/voting/progress.rs
```diff
@@ -0,0 +1,39 @@
...
+    callback: GlobalRef,
+}
+
+impl ProofProgressReporter for JniProgressReporter {
+    fn on_progress(&self, progress: f64) {
+        match self.vm.attach_current_thread() {
+            Ok(mut env) => {
+                if let Err(e) = env.call_method(
+                    self.callback.as_obj(),
+                    "onProgress",
```
noop-sk: **Hardcoded callback method name**

The `"onProgress"` method name and `"(D)V"` signature are resolved at runtime via reflection. If `VotingProofProgressCallback.onProgress` is renamed or its parameter type changes (e.g. narrowed to `Float`), this call will silently return a JNI error at runtime instead of failing at compile time. The existing constructor-signature constants in `helpers.rs` have accompanying `// Must match ...` comments; the same pattern should be applied here — a named constant plus a cross-reference comment to `VotingProofProgressCallback`.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
```diff
@@ -103,3 +145,89 @@ internal data class GovernancePcztResult(
...
+internal data class DelegationSubmissionResult(
+    val proof: ByteArray,
+    val rk: ByteArray,
+    val spendAuthSig: ByteArray,
+    val sighash: ByteArray,
+    val nfSigned: ByteArray,
+    val cmxNew: ByteArray,
+    val govComm: ByteArray,
+    val govNullifiers: List<ByteArray>,
+    val alpha: ByteArray,
```
noop-sk: @nuttycom — `DelegationSubmissionResult.alpha` exposes a per-signing ephemeral private scalar to Kotlin callers as a plain `ByteArray`. From an app-developer perspective, `alpha` is only ever used during submission assembly on the Rust side; once `rk` and `spendAuthSig` are in the result, `alpha` has no further role. Does the submission format actually require the caller to hold `alpha`, or can it be zeroized in Rust before the struct crosses the JNI boundary? Keeping it on the Kotlin heap (and in `JniDelegationSubmissionResult`) extends its lifetime and prevents zeroization.

backend-lib/src/main/rust/voting/delegation.rs
```diff
@@ -141,10 +144,403 @@ fn extract_indexed_spend_auth_sig(
...
+                witness.position,
+                note.position
+            ));
+        }
+    }
+
+    Ok(())
+}
+
+fn require_round_phase_not_after(
```
noop-sk, line 474: @nuttycom — `require_round_phase_not_after` reads the round phase, checks it, and returns — then `build_and_prove_delegation` writes back. Two concurrent `buildAndProveDelegation` calls on the same round could both pass the phase check before either advances the phase. Is `VotingDb` serializing writes at the storage layer, or does the Kotlin-side `accessMutex` guarantee single-caller access for each `VotingDb` instance? If the mutex is per-instance and each caller holds a different `VotingDb` handle to the same DB file, the TOCTOU is live.

backend-lib/src/main/rust/voting/progress.rs
```diff
@@ -0,0 +1,39 @@
+use super::*;
+
+struct JniProgressReporter {
+    vm: JavaVM,
+    callback: GlobalRef,
+}
+
+impl ProofProgressReporter for JniProgressReporter {
+    fn on_progress(&self, progress: f64) {
+        match self.vm.attach_current_thread() {
```
noop-sk, line 17: @nuttycom — `JniProgressReporter::on_progress` calls `vm.attach_current_thread()` on every progress tick. The PR description notes this is the first per-thread JNI attach pattern in the voting backend. If `build_and_prove_delegation` internally spawns worker threads for proof generation, each such thread would attach and detach on every callback invocation — this can be expensive and may cause issues if `attach_current_thread` returns a daemon-thread-scoped env that outlives the callback scope. Is the proof generation always single-threaded, or can progress callbacks fire from threads not owned by the Kotlin coroutine dispatcher?

backend-lib/src/main/rust/voting/json.rs
```diff
@@ -5,6 +5,7 @@ use serde::{Deserialize, Serialize};
 const NOTE_SCOPE_EXTERNAL: u32 = 0;
 const NOTE_SCOPE_INTERNAL: u32 = 1;
 const ORCHARD_DIVERSIFIER_BYTES: usize = 11;
+const ORCHARD_WITNESS_PATH_DEPTH: usize = 32;
```
greg0x: Note to reviewer: This is rebase / stack PR regression. We are moving towards typed JNI and away from JSON.

backend-lib/src/main/rust/voting/delegation.rs
```diff
@@ -141,10 +144,403 @@ fn extract_indexed_spend_auth_sig(
...
+        let notes: Vec<NoteInfo> = json_notes
+            .into_iter()
+            .map(NoteInfo::try_from)
+            .collect::<anyhow::Result<_>>()?;
+        let json_witnesses: Vec<JsonWitnessData> =
+            json_from_jstring(env, &witnesses_json, "witnessesJson")?;
+        let witnesses: Vec<WitnessData> = json_witnesses
+            .into_iter()
+            .map(WitnessData::try_from)
+            .collect::<anyhow::Result<_>>()?;
```
nullcopy: Do we need to check that `notes.len() == witnesses.len()` here? Or is that validated in `require_witnesses_match_bundle()`?

nullcopy: nvm, I see it is checked

backend-lib/src/main/rust/voting/delegation.rs
nullcopy: Would be good to have a test case either way

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -406,6 +424,159 @@ pub(super) fn make_jni_bundle_setup_result<'local>(
...
+    if values.len() != expected_count {
+        return Err(anyhow!(
+            "{field} must contain {expected_count} entries, got {}",
+            values.len()
+        ));
+    }
+
+    let values = values
+        .into_iter()
+        .enumerate()
```
nullcopy: (nit) I see this common adapter pattern of `.enumerate().map(|...| require_len(..., &format!("{field}[{index}]"), ...).collect()` throughout this PR. This feels like a good opportunity for a helper

nullcopy: not blocking

backend-lib/src/main/rust/voting/json.rs
greg0x: 173060e469c4 has been dropped in this commit

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
greg0x: 88f375a80f0c

I have intentionally not used the length 96 because upstream is not enforcing it either. So I mimicked what upstream does. 

- zcash_voting 0.5.9 exposes DelegationProofResult.proof: Vec<u8> as “Halo2 proof bytes.”
- voting-circuits/Halo2 produces those bytes via transcript serialization.

backend-lib/src/main/rust/voting/progress.rs
greg0x: d1e83bc4

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
greg0x: 9b21b0da

backend-lib/src/main/rust/voting/delegation.rs
greg0x: 4c4f2094

The Kotlin mutex is per-wrapper, so it is not the guarantee here. The JNI layer coalesces same `(dbPath, walletId)` opens into one shared `VotingDbHandle`, and `buildAndProveDelegationNative` holds that handle’s native `access_lock()` from phase check through proof writeback. Added `4c4f2094` to test that same path/wallet opens share the native lock.

backend-lib/src/main/rust/voting/progress.rs
greg0x: Current `zcash_voting 0.5.9` reports progress at coarse milestones outside the spawned Halo2 proving closure, so callbacks are not fired from the proving worker thread today. The bridge still attaches the invoking native thread per callback so it remains correct if that changes later. Documented this in `2d03d475`.

backend-lib/src/main/rust/voting/helpers.rs
greg0x: 69c2cce7

nullcopy (approved):
utACK 23144bbb1aba7678ff64644e0883ddeaed283ff8

PR #1951, opened 2026-05-12
[#1924] Add Android voting snapshot wallet notes and note witnesses

greg0x:
Stacked on zcash/zcash-android-wallet-sdk#1950 (PR 5 — delegation proof + submission).
Resolves MOB-1109

## Why

Sixth slice of the Android shielded-voting integration. PR 5 (#1950) finishes
the delegation half (PIR precompute, proof + progress callback, submission).
This PR introduces the snapshot input-prep path that vote construction will
consume in PR 7: extracting wallet notes at a historical snapshot height,
storing the matching tree state in the voting DB, and generating per-bundle
note witnesses against that snapshot.

No public `Synchronizer` voting API is added. Wallet DB access from the voting
backend goes through a new internal `SdkSynchronizer.getWalletDbPathForVoting()`
rather than a public `getWalletDbPath()` accessor (one of the open decisions in
the split plan).

## Not included

- Vote tree sync / VAN witnesses / vote commitment / cast-vote payloads (PR 7).
- Recovery records and share tracking (PR 8).
- Any new submission or proof-generation behavior.

## Review focus

- `getWalletDbPathForVoting()` as `internal suspend` on `SdkSynchronizer` —
  whether this is the right shape, or whether voting note extraction should go
  through a narrower SDK API instead.
- Read-only wallet DB access from the voting backend, including account UUID,
  UFVK, and network binding for `getWalletNotes`.
- Historical note selection / snapshot anchor correctness in
  `voting/notes.rs`.
- TreeState frontier parsing and root consistency in `extractNcRoot`.
- Witness generation against a stored snapshot, including bundle-specific note
  filtering and stale cached witness replacement behavior.
- JNI byte-boundary validation via the shared helpers in `voting/helpers.rs`
  (Orchard FVK bytes, TreeState bytes, account UUID, witness payload sizes).
- Typed FFI models (`JniNoteInfo`, `JniWitnessData`) and how the typesafe layer
  enforces size/count invariants before handing data back to the SDK.

## Validation

- `./gradlew checkProperties`
- `git diff --check`

# Author

# Reviewer

### DIFF
--- a/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
+++ b/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
@@ -375,7 +375,7 @@ class VotingRustBackendTest {
         }
 
     @Test
-    fun store_tree_state_accepts_cached_bytes() =
+    fun store_tree_state_rejects_invalid_cached_bytes() =
         runTest {
             val db = VotingRustBackend.new().openVotingDb(newDbPath(), WALLET_ID)
             try {
@@ -388,7 +388,30 @@ class VotingRustBackendTest {
                     sessionJson = null
                 )
 
-                db.storeTreeState(ROUND_ID, byteArrayOf(1, 2, 3))
+                assertFailsWith<RuntimeException> {
+                    db.storeTreeState(ROUND_ID, byteArrayOf(1, 2, 3))
+                }
+            } finally {
+                db.close()
+            }
+        }
+
+    @Test
+    fun store_tree_state_accepts_matching_snapshot_fixture() =
+        runTest {
+            val backend = VotingRustBackend.new()
+            val db = backend.openVotingDb(newDbPath(), WALLET_ID)
+            try {
+                db.initRound(
+                    roundId = ROUND_ID,
+                    snapshotHeight = 1,
+                    eaPK = EA_PK,
+                    ncRoot = EMPTY_ORCHARD_WITNESS_ROOT.hexToByteArray(),
+                    nullifierIMTRoot = NULLIFIER_IMT_ROOT,
+                    sessionJson = null
+                )
+
+                db.storeTreeState(ROUND_ID, backend.treeStateFixtureForTesting())
             } finally {
                 db.close()
             }
--- a/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
+++ b/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
@@ -397,6 +397,52 @@ class VotingRustBackendTest {
         }
 
     @Test
+    fun store_tree_state_rejects_snapshot_height_mismatch() =
+        runTest {
+            val backend = VotingRustBackend.new()
+            val db = backend.openVotingDb(newDbPath(), WALLET_ID)
+            try {
+                db.initRound(
+                    roundId = ROUND_ID,
+                    snapshotHeight = SNAPSHOT_HEIGHT,
+                    eaPK = EA_PK,
+                    ncRoot = EMPTY_ORCHARD_WITNESS_ROOT.hexToByteArray(),
+                    nullifierIMTRoot = NULLIFIER_IMT_ROOT,
+                    sessionJson = null
+                )
+
+                assertFailsWith<RuntimeException> {
+                    db.storeTreeState(ROUND_ID, backend.treeStateFixtureForTesting())
+                }
+            } finally {
+                db.close()
+            }
+        }
+
+    @Test
+    fun store_tree_state_rejects_nc_root_mismatch() =
+        runTest {
+            val backend = VotingRustBackend.new()
+            val db = backend.openVotingDb(newDbPath(), WALLET_ID)
+            try {
+                db.initRound(
+                    roundId = ROUND_ID,
+                    snapshotHeight = 1,
+                    eaPK = EA_PK,
+                    ncRoot = NC_ROOT,
+                    nullifierIMTRoot = NULLIFIER_IMT_ROOT,
+                    sessionJson = null
+                )
+
+                assertFailsWith<RuntimeException> {
+                    db.storeTreeState(ROUND_ID, backend.treeStateFixtureForTesting())
+                }
+            } finally {
+                db.close()
+            }
+        }
+
+    @Test
     fun store_tree_state_accepts_matching_snapshot_fixture() =
         runTest {
             val backend = VotingRustBackend.new()
--- a/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
+++ b/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
@@ -22,6 +22,7 @@ class VotingRustBackendTest {
         private const val SHARE_INDEX = 5
         private const val OUT_OF_RANGE_SHARE_INDEX = 16
         private const val DIVERSIFIER_BYTES = 11
+        private const val ORCHARD_FVK_BYTES = 96
         private const val ORCHARD_WITNESS_PATH_DEPTH = 32
         private val VOTE_COMMITMENT = ByteArray(FIELD_BYTES) { 1 }
         private val BLIND = ByteArray(FIELD_BYTES) { 2 }
@@ -125,6 +126,52 @@ class VotingRustBackendTest {
     fun warm_proving_caches_smoke() =
         runTest {
             VotingRustBackend.new().warmProvingCaches()
+        }
+
+    @Test
+    fun extract_orchard_fvk_from_ufvk_is_deterministic() =
+        runTest {
+            val backend = VotingRustBackend.new()
+            val ufvk = deriveTestUfvk()
+
+            val first = backend.extractOrchardFvkFromUfvk(ufvk, TESTNET_NETWORK_ID)
+            val second = backend.extractOrchardFvkFromUfvk(ufvk, TESTNET_NETWORK_ID)
+
+            assertEquals(ORCHARD_FVK_BYTES, first.size)
+            assertContentEquals(first, second)
+            assertFailsWith<RuntimeException> {
+                backend.extractOrchardFvkFromUfvk("not-a-ufvk", TESTNET_NETWORK_ID)
+            }
+        }
+
+    @Test
+    fun extract_nc_root_decodes_tree_state() =
+        runTest {
+            val backend = VotingRustBackend.new()
+
+            val root = backend.extractNcRoot(backend.treeStateFixtureForTesting())
+
+            assertContentEquals(EMPTY_ORCHARD_WITNESS_ROOT.hexToByteArray(), root)
+            assertFailsWith<RuntimeException> {
+                backend.extractNcRoot(byteArrayOf(1, 2, 3))
+            }
+        }
+
+    @Test
+    fun verify_witness_returns_boolean() =
+        runTest {
+            val backend = VotingRustBackend.new()
+            val validWitness = witnesses().single()
+
+            assertTrue(backend.verifyWitness(validWitness))
+            assertFalse(backend.verifyWitness(validWitness.copy(root = NC_ROOT)))
+            assertFailsWith<RuntimeException> {
+                backend.verifyWitness(
+                    validWitness.copy(
+                        authPath = validWitness.authPath.dropLast(1)
+                    )
+                )
+            }
         }
 
     @Test
@@ -322,6 +369,26 @@ class VotingRustBackendTest {
                 val deletedRows = db.deleteSkippedBundles(ROUND_ID, keepCount = 1)
                 assertEquals(1L, deletedRows)
                 assertEquals(1, db.getBundleCount(ROUND_ID))
+            } finally {
+                db.close()
+            }
+        }
+
+    @Test
+    fun store_tree_state_accepts_cached_bytes() =
+        runTest {
+            val db = VotingRustBackend.new().openVotingDb(newDbPath(), WALLET_ID)
+            try {
+                db.initRound(
+                    roundId = ROUND_ID,
+                    snapshotHeight = SNAPSHOT_HEIGHT,
+                    eaPK = EA_PK,
+                    ncRoot = NC_ROOT,
+                    nullifierIMTRoot = NULLIFIER_IMT_ROOT,
+                    sessionJson = null
+                )
+
+                db.storeTreeState(ROUND_ID, byteArrayOf(1, 2, 3))
             } finally {
                 db.close()
             }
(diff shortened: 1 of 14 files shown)


backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -20,6 +22,12 @@ const JNI_DELEGATION_PROOF_RESULT: &str =
 const JNI_DELEGATION_SUBMISSION_RESULT: &str =
     "cash/z/ecc/android/sdk/internal/model/voting/JniDelegationSubmissionResult";
 
+// Must match JniNoteInfo(ByteArray, ByteArray, Long, Long, ByteArray,
+// ByteArray, ByteArray, Int, String) in JniVotingModels.kt.
+const JNI_NOTE_INFO_CTOR_SIG: &str = "([B[BJJ[B[B[BILjava/lang/String;)V";
```
noop-sk, line 28: **Hardcoded JNI constructor signatures for new types**

`JNI_NOTE_INFO_CTOR_SIG` and `JNI_WITNESS_DATA_CTOR_SIG` are hardcoded strings that must exactly match the Kotlin data class constructors in `JniVotingModels.kt`. Any refactoring of `JniNoteInfo` or `JniWitnessData` — reordering fields, changing a type, adding a default — silently breaks at runtime with `NoSuchMethodError` rather than a compile-time error. The same pattern exists for earlier models, but given these two new types carry note commitments and auth paths, the failure surface is higher. Consider a reflective JVM test (or a note in the companion object) that validates the signature matches the actual constructor at test time.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
```diff
@@ -22,6 +24,19 @@ internal interface TypesafeVotingBackend {
...
+    suspend fun extractNcRoot(treeStateBytes: ByteArray): ByteArray
+
+    suspend fun verifyWitness(witness: JniWitnessData): Boolean
+
+    suspend fun getWalletNotes(
+        walletDbPath: String,
+        snapshotHeight: BlockHeight,
+        networkId: Int,
+        accountUuid: AccountUuid
+    ): List<JniNoteInfo>
```
noop-sk: **JNI type leaks through Typesafe interface return**

`getWalletNotes` takes domain-typed inputs (`BlockHeight`, `AccountUuid`) but returns `List<JniNoteInfo>` — a JNI-layer holder. The `Typesafe*` naming implies a layer that converts JNI primitives to SDK domain types; callers of this method (PR 7+) will receive `JniNoteInfo` objects and need to know their field semantics (e.g. `scope: Int` constants). If a `NoteInfo` domain type is planned for a later slice, noting that here would help reviewers track the intended end state; if `JniNoteInfo` is intentionally the final type, renaming the interface to drop "Typesafe" avoids the misleading implication.

backend-lib/src/main/rust/voting/notes.rs
```diff
@@ -18,6 +51,186 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_com
...
+
+        let path = java_string_to_rust(env, &wallet_db_path)?;
+        let network = network_from_id(network_id)?;
+        let height = BlockHeight::from_u32(jlong_to_u32(snapshot_height, "snapshot_height")?);
+        let account_uuid_bytes =
+            java_fixed_bytes::<ACCOUNT_UUID_BYTES>(env, &account_uuid_bytes, "accountUuidBytes")?;
+        let account_uuid =
+            zcash_client_sqlite::AccountUuid::from_uuid(uuid::Uuid::from_bytes(account_uuid_bytes));
+
+        let wallet_db = zcash_client_sqlite::WalletDb::for_path(
```
noop-sk: @nullcopy — `WalletDb::for_path` is called here without an explicit read-only flag (and a second time inside `generateNoteWitnessesNative`). If `for_path` opens in read-write mode it may acquire a write lock that contends with the wallet sync process running concurrently on the same file. Does `zcash_client_sqlite::WalletDb::for_path` open the database read-only when only `WalletRead` methods are called, or should this use a dedicated read-only constructor to avoid contention?

backend-lib/src/main/rust/voting/notes.rs
```diff
@@ -18,6 +51,186 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_com
...
+            .get_account(account_uuid)
+            .map_err(|e| anyhow!("get_account: {}", e))?
+            .ok_or_else(|| anyhow!("account not found in wallet DB"))?;
+        let ufvk = account
+            .ufvk()
+            .ok_or_else(|| anyhow!("account has no UFVK"))?
+            .clone();
+
+        let notes = wallet_db
+            .get_unspent_orchard_notes_at_historical_height(account_uuid, height)
```
noop-sk: @nullcopy — The name `get_unspent_orchard_notes_at_historical_height` is ambiguous: does "unspent" mean unspent *as of the snapshot height*, or unspent *as of now* (i.e. not spent in any block after the snapshot)? Voting correctness requires the former — a note spent after the snapshot height should still be included. If the implementation filters by current spend status, notes that were valid at snapshot time but have since been spent would be silently excluded from witness generation.

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -871,6 +996,130 @@ pub(super) fn require_round_phase_for_delegation_construction(
...
+        if !valid {
+            return Err(anyhow!(
+                "witness verification failed for position {}",
+                witness.position
+            ));
+        }
+    }
+
+    let tx = conn
+        .unchecked_transaction()
```
noop-sk: @nullcopy — `unchecked_transaction()` is used for the delete-then-reinsert of witnesses in `replace_bundle_witnesses`. If the voting DB connection is already in an error state or a savepoint context, `unchecked_transaction` will begin a nested transaction without checking ambient state, which could leave partial witness data on a commit failure. Is the `access_lock` held by callers sufficient to guarantee the connection is clean here, or should this be `transaction()` / a savepoint instead?

backend-lib/src/main/rust/voting/helpers.rs
greg0x: fcb1a14b

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
greg0x: 7a0fc40c

backend-lib/src/main/rust/voting/notes.rs
greg0x: ccc1874e

Read only is the way!

backend-lib/src/main/rust/voting/notes.rs
greg0x: bdc94bc1

backend-lib/src/main/rust/voting/helpers.rs
greg0x: Fixed in 5cc3c412 by replacing unchecked_transaction() with checked transaction() for witness replacement

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -871,6 +996,101 @@ pub(super) fn require_round_phase_for_delegation_construction(
...
+
+pub(super) fn replace_bundle_witnesses(
+    conn: &rusqlite::Connection,
+    round_id: &str,
+    wallet_id: &str,
+    bundle_index: u32,
+    witnesses: &[WitnessData],
+) -> anyhow::Result<()> {
+    conn.execute(
+        "DELETE FROM witnesses
```
greg0x: Note for reviewer: I'm moving this into `zcash_voting`

greg0x: cb040afc0475

backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
```diff
@@ -327,6 +374,49 @@ class VotingRustBackendTest {
...
+                assertFailsWith<RuntimeException> {
+                    db.storeTreeState(ROUND_ID, byteArrayOf(1, 2, 3))
+                }
+            } finally {
+                db.close()
+            }
+        }
+
+    @Test
+    fun store_tree_state_accepts_matching_snapshot_fixture() =
```
noop-sk, line 450: **Missing nc_root / height mismatch test**

`store_tree_state_rejects_invalid_cached_bytes` covers unparseable proto bytes; `store_tree_state_accepts_matching_snapshot_fixture` covers the happy path. `validate_cached_tree_state_for_round` has two additional rejection branches — height mismatch and nc_root mismatch — with no JNI-level coverage. A test that initialises the round with a mismatched `snapshotHeight` or `ncRoot` and calls `storeTreeState` with the fixture would cover both paths.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
```diff
@@ -316,6 +376,33 @@ class VotingRustBackend private constructor() {
...
+        @Throws(RuntimeException::class)
+        suspend fun storeTreeState(
+            roundId: String,
+            treeStateBytes: ByteArray
+        ) = withHandle { handle ->
+            storeTreeStateNative(handle, roundId, treeStateBytes)
+        }
+
+        @Throws(RuntimeException::class)
+        suspend fun generateNoteWitnesses(
```
noop-sk, line 395: **No JNI smoke test for `generateNoteWitnesses`**

`storeTreeState` gets two JNI-level instrumented tests; `generateNoteWitnesses` gets none. A minimal test asserting that passing a non-existent wallet DB path throws `RuntimeException` would catch native-link errors at CI time without requiring a real wallet DB.

backend-lib/src/main/rust/voting/notes.rs
```diff
@@ -18,6 +110,188 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_com
...
+
+        let height =
+            BlockHeight::from_u32(u32::try_from(params.snapshot_height).map_err(|_| {
+                anyhow!(
+                    "stored snapshot_height must be in range 0..=u32::MAX, got {}",
+                    params.snapshot_height
+                )
+            })?);
+
+        let wallet_db = open_wallet_db_read_only(&wallet_path, network)?;
```
noop-sk: @nullcopy — `generateNoteWitnessesNative` opens a read-only `WalletDb` connection while holding the voting DB `access_lock`. If the main SDK sync task is writing new blocks to the wallet DB concurrently (via a separate connection in WAL mode), can this read-only connection observe a partially-written snapshot, or does SQLite WAL provide a consistent read from the moment the connection is opened? We need confirmation that calling `generateNoteWitnesses` during active block sync cannot produce witnesses against an inconsistent note set.

backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
greg0x: 817c6cce

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
greg0x: db19cde0

backend-lib/src/main/rust/voting/notes.rs
greg0x: 97925a9 wraps witness generation in one wallet DB transaction, so the shard roots, shard rows, and cap are all read from the same committed SQLite snapshot during active sync. SQLite does not expose partial/uncommitted writes across separate connections.

noop-sk (dismissed):

backend-lib/src/main/rust/voting/notes.rs
```diff
@@ -18,6 +110,196 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_com
...
+            .map_err(|e| anyhow!("get_account: {}", e))?
+            .ok_or_else(|| anyhow!("account not found in wallet DB"))?;
+        let ufvk = account
+            .ufvk()
+            .ok_or_else(|| anyhow!("account has no UFVK"))?
+            .clone();
+
+        // Upstream interprets "unspent" at the requested height: spends mined
+        // after the snapshot remain eligible for that snapshot.
+        let notes = wallet_db
```
nullcopy: db read should execute transactionally (same as you do e.g. in `generateNoteWitnessesNative`)

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
```diff
@@ -37,12 +39,35 @@ internal class TypesafeVotingBackendImpl : TypesafeVotingBackend {
...
+    override suspend fun extractOrchardFvkFromUfvk(ufvk: String, networkId: Int): ByteArray =
+        rustBackend().extractOrchardFvkFromUfvk(ufvk, networkId)
+
+    override suspend fun extractNcRoot(treeStateBytes: ByteArray): ByteArray =
+        rustBackend().extractNcRoot(treeStateBytes)
+
+    override suspend fun verifyWitness(witness: JniWitnessData): Boolean =
+        rustBackend().verifyWitness(witness)
+
+    override suspend fun getWalletNotes(
```
nullcopy, line 61: Need tests for `getWalletNotes`

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
```diff
@@ -298,6 +333,24 @@ private class RustVotingDbBackend(
             keystoneSig,
             keystoneSighash
         )
+
+    override suspend fun storeTreeState(roundId: String, treeStateBytes: ByteArray) =
+        votingDb.storeTreeState(roundId, treeStateBytes)
+
+    override suspend fun generateNoteWitnesses(
```
nullcopy, line 425: `generate_note_witnesses_*` only tests the "missing wallet DB path" case. No coverage for: account-not-in DB, account-with-no-UFVK, snapshot-above-fully-scanned, empty-frontier ("no Orchard activity at snapshot"). These are all reachable error paths with anyhow! messages worth pinning

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
```diff
@@ -122,6 +136,68 @@ internal interface TypesafeVotingDb {
...
+            value == other.value &&
+            position == other.position &&
+            diversifier.contentEquals(other.diversifier) &&
+            rho.contentEquals(other.rho) &&
+            rseed.contentEquals(other.rseed) &&
+            scope == other.scope &&
+            ufvk == other.ufvk
+    }
+
+    override fun hashCode(): Int {
```
nullcopy, line 176: I thought this came in another PR... am I halucinating? Just double check this isn't duplicated effort 😅

nullcopy (changes requested):
Reviewed up to 97925a94bc5af9ad8de8c78437bd5e67e26354b7

backend-lib/src/main/rust/voting/notes.rs
greg0x: 280b1122

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
greg0x: ea313ddf

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
greg0x: 05e33562

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
greg0x: You’re remembering `JniNoteInfo` from #1950. This one is intentionally similar but sits at a different layer: `JniNoteInfo` remains the raw JNI DTO, while `VotingNoteInfo` is the SDK/typesafe-layer model.

This split was added in this PR after feedback that [`TypesafeVotingBackend.getWalletNotes()` should not expose `List<JniNoteInfo>` / raw `scope: Int`](https://github.com/zcash/zcash-android-wallet-sdk/pull/1951#discussion_r3229473126). `VotingNoteInfoMapper` is the conversion boundary between the two.

This is more verbose, but I wanted to err on the correctness side rather than dropping the typesafe.

nullcopy (dismissed):
utACK a2c049ee6552b94743ea29946da014208c3f6ab8

nullcopy (approved):
utACK https://github.com/zcash/zcash-android-wallet-sdk/commit/de35143ed8e7d32ad7f5c62f6d2526e28d985e32

PR #1952, opened 2026-05-12
[#1924] Add Android voting tree sync, commitment, and cast-vote payloads

greg0x:
Stacked on zcash/zcash-android-wallet-sdk#1951 (PR 6 — snapshot wallet notes and note witnesses).
Resolves MOB-1109

## Why

Seventh slice of the Android shielded-voting integration. PR 6 (#1951) lands
the snapshot input-prep path (wallet notes, tree state, note witnesses). This
PR builds the vote-construction path on top of that snapshot: syncing the vote
commitment tree, storing VAN positions, generating VAN witnesses, building the
vote commitment proof, decomposing weight, constructing share payloads, and
signing the cast-vote message.

`zcash_voting` is now built with `client-tree-sync` in addition to
`client-pir`. Together with the snapshot inputs from PR 6, this completes the
SDK-side path from snapshot-note witnesses to a fully signed cast-vote message
ready for submission.

No new app-facing `Synchronizer` voting API is added. All new surface lives on
the internal `TypesafeVotingBackend` and its JNI implementation.

## Not included

- Recovery records and share tracking (PR 8).
- Any new submission or proof-generation behavior outside vote construction.
- Note witness generation / wallet DB access (lands in PR 6).

## Review focus

- `client-tree-sync` feature enablement on `zcash_voting` in
  `backend-lib/Cargo.toml`, expected transitives, and lockfile churn.
- `voting/tree.rs`: vote tree sync, tree-client reset, VAN position storage,
  and VAN witness generation against a stored snapshot anchor — node URL /
  network behavior at the SDK boundary, anchor-height validation, and JSON
  schema stability for the returned VAN witness payload.
- `voting/vote.rs`: vote commitment proof inputs and byte-boundary validation,
  VAN witness consumption, weight decomposition, share payload serialization,
  and cast-vote signature domain/network binding.
- JNI byte-boundary validation via the shared helpers in `voting/helpers.rs`
  (bundle/proposal/choice ids, weight bytes, witness payload sizes, signing
  keys, sighash bytes).
- Typed FFI models (`JniWeightDecomposition`, `JniSharePayloadInput`,
  `JniCastVoteInput`, vote-commitment input/result, VAN witness payload) and
  how the typesafe wrapper enforces size/count invariants before handing data
  back to the SDK.
- `TypesafeVotingBackend` / `TypesafeVotingBackendImpl` ergonomics for the
  vote-construction workflow, including suspend/dispatcher choice and error
  mapping.
- Whether progress-callback plumbing should appear at this layer or stay
  deferred until a workflow actually drives long proof work through it.

## Validation

- `./gradlew checkProperties`
- `git diff --check`

# Author

# Reviewer

### DIFF
--- a/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
+++ b/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
@@ -971,6 +971,54 @@ class VotingRustBackendTest {
         }
 
     @Test
+    fun sync_vote_tree_reaches_native_boundary() =
+        runTest {
+            val db = VotingRustBackend.new().openVotingDb(newDbPath(), WALLET_ID)
+            try {
+                db.initPcztRoundWithBundles(notes(noteCount = 6, value = PCZT_NOTE_VALUE))
+
+                assertFailsWith<RuntimeException> {
+                    db.syncVoteTree(PCZT_ROUND_ID, "not-a-url")
+                }
+            } finally {
+                db.close()
+            }
+        }
+
+    @Test
+    fun store_van_position_reaches_native_boundary() =
+        runTest {
+            val db = VotingRustBackend.new().openVotingDb(newDbPath(), WALLET_ID)
+            try {
+                db.initPcztRoundWithBundles(notes(noteCount = 6, value = PCZT_NOTE_VALUE))
+
+                db.storeVanPosition(PCZT_ROUND_ID, bundleIndex = 1, position = 42)
+            } finally {
+                db.close()
+            }
+        }
+
+    @Test
+    fun generate_van_witness_reaches_native_boundary() =
+        runTest {
+            val db = VotingRustBackend.new().openVotingDb(newDbPath(), WALLET_ID)
+            try {
+                db.initPcztRoundWithBundles(notes(noteCount = 6, value = PCZT_NOTE_VALUE))
+                db.storeVanPosition(PCZT_ROUND_ID, bundleIndex = 1, position = 42)
+
+                assertFailsWith<RuntimeException> {
+                    db.generateVanWitness(
+                        roundId = PCZT_ROUND_ID,
+                        bundleIndex = 1,
+                        anchorHeight = SNAPSHOT_HEIGHT
+                    )
+                }
+            } finally {
+                db.close()
+            }
+        }
+
+    @Test
     fun build_vote_commitment_requires_delegation_ready() =
         runTest {
             val db = VotingRustBackend.new().openVotingDb(newDbPath(), WALLET_ID)
--- a/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
+++ b/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
@@ -924,6 +924,35 @@ class VotingRustBackendTest {
         }
 
     @Test
+    fun build_share_payloads_single_share_mode_keeps_commitment_context() =
+        runTest {
+            val backend = VotingRustBackend.new()
+            val commitment = jniVoteCommitmentResult()
+
+            val payloads =
+                backend.buildSharePayloads(
+                    commitment = commitment,
+                    voteDecision = 1,
+                    numOptions = 2,
+                    vcTreePosition = 42,
+                    singleShareMode = true
+                )
+
+            assertEquals(1, payloads.size)
+            val payload = payloads.single()
+            assertContentEquals(commitment.sharesHash, payload.sharesHash)
+            assertEquals(commitment.proposalId, payload.proposalId)
+            assertEquals(1, payload.voteDecision)
+            assertEquals(42, payload.treePosition)
+            assertEquals(0, payload.encShare.shareIndex)
+            assertEquals(commitment.encShares.first(), payload.encShare)
+            assertEquals(commitment.encShares, payload.allEncShares)
+            assertEquals(JNI_VOTE_SHARE_COUNT, payload.allEncShares.size)
+            assertContentEquals(commitment.shareComms.first(), payload.shareComms.first())
+            assertContentEquals(commitment.shareBlinds.first(), payload.primaryBlind)
+        }
+
+    @Test
     fun build_share_payloads_rejects_malformed_share_counts() =
         runTest {
             val backend = VotingRustBackend.new()
--- a/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
+++ b/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
@@ -4,6 +4,9 @@ import android.content.ContentValues
 import android.database.sqlite.SQLiteDatabase
 import cash.z.ecc.android.sdk.internal.model.voting.JniNoteInfo
 import cash.z.ecc.android.sdk.internal.model.voting.JniRoundPhase
+import cash.z.ecc.android.sdk.internal.model.voting.JniVanWitness
+import cash.z.ecc.android.sdk.internal.model.voting.JniVoteCommitmentResult
+import cash.z.ecc.android.sdk.internal.model.voting.JniWireEncryptedShare
 import cash.z.ecc.android.sdk.internal.model.voting.JniWitnessData
 import kotlinx.coroutines.test.runTest
 import org.junit.Test
@@ -124,6 +127,14 @@ class VotingRustBackendTest {
             assertFailsWith<RuntimeException> {
                 backend.computeShareNullifier(VOTE_COMMITMENT, OUT_OF_RANGE_SHARE_INDEX, BLIND)
             }
+        }
+
+    @Test
+    fun decompose_weight_returns_exact_shares() =
+        runTest {
+            val shares = VotingRustBackend.new().decomposeWeight(65_535).toList()
+
+            assertEquals((0 until JNI_VOTE_SHARE_COUNT).map { 1L shl it }, shares)
         }
 
     @Test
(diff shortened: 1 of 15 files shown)


sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
```diff
@@ -530,6 +653,57 @@ internal class TypesafeVotingDbImpl(
...
     }
+
+    override suspend fun syncVoteTree(roundId: String, nodeUrl: String): Long =
+        votingDb.syncVoteTree(roundId, nodeUrl)
+
+    override suspend fun resetTreeClient(roundId: String) =
+        votingDb.resetTreeClient(roundId)
+
+    override suspend fun resetAllTreeClients() =
+        votingDb.resetTreeClient("")
```
noop-sk: **Empty-string sentinel leaks through interface**

`resetAllTreeClients()` delegates to `votingDb.resetTreeClient("")`, relying on an empty string to mean "reset all." `VotingDbBackend.resetTreeClient(roundId: String)` has no documented behaviour for an empty string, so any future caller or test double that passes `""` directly will silently reset all tree clients. The "reset all" path should be a distinct method on `VotingDbBackend`, or the convention must be documented on the interface.

backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
```diff
@@ -710,6 +721,99 @@ class VotingRustBackendTest {
...
+                    voteDecision = 1,
+                    numOptions = 2,
+                    vcTreePosition = 42,
+                    singleShareMode = false
+                )
+            }
+        }
+
+    @Test
+    fun sign_cast_vote_rejects_unsupported_account_index() =
```
noop-sk, line 988: **Missing `signCastVote` success-path test**

`sign_cast_vote_rejects_unsupported_account_index` covers only the rejection case. There is no instrumented test that calls `signCastVote` with `accountIndex=0` and a valid commitment and asserts the returned `ByteArray` is `JNI_SPEND_AUTH_SIG_BYTES_SIZE` bytes. A signature-length assertion would catch JNI constructor drift silently.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
```diff
@@ -403,6 +447,77 @@ class VotingRustBackend private constructor() {
                 ) ?: error("generateNoteWitnesses returned null")
             }
 
+        @Throws(RuntimeException::class)
+        suspend fun syncVoteTree(roundId: String, nodeUrl: String): Long =
```
noop-sk, line 458: **No JNI smoke tests for tree operations**

`syncVoteTree`, `storeVanPosition`, and `generateVanWitness` have no instrumented JNI-level tests. `buildVoteCommitment` gets `build_vote_commitment_requires_delegation_ready`; the tree-sync path gets nothing. A minimal smoke test per function — calling the native and asserting the return type or a thrown exception — would catch native-link errors at CI time.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
```diff
@@ -80,6 +80,208 @@ data class JniWitnessData(
...
+    val vanNullifier: ByteArray,
+    val voteAuthorityNoteNew: ByteArray,
+    val voteCommitment: ByteArray,
+    val proposalId: Int,
+    val proof: ByteArray,
+    val encShares: List<JniWireEncryptedShare>,
+    val anchorHeight: Long,
+    val voteRoundId: String,
+    val sharesHash: ByteArray,
+    val shareBlinds: List<ByteArray>,
```
noop-sk, line 157: @nullcopy — `shareBlinds`, `shareComms`, `rVpk`, and `alphaV` land on the Kotlin GC heap inside `JniVoteCommitmentResult` and live there until `signCastVote` reads them back. The comment in `helpers.rs` says "plaintext/randomness fields intentionally never cross JNI" — but that appears to refer only to the unencrypted `enc_shares`. Are the blinding factors considered secret material that should be zeroized, or are they safe to hold unprotected on the Kotlin heap between the two JNI calls?

backend-lib/src/main/rust/voting/vote.rs
```diff
@@ -0,0 +1,186 @@
...
+        ))
+    }
+}
+
+fn require_delegation_ready_for_vote(
+    db: &VotingDbHandle,
+    round_id: &str,
+    bundle_index: u32,
+    witness: &voting::tree_sync::VanWitness,
+) -> anyhow::Result<()> {
```
noop-sk, line 164: @nullcopy — `require_delegation_ready_for_vote` does two DB reads (load delegation submission data, load VAN position) and a position equality check while `_access_lock` is held but without a DB transaction. If `VotingDb` serialises writes internally, the mutex guarantee is sufficient; if not, a concurrent write to `van_position` between the two reads could pass or fail unexpectedly. Is the Kotlin `accessMutex` / Rust `access_lock` sufficient to make this read-check sequence atomic, or does it need a DB transaction?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
greg0x: 9ac69253

backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
greg0x: 35f94855

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
greg0x: 6be590f9

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
greg0x: e52a4f2d

backend-lib/src/main/rust/voting/vote.rs
greg0x: 4a1080f7

`access_lock` is sufficient for SDK-exposed voting DB calls: handles for the same DB path/wallet share one VotingDbHandle, and every JNI voting DB operation takes that lock. This is not a DB-transaction guarantee against external SQLite writers, but that is outside the SDK boundary.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
```diff
@@ -80,6 +80,216 @@ data class JniWitnessData(
...
+/**
+ * Typed JNI carrier for vote commitment outputs.
+ *
+ * `shareBlinds` and `alphaV` are transient reveal/signing inputs. They should
+ * not be persisted or logged; they are carried here only because follow-up JNI
+ * calls consume the typed commitment result. Encrypted-share plaintext and
+ * encryption randomness remain Rust-only and are not included in [encShares].
+ */
+@Keep
+data class JniVoteCommitmentResult(
```
noop-sk, line 147: **`data class` `toString()` exposes signing material**

The docstring on lines 140–143 states that `shareBlinds`, `alphaV`, and `rVpk` must not be persisted or logged, but the compiler-generated `toString()` includes all 13 fields. Any `Log.d(...)`, coroutine exception, or `error("...")` call that captures this object will expose those values in logcat. Override `toString()` to redact those three fields.

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -29,6 +35,21 @@ const JNI_NOTE_INFO_CTOR_SIG: &str = "([B[BJJ[B[B[BILjava/lang/String;)V";
 // Must match JniWitnessData(ByteArray, Long, ByteArray, Array<ByteArray>)
 // in JniVotingModels.kt. Guarded by JniVotingModelsTest.
 const JNI_WITNESS_DATA_CTOR_SIG: &str = "([BJ[B[[B)V";
+// Must match JniVanWitness(Array<ByteArray>, Long, Long) in
+// JniVotingModels.kt.
+const JNI_VAN_WITNESS_CTOR_SIG: &str = "([[BJJ)V";
```
noop-sk, line 40: **New JNI constructor signatures not covered by `JniVotingModelsTest`**

The existing `JNI_NOTE_INFO_CTOR_SIG` and `JNI_WITNESS_DATA_CTOR_SIG` are marked `// Guarded by JniVotingModelsTest.` and have reflection tests that verify the signature string matches the actual constructor. The three new signatures here (`JNI_VAN_WITNESS_CTOR_SIG`, `JNI_VOTE_COMMITMENT_RESULT_CTOR_SIG`, `JNI_SHARE_PAYLOAD_CTOR_SIG`) have no equivalent tests — and `JniVotingModelsTest.jniDescriptor()` currently can't handle `JObject`-type parameters (e.g. `JniWireEncryptedShare`), so the helper would need extending. A parameter-order refactor to any of these data classes will produce a silent `NoSuchMethodError` at runtime.

backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
```diff
@@ -794,6 +805,163 @@ class VotingRustBackendTest {
             }
         }
 
+    @Test
+    fun build_share_payloads_round_trips_commitment_fields() =
```
noop-sk, line 901: **`singleShareMode=true` path not tested**

Both `buildSharePayloads` tests here use `singleShareMode = false`. `TypesafeVotingBackendImpl.buildSharePayloads` has a distinct `expectedCount = 1` branch for `singleShareMode = true`, and the Rust side presumably produces a single share payload in that mode. Add a test with `singleShareMode = true` asserting that exactly one payload is returned and that its fields are consistent with the commitment inputs.

backend-lib/src/main/rust/voting/db.rs
```diff
@@ -16,6 +16,7 @@ struct DbKey {
 
 pub(super) struct VotingDbHandle {
     db: VotingDb,
+    pub(super) tree_sync: VoteTreeSync,
```
noop-sk, line 23: @nullcopy — `VoteTreeSync` is a `pub(super)` field on `VotingDbHandle` next to `access_mutex`. Every JNI call in `tree.rs` acquires `access_lock()` before touching `tree_sync`, but `syncVoteTreeNative` passes `&db` back into `tree_sync.sync(&db, ...)` while `tree_sync` is already borrowed through `db`. Can you confirm that `VoteTreeSync` does not re-borrow `db.tree_sync` internally, and that the `access_mutex` alone is sufficient to serialize all mutations to `VoteTreeSync` state (no internal async state that escapes the lock)?

backend-lib/src/main/rust/voting/vote.rs
```diff
@@ -0,0 +1,190 @@
...
+    network_id: jint,
+    account_index: jint,
+    commitment: JObject<'local>,
+) -> jbyteArray {
+    let res = catch_unwind(&mut env, |env| {
+        require_supported_vote_account(account_index)?;
+        let hotkey_seed =
+            java_secret_bytes_at_least(env, &hotkey_seed, "hotkey_seed", PROTOCOL_FIELD_BYTES)?;
+        let commitment = java_vote_commitment_bundle(env, &commitment)?.bundle;
+        let sig = voting::vote_commitment::sign_cast_vote(
```
noop-sk, line 73: @nullcopy — `sign_cast_vote` is called with `vote_round_id`, `r_vpk_bytes`, `van_nullifier`, `vote_commitment`, `proposal_id`, `anchor_height`, and `alpha_v` sourced from `JniVoteCommitmentResult` (which originated from `build_vote_commitment` on the same round). Can you confirm that these are the exact inputs required by the circuit's sighash domain, and that the field ordering and any domain separator match what the submission verifier expects?

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
greg0x: c191351b

backend-lib/src/main/rust/voting/helpers.rs
greg0x: f4e6ef39

backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
greg0x: 47279dd1

backend-lib/src/main/rust/voting/db.rs
greg0x: 4403cde8

`VoteTreeSync` has only `clients: Mutex<HashMap<String, TreeClient>>` plus `transport: Arc<HyperTransport>`. `sync()` reads from `VotingDb`, then locks `clients` and calls `client.sync(&api)` synchronously. It does not re-borrow `db.tree_sync`, store `db`, or spawn state that escapes the lock.

backend-lib/src/main/rust/voting/vote.rs
greg0x: c3d8285e

Confirmed. The SDK forwards the typed fields from `JniVoteCommitmentResult`, produced by `build_vote_commitment`, into [`zcash_voting::vote_commitment::sign_cast_vote`](https://github.com/valargroup/zcash_voting/blob/zcash_voting-v0.5.11/zcash_voting/src/vote_commitment.rs#L65-L141). The canonical sighash domain is `SVOTE_CAST_VOTE_SIGHASH_V0`, and the field order is `vote_round_id`, `r_vpk`, `van_nullifier`, `vote_authority_note_new`, `vote_commitment`, `proposal_id` as 4-byte LE padded to 32 bytes, and `anchor_height` as 8-byte LE padded to 32 bytes. That matches the submission verifier in [`vote-sdk` `types.ComputeCastVoteSighash`](https://github.com/valargroup/vote-sdk/blob/745aa4f1d6fc5dad86ec7a479e7a9114b313620b/x/vote/types/sighash.go#L22-L64). `alpha_v` is not hashed; it is used to derive the randomized signing key corresponding to `r_vpk`.

backend-lib/src/main/rust/voting/vote.rs
```diff
@@ -0,0 +1,193 @@
...
+        require_supported_vote_account(account_index)?;
+        let hotkey_seed =
+            java_secret_bytes_at_least(env, &hotkey_seed, "hotkey_seed", PROTOCOL_FIELD_BYTES)?;
+        let commitment = java_vote_commitment_bundle(env, &commitment)?.bundle;
+        // zcash_voting owns the canonical cast-vote sighash domain and field
+        // order. The JNI layer keeps this typed and passes the bundle fields
+        // through instead of reconstructing the sighash locally.
+        let sig = voting::vote_commitment::sign_cast_vote(
+            hotkey_seed.expose_secret(),
+            u32::try_from(network_id)
```
nullcopy, line 75: Almost everywhere else uses jint_to_u32 to for u32 type casting. I'm not sure which is more idiomatic, but I'd prefer to be consistent

backend-lib/src/main/rust/voting/vote.rs
```diff
@@ -0,0 +1,193 @@
...
+            jint_to_u32(bundle_index, "bundle_index")?,
+            &witness,
+        )?;
+        let reporter = progress_reporter_from_callback(env, &progress_callback)?;
+        let bundle = db
+            .build_vote_commitment(
+                &round_id,
+                jint_to_u32(bundle_index, "bundle_index")?,
+                hotkey_seed.expose_secret(),
+                u32::try_from(network_id)
```
nullcopy, line 131: Same comment about using jint_to_u32

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -662,6 +856,226 @@ fn make_jni_witness_data<'local>(
     })
 }
 
+pub(super) fn make_jni_van_witness<'local>(
```
nullcopy, line 859: (non-blocking) All these type conversion helpers look like they could be instead implemented as `TryFrom`. Is there a reason not to?

nullcopy (commented):
flushing comments

nullcopy (approved):
None of my comments are blocking

utACK 07e069e7b2ea0982c1f882efac39defd3ad7c4c1

backend-lib/src/main/rust/voting/vote.rs
greg0x: 10ab1f7a9856

backend-lib/src/main/rust/voting/vote.rs
greg0x: 10ab1f7a9856

backend-lib/src/main/rust/voting/helpers.rs
greg0x: 10ab1f7a9856

PR #1953, opened 2026-05-13
[#1924] Add Android voting recovery and share tracking

greg0x:
Stacked on zcash/zcash-android-wallet-sdk#1952 (PR 7 — vote tree sync, commitment, and cast-vote payloads).
Resolves MOB-1109

## Why

Eighth slice of the Android shielded-voting integration. PR 7 (#1952) lands the
SDK-side path from snapshot-note witnesses to a fully signed cast-vote message.
This PR adds the recovery and share-tracking state that lets a wallet replay,
retry, and confirm those workflows after restart or partial failure.

Native side: a new `voting/recovery.rs` module exposes 12 JNI entry points
covering delegation/vote tx-hash store+lookup with typed missing-state, a
`markVoteSubmitted` flag, commitment-bundle persistence with hex-encoded JSON
storage, `clearRecoveryState`, share-delegation recording/listing, an
unconfirmed-delegations query, share confirmation, and additive
`addSentServers`. Commitment bundles round-trip through a shared
`JniVoteCommitmentResultPayload` builder so freshly built and replayed bundles
go through the same byte-boundary validation.

SDK side: 13 new suspend functions on `TypesafeVotingDb` + matching
`VotingDbBackend` / `RustVotingDbBackend` / `TypesafeVotingDbImpl` wiring, with
typed `VotingTxHashLookup` (`Missing` / `Found`), `CommitmentBundleRecord`,
and `ShareDelegationRecord` models so the SDK boundary distinguishes missing
state from native errors.

`zcash_voting` is bumped to 0.5.12 for the recovery DB methods this PR calls
and the matching governance/delegation hotkey-seed parameterization. No new
app-facing `Synchronizer` voting API is added.

## Validation

- `./gradlew checkProperties`
- `git diff --check`

# Author

- [x] Check the **code coverage** report for the automated tests
- [x] Pull in the latest changes from the **main** branch and **squash** your commits before assigning a reviewer

# Reviewer

### DIFF
--- a/backend-lib/Cargo.lock
+++ b/backend-lib/Cargo.lock
@@ -1230,7 +1230,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1480,7 +1480,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2389,7 +2389,7 @@ version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7a1886916523694cd6ea3d175f03a1e5010699a2a4cc13696d83d7bea1d80638"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2679,7 +2679,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3791,7 +3791,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -4320,7 +4320,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3a766e1110788c36f4fa1c2b71b387a7815aa65f88ce0229841826633d93723e"
 dependencies = [
  "libc",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -4495,7 +4495,7 @@ dependencies = [
  "getrandom 0.4.2",
  "once_cell",
  "rustix",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -6293,7 +6293,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -6813,6 +6813,8 @@ dependencies = [
  "rust_decimal",
  "sapling-crypto",
  "secrecy",
+ "serde",
+ "serde_json",
  "tonic",
  "tor-rtcompat",
  "tracing",
@@ -7135,9 +7137,9 @@ dependencies = [
 
 [[package]]
 name = "zcash_voting"
-version = "0.5.11"
+version = "0.5.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d7835b7e3d823bfc889279f9c7ab3d4033d3d707004dbf3a0fa158f6f1ef8b43"
+checksum = "066eefe9678138a8450c20ec222da402bb92bb3f46351958a9661a75733c0bc8"
 dependencies = [
  "anyhow",
  "blake2b_simd",
--- a/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
+++ b/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
@@ -1104,6 +1104,60 @@ class VotingRustBackendTest {
         }
 
     @Test
+    fun add_sent_servers_appends_and_deduplicates_native_jni() =
+        runTest {
+            val db = VotingRustBackend.new().openVotingDb(newDbPath(), WALLET_ID)
+            try {
+                db.initPcztRoundWithBundles(notes(noteCount = 6, value = PCZT_NOTE_VALUE))
+                db.storeVoteFixtureForTesting(
+                    roundId = PCZT_ROUND_ID,
+                    bundleIndex = 1,
+                    proposalId = 1,
+                    choice = 0
+                )
+                db.recordShareDelegation(
+                    roundId = PCZT_ROUND_ID,
+                    bundleIndex = 1,
+                    proposalId = 1,
+                    shareIndex = SHARE_INDEX,
+                    sentToUrls = listOf("https://helper-1.example"),
+                    nullifier = ByteArray(FIELD_BYTES) { 0x55 },
+                    submitAt = 123
+                )
+
+                db.addSentServers(
+                    roundId = PCZT_ROUND_ID,
+                    bundleIndex = 1,
+                    proposalId = 1,
+                    shareIndex = SHARE_INDEX,
+                    newUrls = listOf("https://helper-2.example")
+                )
+                assertEquals(
+                    listOf("https://helper-1.example", "https://helper-2.example"),
+                    db.getShareDelegations(PCZT_ROUND_ID).single().sentToUrls
+                )
+
+                db.addSentServers(
+                    roundId = PCZT_ROUND_ID,
+                    bundleIndex = 1,
+                    proposalId = 1,
+                    shareIndex = SHARE_INDEX,
+                    newUrls = listOf("https://helper-2.example", "https://helper-3.example")
+                )
+                assertEquals(
+                    listOf(
+                        "https://helper-1.example",
+                        "https://helper-2.example",
+                        "https://helper-3.example"
+                    ),
+                    db.getShareDelegations(PCZT_ROUND_ID).single().sentToUrls
+                )
+            } finally {
+                db.close()
+            }
+        }
+
+    @Test
     fun delegation_proof_result_fixture_crosses_rust_to_kotlin_object_construction() =
         runTest {
             val result =
--- a/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
+++ b/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
@@ -1583,12 +1583,14 @@ class VotingRustBackendTest {
         encShares: List<JniWireEncryptedShare> = wireShares(),
         shareBlinds: List<ByteArray> = fieldElements(JNI_VOTE_SHARE_COUNT, 5),
         shareComms: List<ByteArray> = fieldElements(JNI_VOTE_SHARE_COUNT, 6),
+        bundleIndex: Int = 1,
         alphaV: ByteArray = ByteArray(FIELD_BYTES) { 8 }
     ) = JniVoteCommitmentResult(
         vanNullifier = ByteArray(FIELD_BYTES) { 1 },
         voteAuthorityNoteNew = ByteArray(FIELD_BYTES) { 2 },
         voteCommitment = ByteArray(FIELD_BYTES) { 3 },
         proposalId = 1,
+        bundleIndex = bundleIndex,
         proof = byteArrayOf(4),
         encShares = encShares,
         anchorHeight = SNAPSHOT_HEIGHT,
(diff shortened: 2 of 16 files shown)


backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
```diff
@@ -869,9 +869,7 @@ class VotingRustBackendTest {
                         pirServerUrl = "http://127.0.0.1:1",
                         networkId = TESTNET_NETWORK_ID,
                         notes = notes,
-                        walletSeed = SHORT_FIELD,
-                        accountIndex = ACCOUNT_INDEX,
-                        addressIndex = ADDRESS_INDEX,
```
p0mvn, line 874: Note to reviewer: these have been proven to be superfluos and removed

backend-lib/src/main/rust/voting/recovery.rs
```diff
@@ -0,0 +1,411 @@
...
+) -> anyhow::Result<Option<T>>
+where
+    E: std::fmt::Display,
+{
+    match result {
+        Ok(value) => Ok(value),
+        Err(error) if is_query_returned_no_rows(&error) => Ok(None),
+        Err(error) => Err(anyhow!("{label}: {error}")),
+    }
+}
```
p0mvn, line 150: Note to reviewer: this was returned to make empty results propagate None so that recovery paths can notice the absence of a round state and appropriately action on, as opposed to handling an error.

backend-lib/Cargo.lock
```diff
@@ -1230,7 +1230,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
```
p0mvn, line 1233: Note to reviewer: diff stems from dependency graph regeneration due to `zcash_voting` patch

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
```diff
@@ -525,6 +523,176 @@ class VotingRustBackend private constructor() {
...
+            shareIndex: Int
+        ) = withHandle { handle ->
+            check(markShareConfirmedNative(handle, roundId, bundleIndex, proposalId, shareIndex)) {
+                "markShareConfirmed failed for roundId=$roundId " +
+                    "bundleIndex=$bundleIndex proposalId=$proposalId shareIndex=$shareIndex"
+            }
+        }
+
+        @Throws(RuntimeException::class)
+        suspend fun addSentServers(
```
noop-sk, line 677: **`addSentServers` replace-or-append unclear**

The VotingRustBackendTest calls this with the full expanded list (original URL + new URL) and the assertion checks for exactly both entries, implying "set the new complete list". But the name `addSentServers` reads as "append the new entries". A wallet caller who passes only the newly-discovered URL expecting a merge will instead overwrite the stored list. Should the parameter be named `updatedUrls` and the doc clarify full-replace semantics, or should the function take a single new URL to append?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
```diff
@@ -954,6 +1233,25 @@ private fun JniSharePayload.requireValid() {
 private fun ((Double) -> Unit).asVotingProgressCallback() =
     VotingProofProgressCallback { progress -> invoke(progress) }
 
+@Suppress("TooGenericExceptionCaught")
+private suspend fun <T> runExpectedMissingRowLookup(block: suspend () -> T): T? =
```
noop-sk: **Exception backstop tests unreachable path**

`runExpectedMissingRowLookup` catches `RuntimeException` whose cause chain contains "Query returned no rows". In practice the Rust getters (`getDelegationTxHashNative`, `getVoteTxHashNative`, `getCommitmentBundleNative`) already absorb no-rows via `optional_recovery_lookup` and return null instead of throwing. So this Kotlin catch block cannot fire with the current Rust implementation. The test `missing_recovery_rows_from_native_exception_map_to_expected_state` therefore tests a dead path. If this is intentional defence-in-depth, document it; if Rust is the guaranteed conversion point, remove the Kotlin backstop to avoid misleading coverage.

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -1076,20 +1346,86 @@ fn make_jni_share_payload<'local>(
...
 ) -> anyhow::Result<jobject> {
     let class = env.find_class(JNI_VOTING_HOTKEY)?;
-    let secret_key = SecretVec::new(hotkey.secret_key);
-    let secret_key_len = secret_key.expose_secret().len();
-    if secret_key_len != PROTOCOL_FIELD_BYTES {
-        return Err(anyhow!(
-            "hotkey_secret_key must be exactly {PROTOCOL_FIELD_BYTES} bytes, got {secret_key_len}"
-        ));
-    }
+    require_len(
```
noop-sk: @nullcopy — `hotkey.secret_key` was previously wrapped in `SecretVec::new(...)` before the length check, so the key bytes were zeroized on drop. Now they are moved into `require_len`, which returns a plain `Vec<u8>` that is silently dropped with no zeroization. Is this regression intentional given the secret key bytes are about to be discarded here anyway, or should `require_len` be replaced with a `SecretVec`-aware length check to preserve the zeroization guarantee?

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -103,6 +120,48 @@ struct JniVoteRecordPayload {
...
+struct StoredWireEncryptedShare {
+    c1: String,
+    c2: String,
+    share_index: u32,
+}
+
+// Recovery replay needs the reveal and signing inputs below. Keep this JSON out
+// of logs; it contains the same sensitive fields as JniVoteCommitmentResult.
+#[derive(Deserialize, Serialize)]
+pub(super) struct StoredVoteCommitmentBundle {
```
noop-sk, line 151: @nullcopy — `StoredVoteCommitmentBundle` serializes `share_blinds`, `r_vpk_bytes`, and `alpha_v` (the reveal and signing inputs) as hex-encoded JSON to the voting DB. Two questions for you: (1) Is the voting DB encrypted at rest on Android, or do these fields land in plaintext SQLite on the device? (2) Does `clear_recovery_state` zero or overwrite those rows before deletion, or does SQLite just mark the pages as free (leaving the bytes recoverable from the raw file on a rooted device)?

backend-lib/src/main/rust/voting/delegation.rs
```diff
@@ -19,35 +19,32 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_bui
...
             ));
         }
-        let hotkey_raw_address = hotkey_orchard_raw_address_from_wallet_seed(
-            seed_bytes.expose_secret(),
-            network,
-            account_index,
-            address_index,
-        )?;
+        let hotkey_raw_address =
+            hotkey_orchard_raw_address(hotkey_seed.expose_secret(), network, 0)?;
```
noop-sk, line 47: @nullcopy — `hotkey_orchard_raw_address` now always derives from `account_index = 0` and `HOTKEY_ADDRESS_INDEX = 0`. The old API accepted both indices as parameters (and the removed test verified that different address indices produce different addresses). For a recovery replay to succeed, the address reconstructed during recovery must match the one embedded in the original ZKP. Can you confirm that `zcash_voting 0.5.12` uses address index 0 at account 0 for the hotkey address in both the prove and verify paths, so no existing votes will fail to replay?

backend-lib/src/main/rust/voting/helpers.rs
greg0x: 163adce7509e

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
greg0x: 01670722  - Clarifying comments

5fb51b055bc3 - Added an extra test to showcase the add behavior.

Confirmed this is append/dedupe semantics.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
greg0x: 2ac348ef

Agreed. Rust JNI is the guaranteed missing-row conversion point, so I removed the Kotlin backstop and the dead synthetic test.

backend-lib/src/main/rust/voting/helpers.rs
greg0x: The voting DB is currently app-private plaintext local voting state. It stores proposal/choice metadata and recovery progress needed to finish the voting flow. 

The commitment bundle is recovery/replay material for resubmitting shares for an already-committed vote; it is not wallet spend authority, does not contain share plaintext or decryption keys, and does not allow changing the committed vote. If the local DB is compromised, vote metadata confidentiality is already lost, and the practical impact of these fields is mainly that share submission can be replayed for that same round. 

`clear_recovery_state` is a logical cleanup, and doesn't overwrite the bytes.

backend-lib/src/main/rust/voting/delegation.rs
greg0x: Android fixes the hotkey address index to 0 and passes that into delegation state. zcash_voting then reloads the persisted address_index for ZKP #2, derives the vote/signing Orchard spending key through the account-0 helper, and passes that same address_index into voting-circuits. voting-circuits does not hardcode another index; it uses the supplied address_index directly with `fvk.address_at(address_index, Scope::External)`. 


  - Android HOTKEY_ADDRESS_INDEX = 0: https://github.com/valargroup/zcash-android-wallet-sdk/blob/5fb51b055bc335a4162a8c5dde0cb691463b201a/backend-lib/src/main/rust/voting/helpers.rs#L93-L94
  - Android passes account 0 and HOTKEY_ADDRESS_INDEX:
    https://github.com/valargroup/zcash-android-wallet-sdk/blob/5fb51b055bc335a4162a8c5dde0cb691463b201a/backend-lib/src/main/rust/voting/delegation.rs#L46-L68
  - zcash_voting account-0 spending key helper: https://github.com/valargroup/zcash_voting/blob/zcash_voting-v0.5.12/zcash_voting/src/zkp2.rs#L222-L229
  - zcash_voting reloads/passes persisted address_index: https://github.com/valargroup/zcash_voting/blob/zcash_voting-v0.5.12/zcash_voting/src/storage/operations.rs#L825-L865
  - voting-circuits uses supplied address_index: https://github.com/valargroup/voting-circuits/blob/v0.4.2/voting-circuits/src/vote_proof/builder.rs#L440-L463

backend-lib/src/main/rust/voting/helpers.rs
noop-sk: Approved

noop-sk (dismissed):

backend-lib/src/main/rust/voting/recovery.rs
```diff
@@ -0,0 +1,598 @@
...
+) -> jboolean {
+    let res = catch_unwind(&mut env, |env| {
+        let db = db_from_handle(db_handle)?;
+        let _access_lock = db.access_lock()?;
+        let round_id = java_string_to_rust(env, &round_id)?;
+        let bundle_index = jint_to_u32(bundle_index, "bundle_index")?;
+        let tx_hash = java_string_to_rust(env, &tx_hash)?;
+        db.store_delegation_tx_hash(&round_id, bundle_index, &tx_hash)
+            .map_err(|e| anyhow!("store_delegation_tx_hash: {e}"))?;
+        let stored = optional_recovery_lookup(
```
nullcopy: Please do not readback. This is defending against a failure in the db API, which should really be tested in the crate

nullcopy: This pattern is repeated several times throughout the PR. Please remove all readback checks.

backend-lib/src/main/rust/voting/recovery.rs
```diff
@@ -0,0 +1,598 @@
...
+            "commitment voteRoundId {} does not match roundId {round_id}",
+            commitment.bundle.vote_round_id
+        ));
+    }
+    if commitment.bundle.proposal_id != proposal_id {
+        return Err(anyhow!(
+            "commitment proposalId {} does not match proposalId {proposal_id}",
+            commitment.bundle.proposal_id
+        ));
+    }
```
nullcopy, line 214: Need to also check `bundle_index`, right?

backend-lib/src/main/rust/voting/helpers.rs
```diff
@@ -1076,27 +1363,96 @@ fn make_jni_share_payload<'local>(
     )?)
 }
 
+pub(super) fn make_jni_share_delegation_record_array<'local>(
```
nullcopy, line 1392: This and `make_jni_share_payload_array` (preexisting) don't have the same JNI local-ref cleanup as the rest of the interface. e.g. `make_jni_note_info_array` and `make_jni_witness_data_array` call `env.delete_local_ref(...)?` per iteration, but these do not

nullcopy (changes requested):
Reviewed up to 5fb51b055bc335a4162a8c5dde0cb691463b201a

backend-lib/src/main/rust/voting/recovery.rs
greg0x: 8d308b83

backend-lib/src/main/rust/voting/recovery.rs
greg0x: 98894052

backend-lib/src/main/rust/voting/helpers.rs
greg0x: af2e1520

nullcopy (approved):
utACK 0e5fca74593c3fe0a7009d06d27d94dd6bc8c69c

PR #1954, opened 2026-05-14
[#1924] Split governance PCZT API for Keystone vs software wallets

p0mvn:
## Summary

Splits the governance PCZT JNI into two explicit paths so Keystone (and any other hardware-wallet integration that cannot expose the wallet seed) can drive a delegation round, while software wallets keep their UFVK/walletSeed validation invariant.

## Why this shape

The previous `buildGovernancePczt` JNI required `walletSeed` and validated, inside Rust, that the seed derived the account UFVK before constructing the PCZT. That contract is correct for software wallets but is unsatisfiable on Keystone, which never surfaces the wallet seed. The Android Keystone integration was working around it by passing the hotkey seed as the wallet seed, which then failed with `ufvk does not match walletSeed for network_id=1, account_index=0` on a real round, immediately after also tripping `bundle_index 1 is not present in note bundle set` because the call site was pre-filtering the note set to a single bundle (the Rust side re-chunks internally from the full snapshot using `bundle_index`).

We considered three alternatives before landing this shape:

1. Loosening the seed/UFVK check to a runtime opt-in. Rejected: the check is a real safety invariant for software wallets; making it conditional inside Rust creates a quiet way to skip it.
2. Keeping a single API and switching on a `walletKind` flag. Rejected: it forces the FFI to know about wallet flavors, hides which inputs are actually consumed in each branch, and still requires an explicit hotkey-address path for hardware wallets.
3. Two narrowly typed APIs that share the construction core. Chosen: each entry point only takes inputs it can validate, the seed/UFVK invariant stays inside the seed path where it is meaningful, and the bundle/phase logic is shared so the on-disk state machine cannot drift between the two.

This also matches the iOS SDK contract (`zcash-swift-wallet-sdk`), which already exposes the explicit-FVK path for hardware-wallet flows.

## What changed

- `buildGovernancePczt` now takes pre-derived `fvkBytes` (Orchard FVK) and `hotkeyRawAddress`. This is the path Keystone uses; the host derives the FVK from the UFVK and the raw hotkey address from the hotkey seed before calling.
- `buildGovernancePcztFromSeed` is the new software-wallet entry point. It takes `ufvk`, `walletSeed`, `hotkeySeed` and keeps the original UFVK<>walletSeed validation, then derives the hotkey raw address internally.
- Both paths share `build_governance_pczt_for_bundle`, which validates the bundle index, enforces the round phase, persists the constructed delegation state, and only advances the phase to `DelegationConstructed` on success.
- `buildAndProveDelegation` now takes `hotkeyRawAddress` directly. Keystone does not need to surface the hotkey seed beyond derivation, and software wallets derive once via the new helper.
- New `deriveHotkeyRawAddress` JNI exposes raw-address derivation to the app layer so the Keystone path can stay seed-free end-to-end.
- Mirror updates to `TypesafeVotingBackend(Impl)` and the JNI/SDK androidTest harnesses.

## Manual testing

This was manually tested end-to-end on a Keystone hardware wallet: entering a poll no longer raises `bundle_index 1 is not present in note bundle set`, and constructing the delegation no longer raises `ufvk does not match walletSeed`. The software-wallet path was exercised through the existing app flow as well.

# Author

- [x] Check the **code coverage** report for the automated tests
- [x] Pull in the latest changes from the **main** branch and **squash** your commits before assigning a reviewer

# Reviewer

### DIFF
--- a/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
+++ b/backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
@@ -779,20 +779,43 @@ class VotingRustBackendTest {
     @Test
     fun build_governance_pczt_from_seed_uses_wallet_account_but_hotkey_account_zero() =
         runTest {
+            val backend = VotingRustBackend.new()
             val explicitDb = VotingRustBackend.new().openVotingDb(newDbPath(), WALLET_ID)
             val seedDb = VotingRustBackend.new().openVotingDb(newDbPath(), WALLET_ID)
             try {
                 val accountIndex = 1
                 val notes = notes(noteCount = 6, value = PCZT_NOTE_VALUE)
                 val ufvk = deriveTestUfvk(accountIndex = accountIndex)
+                val hotkeyAccountZero =
+                    backend.deriveHotkeyRawAddressForAccountFixture(
+                        HOTKEY_SEED,
+                        TESTNET_NETWORK_ID,
+                        ACCOUNT_INDEX
+                    )
+                val hotkeyWalletAccount =
+                    backend.deriveHotkeyRawAddressForAccountFixture(
+                        HOTKEY_SEED,
+                        TESTNET_NETWORK_ID,
+                        accountIndex
+                    )
                 explicitDb.initPcztRoundWithBundles(notes)
                 seedDb.initPcztRoundWithBundles(notes)
+
+                assertContentEquals(
+                    hotkeyAccountZero,
+                    backend.deriveHotkeyRawAddress(HOTKEY_SEED, TESTNET_NETWORK_ID)
+                )
+                assertFalse(hotkeyAccountZero.contentEquals(hotkeyWalletAccount))
 
                 val explicitPczt =
                     explicitDb.buildTestGovernancePczt(
                         ufvk = ufvk,
                         notes = notes,
-                        options = GovernancePcztOptions(accountIndex = accountIndex)
+                        options =
+                            GovernancePcztOptions(
+                                hotkeyRawAddress = hotkeyAccountZero,
+                                accountIndex = accountIndex
+                            )
                     )
                 val seedPczt =
                     seedDb.buildTestGovernancePcztFromSeed(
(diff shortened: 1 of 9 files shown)


backend-lib/src/main/rust/voting/util.rs
```diff
@@ -68,6 +68,27 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_ext
...
+    hotkey_seed: JByteArray<'local>,
+    network_id: jint,
+    account_index: jint,
+) -> jbyteArray {
+    let res = catch_unwind(&mut env, |env| {
+        let network = network_from_id(network_id)?;
+        let account_index = jint_to_u32(account_index, "account_index")?;
+        let hotkey_seed =
+            java_secret_bytes_at_least(env, &hotkey_seed, "hotkeySeed", PROTOCOL_FIELD_BYTES)?;
+        let bytes = hotkey_orchard_raw_address(hotkey_seed.expose_secret(), network, account_index)?;
```
czarcas7ic: I think we need to keep this account index aligned with the vote path. This derives the hotkey raw address using the caller’s `account_index`, but vote construction/signing still only supports account `0` and derives its hotkey material from account `0`.

If a caller passes a nonzero account here, delegation can be built for a hotkey account that the later vote path cannot use. Can we either guard this to account `0` for now, or thread the account index through the vote APIs and add a nonzero account test?

p0mvn: Addressed here by removing `account_index` from the API
[3ca753c](https://github.com/zcash/zcash-android-wallet-sdk/pull/1954/commits/3ca753ce7364daaa3ee30d19cac905a18368bf6a)

greg0x (dismissed):

backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
```diff
@@ -733,6 +753,91 @@ class VotingRustBackendTest {
...
+                assertValidGovernancePczt(backend, explicitDb, explicitPczt)
+                assertValidGovernancePczt(backend, seedDb, seedPczt)
+            } finally {
+                explicitDb.close()
+                seedDb.close()
+            }
+        }
+
+    @Test
+    fun build_governance_pczt_from_seed_uses_wallet_account_but_hotkey_account_zero() =
```
nullcopy, line 780: The test exercises accountIndex = 1 for the wallet but does not actually assert anything about the hotkey account.

nullcopy (changes requested):

backend-lib/src/androidTest/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackendTest.kt
greg0x: 79589efab6a3

nullcopy (approved):
utACK 79589efab6a305401f06090fa5abbd834c0ccbe1

PR #1961, opened 2026-05-19
Bump zcash_voting to 0.10.1

czarcas7ic:
Updates the Android backend Rust dependency to zcash_voting 0.10.1 and refreshes the locked voting crate set to the published releases. The nearby Cargo comments and changelog now point at the matching zcash_voting and voting-circuits tags.

Validation:
- cargo metadata --manifest-path backend-lib/Cargo.toml --locked --format-version 1
- cargo tree --manifest-path backend-lib/Cargo.toml --locked -i zcash_voting
- cargo check --manifest-path backend-lib/Cargo.toml --locked
- cargo test --manifest-path backend-lib/Cargo.toml --locked voting::

Note: full backend cargo test currently fails in utils::exception::tests::unknown_any, while all voting tests pass.

PR #1966, opened 2026-05-26
chore: Unpin zcash_voting from Cargo.toml in favor of lockfile

nullcopy:
This PR unpins zcash_voting in `Cargo.toml`, in favor of lockfile pinning. Pin was mistakenly added in https://github.com/zcash/zcash-android-wallet-sdk/pull/1961


nuttycom (approved):
utACK

PR #1969, opened 2026-05-28
Add shielded voting public API design

greg0x:
## Summary

This PR has two pieces:

- `docs/public-api-design-principles.md`: the main artifact. This lays out the public API design principles we should discuss and get right first.
- `docs/voting-public-api-design.md`: a shielded voting API design aligned with the current state of those principles. It is rooted in the current implementation, but proposes a cleaner public SDK boundary.

## Proposed Review Sequence

I propose we start by discussing the principles doc and getting that right. Once we are happy with the principles, we can update the voting API design doc based on them.

That gives us the sequence for the refactor:

1. Agree on the public API design principles.
2. Update the voting API design to match those principles.
3. Use the updated voting API design to drive the refactor order.

## Testing

- Not run; documentation-only design change.

# Author

# Reviewer

[^2]: _While we aim for automated testing of the SDK, some aspects require manual testing. If you had to manually test something during development of this pull request, write those steps down._
[^6]: _While the CI server runs the demo app to look for build failures or crashes, humans running the demo app are more likely to notice unexpected log messages, UI inconsistencies, or bad output data. Perform this step last, after verifying the code changes are safe to run locally._

### DIFF
--- a/docs/public-api-design-principles.md
+++ b/docs/public-api-design-principles.md
@@ -127,9 +127,24 @@ unrelated internals. For example, `Synchronizer` may expose a capability for a
 wallet-scoped feature, while that capability may expose account-scoped objects
 for operations that depend on `AccountUuid`.
 
+Capabilities are also security boundaries. IO, storage, network transport,
+hardware signing, and secret access should be represented by explicit capability
+interfaces instead of hidden global dependencies or ad hoc callbacks. A reviewer
+should be able to inspect the public capability and the SDK-internal function or
+constructor signatures that implement it, and understand what authority the code
+has: what it can read, what it can write, what it can send over the network, and
+what secrets or signing authority it can request.
+
 Good capability APIs:
 
 - make the starting point obvious;
+- make authority visible in function and constructor signatures;
+- keep IO behind explicit interfaces owned by the public capability or passed to
+  the SDK-internal logic that needs them;
+- keep network transport, general storage, signing, and secret access as separate
+  authorities;
+- avoid giving transport or general storage capabilities access to spending keys
+  or signing authority;
 - avoid factory methods that imply a new independent backend when the feature is
   actually wallet-scoped;
 - bind wallet and account context once instead of passing raw IDs repeatedly;
@@ -138,7 +153,9 @@ Good capability APIs:
 
 Avoid APIs where callers manually assemble backend dependencies, open private
 databases, pass around handles, or call implementation steps in an undocumented
-order.
+order. These APIs obscure the authority being granted to the code and make it
+harder to audit which operations can perform IO, access secrets, or cross trust
+boundaries.
 
 ## State Machines And Workflows
 
--- a/docs/public-api-design-principles.md
+++ b/docs/public-api-design-principles.md
@@ -1,0 +1,373 @@
+# Android SDK Public API Design Principles
+
+Status: draft
+
+This document defines design principles for public Android SDK APIs. It is a
+working reference for contributors who are adding, changing, or reviewing SDK
+features that wallet applications will consume.
+
+The SDK is a product. Its public APIs should be understandable and safe for
+wallet teams that do not know this repository's internals, Rust/JNI boundaries,
+database layout, or first-party app code.
+
+Public SDK APIs should encapsulate the complex Zcash and wallet mechanics that
+apps should not need to reimplement.
+
+Use this document before implementation starts, while reviewing public API
+pull requests, and when deciding whether feature logic belongs in the SDK, an
+app, Rust, JNI, storage, or transport code.
+
+Feature-specific design documents must apply these principles to concrete
+types, methods, state diagrams, migration plans, and tests.
+
+## Product Stance
+
+The SDK must stand on its own.
+
+An app developer should be able to discover a public capability, understand the
+state machine or workflow it represents, and integrate it without reading:
+
+- first-party wallet app source;
+- SDK internals;
+- Rust source;
+- JNI carrier classes;
+- database schemas;
+- review threads or team chat history.
+
+"Works for one app" is not enough. A public SDK API is ready only when it is
+usable by arbitrary third-party wallet applications.
+
+## Public Boundary
+
+Public SDK consumers should depend on `sdk-lib` public types.
+
+They should not depend on:
+
+- `cash.z.ecc.android.sdk.internal.*`;
+- `backend-lib`;
+- `Jni*` carrier classes;
+- Rust backend wrapper classes;
+- raw database paths;
+- opaque native handles;
+- generated wire types unless a feature is explicitly designed around them;
+- first-party app model classes.
+
+The SDK may use those implementation details internally. The public layer maps
+them into SDK-domain models before values cross the consumer boundary.
+
+## Module Ownership
+
+Place public API in the module that matches its stability and audience.
+
+- `sdk-lib` is the main public SDK surface for wallet applications.
+- `backend-lib` is native/JNI/backend implementation detail and should not be a
+  direct dependency of wallet apps.
+- `sdk-incubator-lib` may expose experimental APIs that can later move into the
+  main SDK package, but consumers must expect higher churn.
+- `lightwallet-client-lib` exposes lower-level lightwalletd networking APIs.
+  These APIs can be useful extension points, but wallet-feature APIs in
+  `sdk-lib` should not leak raw network or unsafe representations when they can
+  expose validated SDK-domain models instead.
+- Demo, benchmark, and test modules are not public SDK feature surfaces.
+
+Incubation status is not a reason to expose internals. Experimental APIs still
+need clear ownership, typed models, documentation, and a migration path toward a
+stable public surface.
+
+## Representation Layers
+
+Data should move upward through representation layers before it reaches app
+code.
+
+Common representation layers are:
+
+- primitive values used by JNI, storage, or wire code;
+- JNI carrier objects that cross the native boundary;
+- generated network or protobuf models;
+- unsafe network models whose suffix indicates they are not fully validated;
+- internal SDK models used for orchestration and persistence;
+- public SDK-domain models exposed to wallet applications.
+
+Lower-layer representations may be appropriate inside adapters, repositories,
+or backend wrappers. Public feature APIs should expose the highest meaningful
+SDK-domain representation and keep lower-layer conversion internal.
+
+## Capability-Oriented APIs
+
+Expose features as narrow capabilities from established SDK entry points.
+
+A caller should receive an object that gives access to one feature area and no
+unrelated internals. For example, `Synchronizer` may expose a capability for a
+wallet-scoped feature, while that capability may expose account-scoped objects
+for operations that depend on `AccountUuid`.
+
+Good capability APIs:
+
+- make the starting point obvious;
+- avoid factory methods that imply a new independent backend when the feature is
+  actually wallet-scoped;
+- bind wallet and account context once instead of passing raw IDs repeatedly;
+- hide lifecycle steps that consumers cannot use safely;
+- expose only operations the caller is meant to perform.
+
+Avoid APIs where callers manually assemble backend dependencies, open private
+databases, pass around handles, or call implementation steps in an undocumented
+order.
+
+## State Machines And Workflows
+
+Public APIs should model user-facing workflows, not only backend calls.
+
+When a feature has meaningful states, recovery behavior, retries, or externally
+visible phases, the API design must describe that state machine before
+implementation is considered reviewable.
+
+The design should answer:
+
+- what states can exist;
+- what transitions are allowed;
+- what input the app must provide at each transition;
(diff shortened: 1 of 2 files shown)


docs/voting-public-api-design.md
```diff
@@ -0,0 +1,1071 @@
+# Shielded Voting Public API Design
```
greg0x, line 1: Note to reviewers: At this point this should be considered as a draft / illustration for how a design would look like following the current state principles doc. 

It's directionally correct, but details are likely imprecise.

docs/voting-public-api-design.md
```diff
@@ -0,0 +1,1071 @@
...
+    ): VotingConfiguration
+
+    /**
+     * Warms proving resources used by voting. Safe to call at app startup.
+     */
+    suspend fun warmProvingCaches()
+}
+
+@ExperimentalVotingApi
+interface AccountVoting : Closeable {
```
noop-sk: **`AccountVoting : Closeable` lifecycle gap**

The design states apps will not manage voting DB handles, but `AccountVoting` extending `Closeable` forces lifecycle management on callers. The document does not specify what happens to active `Flow` collections when `close()` is called, whether `close()` is idempotent, or what a `VotingRoundWorkflow` returned from a closed `AccountVoting` does. The capability-oriented principle says the SDK should "hide lifecycle steps consumers cannot use safely" — this gap needs to be resolved before the interface is implemented.

docs/voting-public-api-design.md
```diff
@@ -0,0 +1,1071 @@
...
+    /**
+     * Opens or creates local SDK voting state for an authenticated round.
+     */
+    fun round(round: AuthenticatedVotingRound): VotingRoundWorkflow
```
noop-sk, line 252: **`round()` StateFlow identity undefined**

`fun round(round: AuthenticatedVotingRound): VotingRoundWorkflow` is non-suspending and returns a `VotingRoundWorkflow` that holds a `StateFlow<VotingWorkflowSnapshot>`. The design does not state whether multiple calls with the same `AuthenticatedVotingRound` return the same cached instance or separate instances. If each call creates a new workflow object, an earlier caller's `state` collection silently goes stale while a later caller receives updates — a bug that is hard to detect in production.

docs/voting-public-api-design.md
```diff
@@ -0,0 +1,1071 @@
...
+@ExperimentalVotingApi
+data class VotingConfigurationPlan(
+    val previousFingerprint: VotingConfigurationFingerprint?,
+    val newFingerprint: VotingConfigurationFingerprint,
+    val configuration: VotingConfiguration,
+    val switchDecision: VotingConfigurationSwitchDecision
+)
+
+@ExperimentalVotingApi
+data class VotingConfigurationSwitchDecision(
```
noop-sk: **`VotingConfigurationSwitchDecision` single-field wrapper**

`VotingConfigurationSwitchDecision` currently wraps only `kind: VotingConfigurationSwitchKind`, making every call site write `plan.switchDecision.kind`. If the intent is to reserve room for additional fields (e.g. affected-server lists under `SameChainServiceUpdate`), the design should say so explicitly. Otherwise, expose `kind` directly on `VotingConfigurationPlan` and promote the wrapper when it gains a second field.

docs/voting-public-api-design.md
```diff
@@ -0,0 +1,1071 @@
...
+    val snapshotHeight: BlockHeight,
+    val createdAtHeight: BlockHeight,
+    val votingStartsAt: Instant,
+    val votingEndsAt: Instant,
+    val status: VotingRoundStatus,
+    val proposals: List<VotingProposal>,
+    val encryptionAuthorityPublicKey: VotingEncryptionAuthorityPublicKey,
+    val noteCommitmentRoot: VotingNoteCommitmentRoot,
+    val nullifierImtRoot: VotingNullifierImtRoot,
+    val rawSession: VotingSessionMetadata
```
noop-sk: **`rawSession` in a public model**

`AuthenticatedVotingRound.rawSession: VotingSessionMetadata` uses "raw" in a public API field name. The principles doc states lower-layer representations must be mapped before they cross the consumer boundary. The "raw" qualifier implies unvalidated or unmapped data is being handed directly to app code. Either rename the field to drop "raw" (e.g. `session`) and document what it contains, or explain why this cannot be promoted to a typed SDK-domain model.

docs/voting-public-api-design.md
```diff
@@ -0,0 +1,1071 @@
...
+        val scannedHeight: BlockHeight?,
+        val snapshotHeight: BlockHeight
+    ) : VotingEligibilityStatus
+}
+```
+
+Configuration switch models:
+
+```kotlin
+enum class VotingConfigurationSwitchKind {
```
noop-sk: **`VotingConfigurationSwitchKind` defined twice**

The enum appears at line 629 (with `@ExperimentalVotingApi`, inside the configuration section) and again here in the "Domain Models" subsection without the annotation. An implementor reading the design sees two conflicting canonical locations. Remove the duplicate and cross-reference the single definition, or clarify if the two blocks are intentionally separate types.

docs/voting-public-api-design.md
```diff
@@ -0,0 +1,1071 @@
...
+    class Unavailable(cause: Throwable? = null) : VotingException(...)
+    class InvalidConfiguration(...) : VotingException(...)
+    class ConfigurationPlanConflict(...) : VotingException(...)
+    class UnsupportedVersion(...) : VotingException(...)
+    class AccountUnavailable(...) : VotingException(...)
+    class WalletNotSynced(...) : VotingException(...)
+    class Ineligible(...) : VotingException(...)
+    class InvalidRoundState(...) : VotingException(...)
+    class InvalidSignature(...) : VotingException(...)
+    class HardwareSigningCancelled(...) : VotingException(...)
```
noop-sk: **`HardwareSigningCancelled` cancellation semantics unclear**

The design says "If `sign` is cancelled or the app process dies, the SDK keeps the workflow in `AwaitingExternalSignature`", but `HardwareSigningCancelled` is listed as a thrown `VotingException`. The design must clarify: (a) is this exception thrown when the user dismisses the hardware device UI — not a Kotlin `CancellationException` — and (b) does the SDK guarantee the workflow is already durably persisted as `AwaitingExternalSignature` before this exception propagates? Without that guarantee the app cannot know whether it is safe to offer a "resume" option or whether recovery state was never written.

docs/voting-public-api-design.md
greg0x: Noting [this comment](https://github.com/zcash/zcash-android-wallet-sdk/pull/1969#discussion_r3322710992). 
This doc is draft atm. After principles are good to go, the next step is to update and review this doc.

Keeping this comment unresolved until then.

docs/voting-public-api-design.md
greg0x: Noting [this comment](https://github.com/zcash/zcash-android-wallet-sdk/pull/1969#discussion_r3322710992). 
This doc is draft atm. After principles are good to go, the next step is to update and review this doc.

Keeping this comment unresolved until then.

docs/voting-public-api-design.md
greg0x: Noting [this comment](https://github.com/zcash/zcash-android-wallet-sdk/pull/1969#discussion_r3322710992). 
This doc is draft atm. After principles are good to go, the next step is to update and review this doc.

Keeping this comment unresolved until then.

docs/voting-public-api-design.md
greg0x: Noting [this comment](https://github.com/zcash/zcash-android-wallet-sdk/pull/1969#discussion_r3322710992). 
This doc is draft atm. After principles are good to go, the next step is to update and review this doc.

Keeping this comment unresolved until then.

docs/voting-public-api-design.md
greg0x: Noting [this comment](https://github.com/zcash/zcash-android-wallet-sdk/pull/1969#discussion_r3322710992). 
This doc is draft atm. After principles are good to go, the next step is to update and review this doc.

Keeping this comment unresolved until then.

docs/voting-public-api-design.md
greg0x: Noting [this comment](https://github.com/zcash/zcash-android-wallet-sdk/pull/1969#discussion_r3322710992). 
This doc is draft atm. After principles are good to go, the next step is to update and review this doc.

Keeping this comment unresolved until then.

docs/public-api-design-principles.md
```diff
@@ -0,0 +1,373 @@
...
+them into SDK-domain models before values cross the consumer boundary.
+
+## Module Ownership
+
+Place public API in the module that matches its stability and audience.
+
+- `sdk-lib` is the main public SDK surface for wallet applications.
+- `backend-lib` is native/JNI/backend implementation detail and should not be a
+  direct dependency of wallet apps.
+- `sdk-incubator-lib` may expose experimental APIs that can later move into the
```
nuttycom: Insofar as it is supported by the particular platform (Android/Kotlin or iOS/Swift) another good option to consider is feature-flagging. We use this extensively in the Rust crates, and a similar approach to modularising the SDKs would be a good idea, if possible. That being said, it's perhaps more important that there be parity between the structure of the two SDKs, so maybe using an indicator-lib module is good for that purpose (and incubation can complement feature-flagging).

docs/public-api-design-principles.md
```diff
@@ -0,0 +1,373 @@
...
+- avoid factory methods that imply a new independent backend when the feature is
+  actually wallet-scoped;
+- bind wallet and account context once instead of passing raw IDs repeatedly;
+- hide lifecycle steps that consumers cannot use safely;
+- expose only operations the caller is meant to perform.
+
+Avoid APIs where callers manually assemble backend dependencies, open private
+databases, pass around handles, or call implementation steps in an undocumented
+order.
+
```
nuttycom, line 159: There's an additional, and arguably more important, dimension to capability-oriented programming, which is security. It important for capability interfaces to clearly represent IO/boundaries. All IO should be mediated by a capability interface that is passed as an explicit argument to the function that performs the "business logic". This promotes a clean separation between such logic and the I/O that's used to implement it; in addition, the merit of a capability-oriented style is that you can understand immediately what permissions a function is granted by inspecting the capabilities it is provided and the types the methods of said capability interfaces consume and return. This is a critical element of making it easier to audit for security; if, for example, you see a capability that does IO requesting a spending key, that's an obvious red flag as that spending key could be going anywhere.

nuttycom (commented):
Overall, this looks like a vast improvement. Is there a similar design (or can this design be generalized) for the Swift wallet SDK?

greg0x:
> Overall, this looks like a vast improvement. Is there a similar design (or can this design be generalized) for the Swift wallet SDK?

(In the context of the principles doc) In a follow-up step, yes, it can be done.  But I would propose we do Android first, refine the doc during refactor then when swift sdk is next, we start the generalisation / refactor. Why? So the focus is sharper and the context is fresh.

docs/voting-public-api-design.md
greg0x: da3709c7

docs/voting-public-api-design.md
greg0x: 0b71871a

docs/voting-public-api-design.md
greg0x: 9b3159df

docs/voting-public-api-design.md
greg0x: 2dff8ab7

docs/voting-public-api-design.md
greg0x: a1d93170

docs/voting-public-api-design.md
greg0x: a783dfa3

docs/public-api-design-principles.md
greg0x: d085e3e6a9d4

docs/public-api-design-principles.md
greg0x: d732c9aeec15

nuttycom:
I've retargeted this PR from `prerelease/v2.6.0` to `merge/maint-2.8.x`.

`prerelease/v2.6.0` is frozen and fully merged into the current release lineage, so the PR was reporting mergeable only because its base had stopped moving — merging it would have landed the design doc on a branch nothing is built from. `merge/maint-2.8.x` is the live branch carrying the shielded-voting work forward; note that it is itself the head of #2075 (→ `main`), so this change is now stacked underneath that merge-back. The retarget was diff-neutral: still 2 files, +1834/−0, no conflicts.

Before this lands, though, the content needs a pass: it was written against the `zcash_voting` 1.x surface, and the public API described here needs to be adapted to the **`zcash_voting` 2.0 APIs**. Worth reconciling the documented types and call shapes against 2.0 before review.

<sub>Filed with Claude Code assistance.</sub>

PR #1977, opened 2026-06-13
Treat "already in mempool" submit rejections as success

Cosmos-Harry:
## Summary
lightwalletd / zcashd returns `transaction already exists in mempool` or `transaction dropped because it is already queued for download` as a non-zero submit error when an identical tx was previously broadcast. The submit was actually a success — the tx is on the network — but `Synchronizer.submitTransaction` (and the broadcaster equivalent that funnels through `Response<SendResponseUnsafe>.toSubmitResult`) currently classifies these messages as `TransactionSubmitResult.Failure`, which makes downstream UI render a misleading failure screen for a tx that landed.

This PR adds an internal helper `isAlreadyKnownToNetwork(message: String?)` that matches the two messages case-insensitively on substring (so wrapped / prefixed variants from different node versions all hit), and re-routes `toSubmitResult` to return `Success` for those messages.

## Changes
- `sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/SubmitTransaction.kt` — restructured `toSubmitResult` from `if/else` to `when`, added `isAlreadyKnownToNetwork` branch.
- `sdk-lib/src/test/java/cash/z/ecc/android/sdk/internal/transaction/SubmitResultClassificationTest.kt` — covers exact phrase, wrapped/prefixed variants, case-insensitivity, and unrelated failures.
- `CHANGELOG.md` — Unreleased / Fixed entry.

## Test plan
- [x] `./gradlew :sdk-lib:testDebugUnitTest --tests "*SubmitResultClassificationTest*"` — passes locally.


nesence-m (commented):
utACK

noop-sk (approved):

PR #1978, opened 2026-06-14
[MOB-1339] Route legacy create-and-submit through PendingSubmitPlanStore

Cosmos-Harry:
Linear: [MOB-1339](https://linear.app/zodl/issue/MOB-1339)

## What this PR does
Routes the legacy `Synchronizer.createProposedTransactions` and `Synchronizer.createTransactionFromPczt` helpers through `PendingSubmitPlanStore` so the sync loop's `resubmitUnminedTransactions` step skips in-flight submits instead of racing them with a second `txManager.submit()` and producing `transaction already exists in mempool` rejections.

Reviews surfaced two adjacent concerns that this PR now also addresses:

1. **Record after submit** — In the original commit, `addSubmitEndpoint(...)` was called *before* `transactionSubmitter.submit(...)`. That transitioned the store from `AwaitingPlan` → `Ready(endpoints)` before the actual submit RPC fired, leaving an in-flight window where the resubmit step no longer saw `AwaitingPlan` and could race the broadcast. The reorder applies to the public `Broadcaster.submit` as well, keeping the three call sites consistent.
2. **Survive submit-mid-cancellation** — A naive reorder would strand the plan at `AwaitingPlan` forever if `transactionSubmitter.submit(...)` threw a `CancellationException` (app teardown, scope cancel), because `addSubmitEndpoint` would never run and the resubmit loop skips `AwaitingPlan` entries. The three call sites fold into a `recordingEndpointAfterSubmit` helper that wraps `addSubmitEndpoint` in `try/finally` with `withContext(NonCancellable)`, matching the existing `EndpointTransactionSubmitter.submit` dispose pattern in this file.

The PR also clarifies a misleading comment about `SubmitPlanExecutor` "dedupe" semantics (it iterates `endpoints.distinct()` and stops on first non-Failure; it doesn't dedupe per-endpoint), and documents why `createSubmitResultFlow` leaves later legs of a partially-failed multi-tx proposal at `AwaitingPlan` (dependent proposals — shield → spend — shouldn't retry the later leg if an earlier leg failed).

## Files
- **Modified**: `sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcaster.kt` — legacy helpers route through `PendingSubmitPlanStore`; public `Broadcaster.submit` and both legacy helpers fold into `recordingEndpointAfterSubmit` (record after submit, `NonCancellable` finally).
- **Modified**: `sdk-lib/src/test/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcasterTest.kt` — coverage for plan registration on both legacy paths, mutex-during-create contract, the `AwaitingPlan`-during-submit window for both the legacy helper and the public `Broadcaster.submit`, and the cancellation-mid-submit contract.
- **Modified**: `CHANGELOG.md` — Unreleased / Fixed entry.

## Test plan
- [ ] CI's `test_android_modules_unit` runs the full `SdkBroadcasterTest` suite (verify the new `broadcaster_submit_records_endpoint_even_if_cancelled_mid_submit` and the renamed/added blocking-submitter tests).
- [x] Existing `submit_targets_requested_endpoint` continues to pass (post-submit state is `Ready` with endpoint recorded).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- The legacy `Synchronizer.createProposedTransactions` and `Synchronizer.createTransactionFromPczt`
+  helpers now register transactions in `PendingSubmitPlanStore` so the sync loop's
+  `resubmitUnminedTransactions` step correctly skips in-flight submits instead of racing them
+  with a second `txManager.submit()` and producing `transaction already exists in mempool`
+  rejections.
+
 ### Added
 - New wallets now fetch a recent tree state from the lightwalletd server, reducing unnecessary block
   scanning for wallets with no transaction history while retaining reorg safety. Initialization falls back
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,19 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - The legacy `Synchronizer.createProposedTransactions` and `Synchronizer.createTransactionFromPczt`
-  helpers now register transactions in `PendingSubmitPlanStore` so the sync loop's
-  `resubmitUnminedTransactions` step correctly skips in-flight submits instead of racing them
-  with a second `txManager.submit()` and producing `transaction already exists in mempool`
-  rejections. The public `Broadcaster.submit` and both legacy helpers also record their endpoint
-  after the submit RPC returns (rather than before), so the in-flight window stays at
-  `AwaitingPlan`; the endpoint write is wrapped in `NonCancellable` so a coroutine cancellation
-  mid-submit cannot leave the plan stranded at `AwaitingPlan` forever.
+  helpers now register transactions in `PendingSubmitPlanStore`. Before this change the legacy
+  paths bypassed the plan store entirely, so a sync-loop `resubmitUnminedTransactions` tick that
+  fired during the active `submit()` RPC could race the foreground submit with a second
+  `txManager.submit()`. With the plan-store dance, the in-flight window is `AwaitingPlan` and the
+  resubmit step skips it. Note that `resubmitUnminedTransactions` is DB-driven (loads
+  unmined-and-not-expired txs from the wallet DB) and does not query the mempool, so a tx that
+  has been accepted into a server's mempool but not yet mined will still be re-broadcast on the
+  next sync tick — that mempool-duplication path is handled by the "verify against the server"
+  reclassification (separate `## Fixed` entry below). This entry narrows the *in-flight* race
+  window specifically. The public `Broadcaster.submit` and both legacy helpers record their
+  endpoint after the submit RPC returns (rather than before), and the write is wrapped in
+  `NonCancellable` so a coroutine cancellation mid-submit cannot leave the plan stranded at
+  `AwaitingPlan`.
 - `Synchronizer.submitTransaction` (and the broadcaster equivalent) now verifies submit failures
   against the server before surfacing them: when the submit RPC returns a non-zero error code
   (and not a gRPC-layer failure), the SDK immediately asks the same lightwalletd whether the tx
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,10 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   helpers now register transactions in `PendingSubmitPlanStore` so the sync loop's
   `resubmitUnminedTransactions` step correctly skips in-flight submits instead of racing them
   with a second `txManager.submit()` and producing `transaction already exists in mempool`
-  rejections.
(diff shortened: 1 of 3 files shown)


sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcaster.kt
```diff
@@ -45,36 +45,56 @@ internal class SdkBroadcaster(
...
+            }
+
+    // Record the endpoint after submit returns so the in-flight window stays at AwaitingPlan
+    // and the sync loop's resubmit step doesn't race it. NonCancellable so a cancelled submit
+    // doesn't strand the plan at AwaitingPlan forever — the resubmit loop skips that state.
+    private suspend fun recordingEndpointAfterSubmit(
+        transaction: CreatedTransaction,
+        endpoint: LightWalletEndpoint
+    ): TransactionSubmitResult {
+        try {
```
noop-sk, line 91: **`finally` fires on RPC failure too**

The `try/finally` guarantees `addSubmitEndpoint` runs on cancellation — that's the stated goal. But `finally` also runs unconditionally when `transactionSubmitter.submit()` throws any non-cancellation exception (gRPC timeout, network error, etc.). That transitions the plan from `AwaitingPlan` → `Ready(endpoint)` for a transaction that never reached the mempool. The resubmit loop's subsequent treatment of that `Ready` entry is now load-bearing for retry correctness.

If the intent on failure is "stay `AwaitingPlan` so the resubmit loop retries", the `finally` should be narrowed to cancellation only (e.g., a separate `catch(e: CancellationException)` that still writes the endpoint, vs a non-cancellation catch that rethrows without the store write). If "transition to `Ready` even on failure" is intentional, a test covering this path would lock in the contract.

sdk-lib/src/test/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcasterTest.kt
```diff
@@ -278,7 +471,8 @@ class SdkBroadcasterTest {
     private class FakeTransactionSubmitter(
         private val resultFactory: (CreatedTransaction) -> TransactionSubmitResult = {
             TransactionSubmitResult.Success(it.txId)
-        }
+        },
+        private val beforeReturning: suspend () -> Unit = {}
```
noop-sk, line 500: **Missing test: submit throws non-cancellation exception**

The cancellation path has a dedicated test (`broadcaster_submit_records_endpoint_even_if_cancelled_mid_submit`), but there is no test for `transactionSubmitter.submit()` throwing a non-cancellation exception (e.g., `IOException`). `FakeTransactionSubmitter.beforeReturning` can already simulate this by throwing. A test verifying the expected plan state — `AwaitingPlan` (retryable) or `Ready` (endpoint recorded) — after a submit exception would lock in the intended state machine contract.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/SdkBroadcaster.kt
zcash-harry: the `try/finally + NonCancellable `does fire on every exit path, not just cancellation. The "every path" semantics are intentional: leaving the plan at `AwaitingPlan` on a non-cancellation throw would strand the tx the same way the cancellation case did, since the resubmit loop skips `AwaitingPlan` entries (treating them as "never submitted from this session"). Recording the endpoint moves the plan to `Ready`, which is the state the resubmit loop retries through `SubmitPlanExecutor`.

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- The legacy `Synchronizer.createProposedTransactions` and `Synchronizer.createTransactionFromPczt`
+  helpers now register transactions in `PendingSubmitPlanStore` so the sync loop's
+  `resubmitUnminedTransactions` step correctly skips in-flight submits instead of racing them
```
nuttycom: I'm just going off of the documentation here, but does `resubmitUnminedTransactions` actually monitor the mempool to determine whether the transaction is already known to the server? Otherwise, it seems like this jut moves the mempool-duplication rejection to a different spot. "Mined" and "not yet submitted" aren't the only possible states for the transaction; they can linger unmined in the mempool until expiration.

nuttycom (dismissed):
utACK with nonblocking question

CHANGELOG.md
zcash-harry: Yes, sorry updated the CHANGELOG.
`findTransactionsEligibleForResubmission` is purely DB-driven, no mempool query. This PR only closes the in-flight `submit()` RPC window (legacy paths previously bypassed `PendingSubmitPlanStore`, letting a sync tick race the active submit). The mempool-pending duplication case is handled separately by the verify-against-server reclassification already on main #1977

noop-sk (approved):

PR #1979, opened 2026-06-14
[PRO-336] Diagnostic logging for the enhance pipeline

Cosmos-Harry:
Linear: [PRO-336](https://linear.app/zodl/issue/PRO-336)

> Substantially revised after feedback on the Swift sister PR ([#1761](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1761) — intra-cycle sleep blocking the sync loop) plus a deeper read of what actually drives the user-visible "Receiving forever" bug. The original tracker subsystem is removed in the second commit on this branch; see commit history for the journey.

## What this PR does now
Adds structured, metadata-only diagnostic logging to `CompactBlockProcessor.enhanceTransactionDetails` and the private `enhanceTransaction` step so engineers can determine what's actually happening when a user reports a stuck "Receiving" transaction. No behavioural change: no new retry strategy, no sleeps, no new state, no tracker, no retry-count bump.

Log shape (one cycle):

```
BlockEnhancer cycle started [a7f3c1] requests=3
BlockEnhancer [a7f3c1.0] type=enhancement start
BlockEnhancer [a7f3c1.0] fetch returned has_tx=true has_mined_height=true
BlockEnhancer [a7f3c1.0] decryptAndStoreTransaction called has_mined_height=true
BlockEnhancer [a7f3c1.1] type=enhancement start
BlockEnhancer [a7f3c1.1] type=enhancement failed error_type=EnhanceTransactionError
...
BlockEnhancer cycle complete [a7f3c1]
```

Opaque correlation ids (`<6-hex>.<index>`) let engineers trace one request's journey through a session's logs without exposing the txid.

The earlier txid-leaking lines (`txid: ${transactionRequest.txIdString()}` in `Starting enhancing transaction…`, `Resolving TransactionDataRequest…`, etc.) are replaced with metadata-only variants.

## Why this and not the original tracker
Reading the code more carefully (and combined with a recent user-support thread): the user-visible failure that the original tracker would have skipped isn't actually surfaced to the UI by the resubmit/enhance path. The original `EnhanceFailureTracker` + bumping `TRANSACTION_FETCH_RETRIES` to 5 didn't change the user-visible outcome — it just added ~15s of sleep budget to the sync loop while making the SDK politer about a failure mode whose root cause we can't yet pinpoint without log evidence.

The three plausible hypotheses for PRO-336 — (1) the auto-loop's `fetchTransaction` failing repeatedly, (2) the server returning `Success(tx, height: nil)` for an actually-mined tx, (3) the rust backend's `transactionDataRequests()` no longer including the request after one mempool-only call — point to fundamentally different fixes. Logging is the prerequisite to picking the right one.

## Files
- **Modified**: `sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt` — adds per-cycle / per-request structured logs, opaque correlation ids, replaces pre-existing txid-leaking debug lines with metadata-only variants. Reverts `TRANSACTION_FETCH_RETRIES` to its pre-PR value.
- **Modified**: `sdk-lib/src/main/java/cash/z/ecc/android/sdk/SdkSynchronizer.kt` — reverts the `enhanceFailureTracker` plumbing from the original PR.
- **Modified**: `CHANGELOG.md` — Unreleased / Added entry.
- **Deleted**: `EnhanceFailureTracker.kt` and `EnhanceFailureTrackerTest.kt`.

## Privacy
No transaction ids, addresses, or other PII appear in any new log line. This aligns with prior security-review guidance that sensitive identifiers should not appear in release-build logs.

## Test plan
- [x] `./gradlew :sdk-lib:compileDebugKotlin ktlint detektAll` passes.
- [ ] Verify in a real wallet that cycle/request logs appear with reasonable cadence and no PII.


sdk-lib/src/test/java/cash/z/ecc/android/sdk/block/processor/EnhanceFailureTrackerTest.kt
```diff
@@ -0,0 +1,95 @@
+package cash.z.ecc.android.sdk.block.processor
+
+import kotlinx.coroutines.runBlocking
+import org.junit.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class EnhanceFailureTrackerTest {
```
noop-sk: **Max-backoff cap is untested**

`backoffFor` is supposed to plateau at `MAX_BACKOFF_MILLIS` (30 min) once enough failures accumulate, but no test exercises that cap. If the exponent calculation regresses — e.g. a future refactor changes `coerceAtLeast(0)` to `coerceAtLeast(1)` — the cap would never fire and a transaction could be throttled for far longer than intended. Add a test that records 20+ failures and asserts `shouldSkipDueToBackoff` returns false after `MAX_BACKOFF_MILLIS + 1` ms have elapsed.

sdk-lib/src/test/java/cash/z/ecc/android/sdk/block/processor/EnhanceFailureTrackerTest.kt
```diff
@@ -0,0 +1,95 @@
...
+
+            tracker.recordFailure(txid1)
+            now = 100_000L
+            tracker.recordFailure(txid2)
+            now = 120_000L
+
+            assertFalse(tracker.shouldSkipDueToBackoff(txid1))
+            assertTrue(tracker.shouldSkipDueToBackoff(txid2))
+        }
+}
```
noop-sk: **Retention-window eviction is untested**

The PR description calls out "Bounded by a 2-hour retention window" as the memory safety property, but the `removeAll` eviction path inside `recordFailure` is never exercised by a test. Add a case: record a failure, advance the clock past `MAX_RETENTION_MILLIS`, call `recordFailure` on any second txid to trigger eviction, then assert `shouldSkipDueToBackoff` on the first txid returns false and behaves as if it were unknown.

zcash-harry:
Just settled for some debugging logs for now to troubleshot the root cause more. Otherwise, it's a fix based on a hypothesis.

nesence-m (approved):
utACK

nuttycom (approved):
utACK; this looks like it actually helps with user privacy in sharing these logs in addition to giving better telemetry

PR #1988, opened 2026-06-18
Harmonize rust toolchain versions across CI and local dev

### LINKED ISSUE #1995: Harmonize rust toolchain versions across CI and local dev

dannywillems:
Relying on 1.84.1 makes the CI fails.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,6 +1,11 @@
 name: 'Setup Java, Rust, and Dependency Cache'
 description: "Configures the build environment and caches Gradle, dependencies, and build outputs."
 inputs:
+  # rust-toolchain.toml must be in sync with this default.
+  rust-version:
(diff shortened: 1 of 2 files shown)


noop-sk (approved):

nesence-m (approved):
utACK

.github/actions/setup/action.yml
```diff
@@ -27,8 +32,8 @@ runs:
     - name: Setup Rust
       shell: bash
       run: |
-        rustup install 1.84.1
-        rustup default 1.84.1
+        rustup install ${{ inputs.rust-version }}
```
nuttycom, line 35: We should instead consider using `dtolnay/rust-toolchain` for the action, like librustzcash does, instead of manually running rustup.

dannywillems: See https://github.com/zcash/zcash-android-wallet-sdk/pull/1999

PR #1996, opened 2026-06-19
CI: attempt to get rid of emulator.wtf

### LINKED ISSUE #1998: Get rid of emulator.wtf
We currently use emulator.wtf for testing. However, we only use the free plans, and the CI fails when we run out of credits.
Instead, we could build the workflow from a vanilla ubuntu-latest.

dannywillems:


noop-sk (dismissed):

nesence-m (approved):
utACK

PR #2007, opened 2026-06-29
MOB-1397: Catch SeedNotRelevant in WalletCoordinator to prevent crash…

noop-sk:
… loop

Wrap Synchronizer.new() in callbackFlow with a catch for SeedNotRelevant,
emitting a SeedMismatch status and exposing isSeedMismatch: StateFlow<Boolean>
via a MutableStateFlow set in the catch block.

Also extract withSynchronizerLockout() helper used by both resetSdk() and deleteSdkDataFlow() to ensure the synchronizer is stopped before any erase.

> **Note**

# Author

# Reviewer

### DIFF
--- a/sdk-incubator-lib/src/main/java/cash/z/ecc/android/sdk/WalletCoordinator.kt
+++ b/sdk-incubator-lib/src/main/java/cash/z/ecc/android/sdk/WalletCoordinator.kt
@@ -1,6 +1,7 @@
 package cash.z.ecc.android.sdk
 
 import android.content.Context
+import cash.z.ecc.android.sdk.exception.InitializeException
 import cash.z.ecc.android.sdk.ext.onFirst
 import cash.z.ecc.android.sdk.internal.Twig
 import cash.z.ecc.android.sdk.model.AccountCreateSetup
@@ -16,6 +17,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.callbackFlow
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.filter
@@ -76,7 +78,12 @@ class WalletCoordinator(
         class Lockout(
             val id: UUID
         ) : InternalSynchronizerStatus()
+
+        object SeedMismatch : InternalSynchronizerStatus()
     }
+
+    private val _isSeedMismatch = MutableStateFlow(false)
+    val isSeedMismatch: StateFlow<Boolean> = _isSeedMismatch.asStateFlow()
 
     @Suppress("DestructuringDeclarationWithTooManyEntries")
     @OptIn(ExperimentalCoroutinesApi::class)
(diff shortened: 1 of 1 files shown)


nesence-m (changes requested):
First checked staleness (old PR number): the head was freshly pushed and rebased onto the Slipstream-era file — targets `engineSynchronizerFactory.new`, preserves the `isSyncBlocked` job, extends `deleteSdkDataFlow`, and the companion app PR zodl-inc/zodl-android#2351 depends on the new `isSeedMismatch`. So this was reviewed on the merits, not dismissed as stale. One blocking finding:

**1. BLOCKING — the catch never fires on the default (Slipstream) engine build.** `WalletCoordinator.kt:155` catches `InitializeException.SeedNotRelevant` around `engineSynchronizerFactory.new(...)`, which works for the upstream engine where `Synchronizer.new` throws synchronously. But the base sets `IS_SLIPSTREAM_ENABLED=true`, and on that path `SlipstreamSynchronizer.new` returns immediately with deferred preparation: `SeedNotRelevant` is thrown from `prepare()`'s `initDataDb` switch (`SlipstreamSynchronizer.kt:691`) inside `runPrepare`, launched on the synchronizer's own scope — `runPrepare` catches `Throwable`, latches it, sets status DISCONNECTED, and invokes `onSetupErrorHandler`; the exception never propagates out of `new()`. Net effect on the shipping build: no crash (that's already gone under Slipstream), but `isSeedMismatch` never turns true, the coordinator emits an `Available` synchronizer that is permanently DISCONNECTED, and #2351's auto-recovery never fires — a silent dead wallet instead of the intended recovery. Fix direction: after `trySend(Available(...))`, attach `closeableSynchronizer.onSetupErrorHandler` and map `SeedNotRelevant` to the mismatch state — Slipstream's handler contract deliberately replays a latched error into a late-attached handler, so this is race-safe by design; keep the try/catch for the upstream engine.

**Non-blocking:**

2. `WalletCoordinator.kt:94-95,159-161` — `isSeedMismatch` reset semantics are fragile and undocumented. The reset lives in the catch-path `awaitClose`, which runs on any cancellation of the inner flow — including subscriber count dropping to zero (`SharingStarted.WhileSubscribed(0, 0)`). Every UI stop/re-collect flickers the flag false → true, so an edge-triggered consumer (exactly #2351's auto-recovery) can re-fire the wipe repeatedly, and `false` never means "resolved". The new public property on a published incubator artifact has no KDoc; it needs one stating the level-triggered, collection-scoped contract.

3. `InitializeException.SeedRequired` from the same `initDataDb` switch (Slipstream line 690; upstream equivalent) still propagates into `walletScope` — `GlobalScope + Dispatchers.Main`, no `SupervisorJob` — reproducing the exact crash loop this PR fixes, one exception type over. Worth a comment or follow-up ticket.

4. No `CHANGELOG.md` entry for the new `isSeedMismatch` API or for `deleteSdkDataFlow`/`resetSdk` now taking a synchronizer lockout (observable: erase waits for a live synchronizer instead of racing it).

5. No tests — pre-existing gap (no `WalletCoordinator` test exists), but the PR adds untested state-machine behavior; even a Turbine-style test of mismatch emission/reset would help. The checklist's automated-tests box is unticked.

6. Nit (pre-existing, but the refactor touched these lines): `deleteSdkDataFlow` (`WalletCoordinator.kt:261-275`) — if `persistableWallet.first()` is null the job returns without `trySend`, and a collector expecting one Boolean waits forever.

**Verified good:** the `withSynchronizerLockout` extraction is faithful to the old `resetSdk` logic (lockout-id filter + `onFirst`; `lockoutMutex → synchronizerMutex` ordering consistent everywhere, no deadlock), and lockout-aware `deleteSdkDataFlow` fixes a real race where the engine erase throws while a synchronizer is live.

PR #2017, opened 2026-07-20
CI: check the Kotlin and Rust sides of the JNI boundary agree

dannywillems:
> Stacked on #2016. Base is `dw/run-rust-tests-in-ci`, because both PRs touch
> `scripts/ci-local.sh`, `.github/workflows/pull-request.yml`, and `AGENTS.md`.
> Review/merge #2016 first; GitHub will retarget this to `main` automatically.

## Problem

The 138 Kotlin `external fun` declarations and the 138 Rust `Java_*` exports are
kept in sync entirely by hand. Nothing verifies them.

That matters because of how JNI resolves a native method. It looks up the
**short** name `Java_pkg_Cls_method`, which encodes no argument types at all;
type-mangled long names are consulted only for overloaded natives. So:

- Wrong **name** -> `UnsatisfiedLinkError`. Loud, safe, obvious.
- Right name, wrong **types or arity** -> the VM links it and calls it. No error.

In the second case arguments are marshalled per the Kotlin declaration and read
per the Rust signature. On arm64 a dropped parameter shifts every later argument
by one register, and an `Int` read as `jlong` picks up garbage in the high 32
bits. For the functions taking a `jlong` pointer handle, that is
`Box::from_raw` on a corrupted pointer: silent memory corruption rather than an
exception.

Nothing else in the toolchain catches this. Clippy cannot see the Kotlin side,
and Android Studio checks signatures only for implicitly-bound methods in an IDE
session.

## Change

`scripts/check-jni-abi.sh` parses both sides and compares:

1. **Symbol sets** - every declaration has an export and vice versa.
2. **Arity** - after dropping the Rust `JNIEnv` + `JClass` prefix.
3. **Parameter types** - via the Kotlin-to-JNI mapping (`Int`/`jint`,
   `Long`/`jlong`, `String`/`JString`, `ByteArray`/`JByteArray`,
   `Array<T>`/`JObjectArray`, other objects/`JObject`).

Pure text analysis: no Rust or Android toolchain, about three seconds. It runs
as its own CI job (`static_analysis_jni_abi`) and as the `jniabi` stage in
`ci-local.sh`, which is in `fast`.

## Testing

Reports `138 JNI exports agree on name, arity, and types` on the current tree.

A check that only ever passes is worthless, so each drift mode was verified by
mutating a declaration and confirming a non-zero exit:

| Mutation | Detected as |
|---|---|
| `freeTorRuntime` renamed to `freeTorRuntimeX` | missing on both sides, with the UnsatisfiedLinkError consequence spelled out |
| `(nativeHandle: Long)` -> `(nativeHandle: Long, flags: Int)` | arity mismatch, Kotlin 2 vs Rust 1 |
| `(nativeHandle: Long)` -> `(nativeHandle: Int)` | type mismatch, `jint` vs `jlong` |

The third is the important one: that is the silent-corruption case, and it is
exactly the shape of a `jlong` handle being mis-declared.

Also verified: `shellcheck` clean (no suppressions), workflow YAML parses, exit
code 1 on drift and 0 when clean, and the script fails loudly rather than
silently passing if it parses zero declarations or meets a nested-class name
needing `_00024` mangling.

## Limits, stated deliberately

- Return types are not checked.
- Object types are compared only as "is an object". A `JniAccountUsk` passed
  where a `JniBlockMeta` is expected is not caught, because both are `JObject`
  at the JNI level. Catching that needs full descriptor mangling, which is a
  much larger and more brittle job; this closes the primitive-width and arity
  cases, which are the ones that corrupt memory.
- `_1` (underscore in an identifier) mangling is not implemented. No method
  needs it today; `_00024` is detected and rejected loudly rather than
  mis-mangled.


scripts/check-jni-abi.sh
```diff
@@ -0,0 +1,265 @@
...
+
+# --- Rust side -------------------------------------------------------------
+# Emits: <symbol> TAB <type,type,...>   with JNIEnv and JClass dropped.
+extract_rust() {
+    # shellcheck disable=SC2016
+    # The awk program is intentionally single-quoted; $0/$1 are awk fields, not
+    # shell parameters.
+    find "${RUST_SRC}" -name "*.rs" -print0 \
+        | xargs -0 awk "${AWK_COMMON}"'
+        /pub extern "C" fn Java_/ {
```
nesence-m, line 90: **Verified silent pass: commented-out exports still count.** This regex matches inside comments. I block-commented the entire `freeTorRuntime` export (`/* ... */` around the whole function) in a copy of this tree and the script still reported `138 JNI exports agree` with exit 0 — while the shipped `.so` would lack the symbol and the first call would throw `UnsatisfiedLinkError` in production. A `//`-commented export is also still extracted (it happened to fail in my test, but only by the accident of `//` junk corrupting the parsed parameter list into an arity mismatch — not by design).

Commenting out a native export while forgetting its Kotlin declaration is a realistic refactoring state, and it is precisely the drift this checker exists to catch. Suggested fix: strip comments before matching — a pre-pass that deletes `/* ... */` ranges and `//`-to-EOL (outside string literals is overkill here; these files don't put `pub extern` in strings) closes both cases cheaply.

Minor, same line: the match is hard-wired to `pub extern "C"`. The jni crate's documented convention is `extern "system"`, and `pub unsafe extern "C"` is also legal. A future export written in either form is silently skipped — that direction fails loud (its Kotlin decl reports "missing Rust export"), but widening to `pub( unsafe)? extern "(C|system)"` avoids the false alarm.

scripts/check-jni-abi.sh
```diff
@@ -0,0 +1,265 @@
...
+if [ "${rust_count}" -eq 0 ] || [ "${kotlin_count}" -eq 0 ]; then
+    echo "error: parsed ${rust_count} Rust exports and ${kotlin_count} Kotlin" >&2
+    echo "       declarations. Expected both to be non-zero; the parser is" >&2
+    echo "       broken or the source layout moved." >&2
+    exit 1
+fi
+
+# Mangling forms this script does not implement. Fail loudly rather than
+# emitting a name that silently does not match.
+if cut -f1 "${KOTLIN_TABLE}" | grep -q '_00024'; then
```
nesence-m, line 212: **Verified: this guard is dead code — it can never fire.** The Kotlin table's symbol column is synthesized a few lines up as `Java_ + PKG + CLS + method`, where `CLS` comes from `basename "${file}" .kt` and the method name matches `[a-zA-Z0-9_]+`. None of those components can ever contain the literal string `_00024`, so this `grep -q` is unreachable.

I verified by adding `@JvmStatic external fun nestedThing(x: Long)` inside a nested `object Inner` in `TorClient.kt`: the guard did not fire. Instead the script fabricated the unmangled name `Java_..._TorClient_nestedThing` and reported it as "declared in Kotlin with no matching Rust export" — loud, but the wrong diagnosis, and it invites the dangerous "fix": if someone then writes the Rust export under that same unmangled name to make the checker green, the checker passes while the VM looks up `Java_..._TorClient_00024Inner_nestedThing` and throws `UnsatisfiedLinkError` at runtime.

Suggested fix, two cheap parts: (a) grep the **Rust** table for `_00024` — that is the side where the mangled form would textually appear; (b) on the Kotlin side, detect that an `external fun` is declared under a nested type (e.g. track that only one top-level `class`/`object` plus its `companion object` has been seen before the declaration, or simply count `class |object ` nesting) and fail with the mangling message this block already has.

scripts/check-jni-abi.sh
```diff
@@ -0,0 +1,265 @@
...
+            if (sym != "") buf = buf " "
+        }
+    ' | sort
+}
+
+# --- Kotlin side -----------------------------------------------------------
+# Emits: <mangled symbol> TAB <expected rust type,...>
+# The JNI short name is Java_ + package with dots replaced by underscores +
+# class + method. An underscore in a Kotlin identifier would mangle to _1 and a
+# nested class to _00024; neither occurs today, and both are rejected loudly
```
nesence-m, line 129: **Verified silent pass: `_1` is not "rejected loudly" — it isn't detected at all.** This comment (and the PR description) says both mangling forms are rejected loudly; only `_00024` even has a guard (and that guard is unreachable — see below on line 212). For `_1` there is no check anywhere, and the failure mode is a *checker-green, runtime-broken* state: I renamed `freeTorRuntime` to `free_torRuntime` in Kotlin and renamed the Rust export to the naive `Java_..._TorClient_free_torRuntime` — the script reported `138 JNI exports agree`, exit 0. At runtime the VM would look up `Java_..._TorClient_free_1torRuntime` and throw `UnsatisfiedLinkError`. That both-sides-naive state is exactly what a developer produces by eyeballing the existing symbol pattern.

Cheap fix: after extraction, fail loudly if any Kotlin method name (the segment after `CLS`) contains an underscore — same policy this script already intends for `_00024`.

scripts/check-jni-abi.sh
```diff
@@ -0,0 +1,265 @@
...
+#   ./scripts/check-jni-abi.sh --list   # print the parsed table, then check
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+cd "${REPO_ROOT}"
+
+RUST_SRC="backend-lib/src/main/rust"
+KOTLIN_SRC="backend-lib/src/main/java"
```
nesence-m, line 47: **Scope-drift hardening (minor).** The scan roots are hard-coded, and the zero-parse guard below cannot notice a *partial* miss: an `external fun` added under `sdk-lib/` or another source set (or a `fn Java_` outside `backend-lib/src/main/rust`) is silently out of scope while the check stays green — the same "quiet non-coverage" class the zero-count guard exists for. Today every declaration lives under these two roots (I checked both this head and current main), so a cheap invariant keeps it that way: repo-wide `git grep -l 'external fun' -- '*.kt'` / `git grep -l 'fn Java_' -- '*.rs'` and fail if any hit falls outside the scanned roots.

nesence-m (commented):
This is a well-motivated check and the engineering discipline shows: the zero-parse guard, the mutation-tested drift modes, the sha-pinned action, and the honest "limits" section are all exactly right. I verified the script passes on this branch (138 exports, ~1s) and — notably — it also passes unmodified against current `main` (134 exports agree), so a rebase should be painless for the script itself. The findings below are all in the same category the PR itself cares about: paths where the checker can go green while the runtime boundary is broken. Three of them I confirmed empirically by mutation.

**Base branch / retarget.** The PR description says "GitHub will retarget this to `main` automatically" once #2016 merges — but #2016 was **closed unmerged** (it targeted `maint/v2.5.x`), so the auto-retarget will never happen. The base `dw/run-rust-tests-in-ci` now sits on a 2.5.x-derived history, which is why this PR shows 215 commits / 82 files / `mergeable_state: dirty` — the visible diff is almost entirely other people's already-landed work. Meanwhile the substance of the stacked prerequisites already landed on `main` independently (`test_rust_unit` and `static_analysis_clippy` both exist in main's `pull-request.yml` today). The real delta of this PR is the single tip commit 5f20631. Suggest: retarget to `main` and rebase just that commit onto it; the `ci-local.sh` / `pull-request.yml` / `AGENTS.md` hunks will need manual conflict resolution against main's current versions (stage numbering has moved on), but `check-jni-abi.sh` itself drops in cleanly.

**One forward-looking note** (no action needed on this branch): the script compares *source text*, not the shipped binary. On branches where JNI families are feature-gated (`#[cfg(feature = ...)]` — the slipstream/migration modules on the ironwood lines), a symbol that is compiled out of the default `.so` still counts as present, and the check stays green while the runtime raises `UnsatisfiedLinkError`. Not an issue on this head (`mod voting;` is unconditional) or on current main, but worth a sentence in the script header so nobody mistakes a green check for a binary-level guarantee. The binary-level complement already exists in release tooling (`llvm-readelf --dyn-syms` on the merged `.so`).

Line comments below cover the three verified silent-pass/dead-guard paths and one scope-drift hardening suggestion.

PR #2021, opened 2026-07-23
Android slipstream ironwood chp

nesence-m:
> **Note**

# Author

# Reviewer


sdk-lib/src/main/java/cash/z/ecc/android/sdk/MigrationSdk.kt
```diff
@@ -0,0 +1,745 @@
...
+
+/**
+ * Kotlin interface for the Orchard → Ironwood migration SDK bridge.
+ *
+ * Boundary: SDK owns all business logic (split algorithm, anchor height selection,
+ * transaction signing, storage, invalidity detection). App owns UI, WorkManager
+ * scheduling, permissions, and notifications.
+ *
+ * Based on: Orchard Migration Flow — Implementation Proposal (Path A)
+ * Sync: 2026-06-17
```
dannywillems, line 18: Do we want that in the final code?

backend-lib/src/main/rust/slipstream/host_read.rs
```diff
@@ -0,0 +1,458 @@
...
+            JValue::Object(&to_address),
+            JValue::Object(&to_account_uuid),
+        ],
+    )?)
+}
+
+/// Builds `listTransactions`' SQL: base projection, an optional reconciliation LEFT JOIN +
+/// filter when `is_recovering`, an optional account filter, ORDER BY. Verbatim from the
+/// Kotlin `VisibleTransactionsQuery` this replaces — see FFI_JNI_CONTRACT.md §9.3.
+fn list_transactions_sql(is_recovering: bool, has_account_filter: bool) -> String {
```
dannywillems, line 192: Don't we want to have the whole logic on the slipstream side?

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/model/CompactBlockUnsafe.kt
```diff
@@ -17,6 +17,10 @@ class CompactBlockUnsafe(
     val time: Int,
     val saplingOutputsCount: UInt,
     val orchardOutputsCount: UInt,
+    // Temporary diagnostic field (IRONWOOD_DIAG) — added to confirm whether the connected
```
dannywillems, line 22: Outdated?

lightwallet-client-lib/src/main/proto/compact_formats.proto
```diff
@@ -52,6 +53,10 @@ message CompactTx {
     repeated CompactSaplingSpend spends = 4;
     repeated CompactSaplingOutput outputs = 5;
     repeated CompactOrchardAction actions = 6;
+    // Ironwood (NU6.3) actions. Ironwood is Orchard note-version V3, so these reuse the
```
dannywillems, line 56: Useful comment, if it follows a standard/ZIP?

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/migration/JniMigrationModels.kt
```diff
@@ -0,0 +1,175 @@
...
+                "Height $nextTransferReadyAtHeight is outside of allowed UInt range and is not -1"
+            }
+        }
+    }
+}
+
+/**
+ * Serves as cross layer (Kotlin, Rust) communication class.
+ */
+@Keep
```
dannywillems, line 32: Are these supposed to match an equivalent in the codebase of librustzcash?

backend-lib/src/main/rust/lib.rs
```diff
@@ -1119,7 +1166,10 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_getTotalT
                 &taddr,
                 target,
                 wallet::ConfirmationsPolicy::MIN,
-                TransparentOutputFilter::All,
+                CoinbaseFilter::AllTransparentOutputs,
+                // Balance display, not a spend selection — show the true total regardless of any
```
dannywillems, line 1197: Nit: always better to have a let-binding to the value, and add the comment above the let binding, instead than at the call site.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1359,7 +1409,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_rewindToH
         let mut db_data = wallet_db(env, network, db_data)?;
 
         let height = BlockHeight::try_from(height)?;
-        let rewind_result = db_data.rewind_to_height(height);
+        let rewind_result = db_data.truncate_to_height(height);
```
dannywillems, line 1439: API renaming. LGTM.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1470,6 +1522,17 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_putSubtre
             orchard::tree::MerkleHashOrchard::read(n)
         })?;
 
+        // Ironwood is Orchard note-version V3 and shares Orchard's commitment-tree machinery, so
```
dannywillems, line 1552: LGTM. Copy from above.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1965,6 +2042,10 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_putUtxo<'
                 script_pubkey,
             },
             Some(BlockHeight::from(height as u32)),
+            // This JNI entry point has no account context to attribute the UTXO to.
```
dannywillems, line 2072: Nit: I would add let-binding to these None, so a reader can see what is the semantic of these values.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2094,6 +2175,8 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeTr
             &change_strategy,
             request,
             wallet::ConfirmationsPolicy::default(),
+            &SpendPolicy::default(),
+            None,
```
dannywillems, line 2206: Nit: I would add let-binding to these None, so a reader can see what is the semantic of these values.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2156,6 +2239,8 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeTr
...
```
dannywillems, line 2270: Nit: I would add let-binding to these None, so a reader can see what is the semantic of these values.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2291,7 +2376,8 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeSh
             &from_addrs,
             account_uuid,
             confirmations_policy,
-            TransparentOutputFilter::All,
+            CoinbaseFilter::AllTransparentOutputs,
+            None,
```
dannywillems, line 2407: Nit: I would add let-binding to these None, so a reader can see what is the semantic of these values.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2389,10 +2475,12 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_createPcz
                 account_id,
                 OvkPolicy::Sender,
                 &proposal,
+                None,
```
dannywillems, line 2505: Nit: I would add let-binding to these None, so a reader can see what is the semantic of these values.

backend-lib/src/main/rust/lib.rs
```diff
@@ -3671,6 +3791,9 @@ fn parse_protocol(code: i32) -> anyhow::Result<ShieldedProtocol> {
     match code {
         2 => Ok(ShieldedProtocol::Sapling),
         3 => Ok(ShieldedProtocol::Orchard),
+        // Must match ZcashProtocol.kt's poolCode and zcash_client_sqlite's own pool-type
```
dannywillems, line 3841: Unecessary comment?

backend-lib/src/main/rust/lib.rs
```diff
@@ -2486,15 +2574,47 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_addProofs
...
+                .ok_or_else(|| {
+                    anyhow!("PCZT's consensus branch does not support the Orchard pool")
+                })?
+                .circuit_version();
+            // Process-wide cached proving key (`zcash_primitives`, adopted 2026-07-23) — building
+            // one is expensive, and this JNI entry point is called on every ordinary shielded
+            // send, not just migrations, so rebuilding it per call was wasted work on every
+            // proof. Still keyed on the branch-derived circuit version above, since this block
+            // also has to prove legacy (pre-Ironwood) Orchard PCZTs during the migration window.
+            let pk = zcash_primitives::transaction::builder::cached_orchard_proving_key(
```
dannywillems, line 2639: Nit: Could be shortened to something like `builder::cached_orchard_proving_key`.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2486,15 +2574,47 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_addProofs
...
+        // consensus branch this PCZT was created for; derive it from the PCZT's branch id before
+        // the PCZT is consumed by the prover. The Ironwood bundle always uses PostNu6_3.
+        let pczt_branch_id = BranchId::try_from(*pczt.global().consensus_branch_id())
+            .map_err(|_| anyhow!("PCZT has an invalid consensus branch id"))?;
+
         let mut prover = Prover::new(pczt);
 
         if prover.requires_orchard_proof() {
+            let orchard_circuit_version =
+                zcash_primitives::transaction::components::orchard::bundle_version_for_branch(
```
dannywillems, line 2626: Could be shortened into something like `orchard::bundle_version_for_branch`.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2486,15 +2574,47 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_addProofs
...
+                    pczt_branch_id,
+                    orchard::ValuePool::Orchard,
+                )
+                .ok_or_else(|| {
+                    anyhow!("PCZT's consensus branch does not support the Orchard pool")
+                })?
+                .circuit_version();
+            // Process-wide cached proving key (`zcash_primitives`, adopted 2026-07-23) — building
```
dannywillems, line 2634: Is it necessary as a comment?

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
+//! JNI bindings for the migration engine.
+//!
+//! Rewired (2026-07-21) from our own hand-rolled `zcash_pool_migration` crate onto the core/
```
dannywillems, line 3: I think we should remove these comments. That's what should be in the git history.

backend-lib/src/main/rust/lib.rs
```diff
@@ -3810,3 +3933,52 @@ fn parse_http_headers(
         })
         .collect()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn map_import_account_error_tags_checkpoint_parity_failures() {
```
dannywillems, line 3989: Is this kind of test we care in the SDK?

backend-lib/Cargo.toml
```diff
@@ -11,33 +11,63 @@ edition = "2024"
...
+# Enables the shielded-voting JNI surface (src/main/rust/voting.rs and its
+# `Java_..._VotingRustBackend_*` exports) and pulls in the optional `zcash_voting` dependency
+# below.
+chp-voting = ["dep:zcash_voting", "dep:pasta_curves", "dep:ff"]
 
 [dependencies]
 # Zcash dependencies
-orchard = { version = "=0.14.0", features = ["unstable-voting-circuits"] }
-pczt = { version = "0.7", features = ["prover", "orchard", "sapling"] }
+# Ironwood (NU6.3): tracks the same git rev as [patch.crates-io] below directly, since its
```
dannywillems: We should remove this comment. We also can use [0.15.4](https://github.com/zcash/orchard/releases/tag/0.15.4)

backend-lib/Cargo.toml
```diff
@@ -11,33 +11,63 @@ edition = "2024"
...
-zcash_primitives = "0.28"
-zcash_proofs = "0.28"
-zcash_protocol = "0.9"
+zcash_primitives = "0.29"
+zcash_proofs = "0.29"
+zcash_protocol = "0.10"
 zcash_script = "0.4"
 zip32 = "0.2"
 eip681 = "0.1.0"
+# Used by migration_finalize.rs to read `ShardTreeError`/`QueryError` variants when checking
```
dannywillems, line 59: This comment could be removed. It is an implementation detail.

backend-lib/Cargo.toml
```diff
@@ -11,33 +11,63 @@ edition = "2024"
...
+# note-witness availability at a migration transaction's anchor boundary.
+shardtree = "0.7"
 # Used by voting/notes.rs to construct historical Orchard witness positions.
```
dannywillems, line 62: This could be removed. It is an implementation detail, and can be seen in the code using `git grep`.

backend-lib/Cargo.toml
```diff
@@ -91,48 +121,127 @@ xz2 = { version = "0.1", features = ["static"] }
...
+zcash_encoding = { git = "https://github.com/zcash/librustzcash", rev = "4b4e2633b28c2f18fdc42c741ad7c6d9f67ec6e5" }
+zcash_transparent = { git = "https://github.com/zcash/librustzcash", rev = "4b4e2633b28c2f18fdc42c741ad7c6d9f67ec6e5" }
+zip321 = { git = "https://github.com/zcash/librustzcash", rev = "4b4e2633b28c2f18fdc42c741ad7c6d9f67ec6e5" }
+pczt = { git = "https://github.com/zcash/librustzcash", rev = "4b4e2633b28c2f18fdc42c741ad7c6d9f67ec6e5" }
+equihash = { git = "https://github.com/zcash/librustzcash", rev = "4b4e2633b28c2f18fdc42c741ad7c6d9f67ec6e5" }
+zcash_voting = { path = "../../../../../zcash_voting/zcash_voting" }
+orchard = { git = "https://github.com/zcash/orchard.git", rev = "3bd2b736d1273226595773265313fc3f3428af16" }
+
+[dev-dependencies]
+# Test-only: converts a BIP-39 mnemonic phrase to seed bytes for `migration::live_wallet_tests`,
```
dannywillems, line 185: This comment could be removed. It is an implementation detail.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/fixture/SingleCompactBlockFixture.kt
```diff
@@ -11,6 +11,7 @@ internal object SingleCompactBlockFixture {
     internal const val DEFAULT_TIME = 0
     internal const val DEFAULT_SAPLING_OUTPUT_COUNT = 1u
     internal const val DEFAULT_ORCHARD_OUTPUT_COUNT = 2u
+    internal const val DEFAULT_IRONWOOD_OUTPUT_COUNT = 0u
```
dannywillems, line 14: Why `0`?

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+/// `schemerz` never re-runs a migration whose id it already recorded as applied, and the DDL is
+/// `CREATE TABLE IF NOT EXISTS` (a no-op on an existing table), a wallet that already ran this
+/// migration under an OLDER shape keeps that older shape forever and crashes with "no such
+/// column" the moment newer code queries the missing one (confirmed live twice now: `account_id`,
+/// then `lock_owner`).
+///
+/// Rather than requiring a full wallet wipe on every such churn, patch the one column difference
+/// we've hit directly: idempotent (checks before altering), and a no-op if the table doesn't
+/// exist yet (a wallet that has never attempted a migration) or already has the column.
+fn ensure_migration_schema_current(conn: &Connection) -> anyhow::Result<()> {
```
dannywillems: We should move this kind of check upstream, in the wallet codebase, if it is useful.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+        .prepare(
+            "SELECT 1 FROM pragma_table_info('orchard_ironwood_migration_transactions') \
+             WHERE name = 'lock_owner'",
+        )
+        .map_err(|e| anyhow!("Error checking migration schema: {}", e))?
+        .exists([])
+        .map_err(|e| anyhow!("Error checking migration schema: {}", e))?;
+    if !has_lock_owner {
+        conn.execute_batch(
+            "ALTER TABLE orchard_ironwood_migration_transactions ADD COLUMN lock_owner BLOB",
```
dannywillems: We should never have this kind of code here. This should be upstream.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+    let network = crate::parse_network(network_id as u32)?;
+    let db_path = crate::path_from_jni(env, db_data)?;
+    let (wallet, store_conn) = open_at(&db_path, network)?;
+    Ok((network, wallet, store_conn))
+}
+
+/// The height migration transactions are built/planned against — one past the current tip,
+/// matching the old crate's convention (and `migration_engine::Backend`'s own note-selection
+/// target).
+fn target_height(wallet: &Wallet) -> anyhow::Result<BlockHeight> {
```
dannywillems, line 223: This should go upstream, or use an existing method. I guess here it is mostly to get an `anyhow::Result` type.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+        .ok_or_else(|| anyhow!("wallet has no chain tip yet"))?;
+    Ok(tip + 1)
+}
+
+/// The wallet's real, currently-witnessable anchor height (the same one ordinary, non-migration
+/// sends use, via `get_target_and_anchor_heights`) — NOT just "chain tip minus one", which isn't
+/// necessarily checkpointed (confirmed live: `root_at_checkpoint_id` returned `None` for a raw
+/// `tip - 1` guess). Used as the anchor for preparation transactions, whose `anchor_boundary()` is
+/// `None` (see `try_prove`'s doc comment).
+fn natural_anchor_height(wallet: &Wallet) -> anyhow::Result<BlockHeight> {
```
dannywillems, line 236: Same as target_height above.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+
+/// Computes a fresh preview plan and caches it (see `migration_plan_cache`'s module doc for why)
+/// so a later commit call signs exactly this plan, not an independently re-randomized one. Also
+/// returns the wallet's current tip, needed as the "now" reference point when encoding transfer
+/// proposals (see `encode_transfer_proposal`'s doc comment for why this matters).
+///
+/// JNI-free — see `open_at`'s doc comment. Every `MIGRATION_DIAG` log line this crate has needed
+/// so far to diagnose a live bug (anchor/witness resolution, schedule spread, note-split
+/// detection) came from here or `migration_finalize`, both callable directly from `cargo test`.
+fn plan_for(
```
dannywillems: This seems to be a wrapper/rewritten function, for debugging and conversion to `anyhow::Result` purpose. Is it expected? Is using `tracing::debug` a convention?

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+                    .collect();
+                return Ok((state, unsigned));
+            }
+        }
+    }
+    let migration_plan = crate::migration_plan_cache::get(account).ok_or_else(|| {
+        anyhow!("No pending migration proposal — call propose/prepare first")
+    })?;
+    let mut backend = Backend::new(wallet, account, usk, store_conn)?;
+    let mut rng = OsRng;
```
dannywillems, line 388: Important: is it the CSRNG we use everywhere in the codebase?

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+}
+
+/// Whether transaction `tx` is ready to PROVE at `target_height` (`chain_tip + 1`) — a local copy
+/// of `zcash_pool_migration_backend::state`'s private `MigrationState::prove_ready`, using only its
+/// public surface (`deps_mined`, `anchor_boundary`, `scheduled_height`). Duplicated rather than
+/// relying on `MigrationState::next_provable` because that returns only the SINGLE next-ready
+/// transaction — looping it would re-return the same id forever on a transient witness/anchor
+/// failure (see `try_prove`'s doc comment), whereas our JNI contract proves every ready transaction
+/// in one call.
+fn is_prove_ready(state: &MigrationState, tx: &engine::MigrationTransaction, target_height: BlockHeight) -> bool {
```
dannywillems: We should make the method public upstream.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+/// boundary and proves against the wallet's current natural anchor instead, matching
+/// `zcash_pool_migration_backend`'s own `prove_chain_sim.rs` integration test.
+///
+/// Returns `Ok(true)` if proved (`state` now has this transaction `Proved`, with the proven PCZT
+/// replacing the stored one), `Ok(false)` if its witness/anchor isn't resolvable yet — the funding
+/// note hasn't been observed as spendable yet, or its checkpoint hasn't been reached or was pruned
+/// (`WalletProveError::UnknownSpentNote`/`AnchorNotFound`/`WitnessNotFound`) — this is the ordinary
+/// transient "not ready yet" condition, not a failure, matching the old stopgap's `Ok(None)`
+/// contract. Any other error is propagated.
+fn try_prove(
```
dannywillems, line 789: Review for me: this is a critical part.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+                    JValue::Object(&pczt_obj),
+                ],
+            )?
+            .into_raw())
+    });
+    unwrap_exc_or(&mut env, res, ptr::null_mut())
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_MigrationRustBackend_extractBroadcastTxNative<
```
dannywillems, line 957: Nit: we should move JNI in their own module, so we split this file in logical pieces. I would say we use something like `rust::jni::migration`.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+    });
+    unwrap_exc_or(&mut env, res, ptr::null_mut())
+}
+
+/// A fixed, well-known lock owner for the "Lock balance" dust-lock feature
+/// (`MigrationSdk.lockRemainingOrchardBalance`) — not a per-proposal lock, so a stable constant
+/// (not `LockOwner::random`) lets re-invoking the feature re-extend the same lock idempotently
+/// (see `WalletWrite::lock_outputs`'s doc comment on same-owner re-locking) and would let a future
+/// "undo" flow release it via this same token.
+const DUST_LOCK_OWNER: LockOwner = LockOwner::new(*b"zashi-migration-dust-lock-owner!");
```
dannywillems, line 1553: Note to me: carefully review this.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+            crate::utils::rust_vec_to_java(env, uuid_bytes, "[B", |env, bytes| {
+                crate::utils::rust_bytes_to_java(env, &bytes)
+            })?
+            .into_raw(),
+        )
+    });
+    unwrap_exc_or(&mut env, res, ptr::null_mut())
+}
+
+/// DEBUG ONLY: wipes this account's in-progress migration entirely (every preparation and
```
dannywillems: If it is debug only, we should have a submodule `debug` where we move these methods.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+                 WHERE account_id = (SELECT id FROM accounts WHERE uuid = ?1)",
+                rusqlite::params![account.expose_uuid()],
+            )
+            .map_err(|e| anyhow!("Error clearing migration: {}", e))?;
+        Ok(deleted as jint)
+    });
+    unwrap_exc_or(&mut env, res, 0)
+}
+
+/// DEBUG ONLY: overrides this account's persisted migration schedule so its transfers become due
```
dannywillems, line 1702: If it is debug only, we should have a submodule debug where we move these methods.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+    });
+    unwrap_exc_or(&mut env, res, ())
+}
+
+// ----- Keystone batch-signing UR bridge (crate::migration_keystone) -----
+//
+// Pure PCZT/UR operations over caller-held bytes — no wallet database, no migration engine.
+// Unaffected by this rewire.
+
+fn decode_byte_array_list(env: &mut JNIEnv, list: &JObjectArray) -> anyhow::Result<Vec<Vec<u8>>> {
```
dannywillems, line 2228: This is a util method, for the JNI, and can be applied generically.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+        wallet
+            .get_account_ids()
+            .expect("list accounts")
+            .into_iter()
+            .next()
+            .expect("wallet has at least one account — restore/sync one first")
+    }
+
+    #[test]
+    #[ignore = "requires MIGRATION_TEST_WALLET_DB pointing at a copy of a real wallet DB"]
```
dannywillems, line 2415: ~~Test ignored. Do we want to keep it then?~~ Just read doc above

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+/// proves. Neither does any network I/O — nothing here is ever broadcast or submitted anywhere.
+/// Needs a real `UnifiedSpendingKey`, provided via `MIGRATION_TEST_SEED_PHRASE` (a BIP-39 mnemonic,
+/// account 0) — never logged or persisted by this test.
+#[cfg(test)]
+mod live_wallet_signing_tests {
+    use super::*;
+    use zcash_client_backend::data_api::Account;
+
+    #[test]
+    #[ignore = "requires MIGRATION_TEST_WALLET_DB and MIGRATION_TEST_SEED_PHRASE"]
```
dannywillems, line 2471: ~~Test ignored. Do we want to keep it then?~~ Just read doc above

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+        println!("{finalized} finalized, {transient} transient — nothing broadcast");
+    }
+
+    /// Proves IMMEDIATE mode's proposal is an ordinary send-max, not the shuffled N-transfer
+    /// engine plan AUTOMATIC mode commits: a single step, drawing only Orchard-pool inputs (no
+    /// transparent, no Sapling). Read-only (a proposal, never committed/signed), so — unlike
+    /// `commit_and_finalize_with_real_signing` above — this needs no `MIGRATION_TEST_SEED_PHRASE`
+    /// and never touches persisted `MigrationState`.
+    #[test]
+    #[ignore = "requires MIGRATION_TEST_WALLET_DB"]
```
dannywillems, line 2566: ~~Test ignored. Do we want to keep it then?~~ Just read doc above

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+    /// `storeSignedNoteSplitPcztNative` applied the external signature but never resolved the
+    /// split's deferred witness/anchor before returning the PCZT for extraction. This exercises
+    /// the same shape as that JNI function (`build_preparation_unsigned` -> sign the split
+    /// out-of-process, matching what handing a redacted PCZT to Keystone and getting it back
+    /// signed looks like -> `apply_signature` -> `finalize_note_split`) via its shared, JNI-free
+    /// `finalize_note_split` helper, then extracts the result exactly like
+    /// `extractBroadcastTxNative` does, to prove the crash is actually fixed end to end, not just
+    /// that `finalize_note_split` returns `Ok`.
+    #[test]
+    #[ignore = "requires MIGRATION_TEST_WALLET_DB and MIGRATION_TEST_SEED_PHRASE"]
```
dannywillems, line 2624: ~~Test ignored. Do we want to keep it then?~~ Just read doc above

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,2987 @@
...
+    /// Demonstrates the SINGLETON_ID cross-account collision directly (see
+    /// `project_core_migration_swap` memory / spec doc §6.3): `pool_migrations`/
+    /// `pool_migration_transactions` have no `account_id` column, and `Backend::get_migration`/
+    /// `replace_migration` (`migration_engine.rs`) pass straight through to the store without
+    /// filtering by `self.account` — confirmed directly in that impl, not assumed. This is a bug
+    /// in OUR OWN adapter, not just a documented upstream limitation: any JNI call for account B
+    /// (`migrationStateNative`, `commit_or_reuse`, ...) reads/writes account A's committed
+    /// migration whenever one exists in the same wallet DB.
+    #[test]
+    #[ignore = "requires MIGRATION_TEST_WALLET_DB"]
```
dannywillems, line 2819: ~~Test ignored. Do we want to keep it then?~~ Just read doc above

backend-lib/src/main/rust/migration_engine.rs
```diff
@@ -0,0 +1,286 @@
...
+    PoolMigrationRead, PoolMigrationWrite,
+};
+use zcash_client_sqlite::pool_migration::orchard_ironwood::PoolMigrations;
+
+use crate::migration::Wallet;
+
+type SpendableNote = (OrchardNote, Position, u64);
+
+/// The migration adapter's `Backend`/`MigrationCrypto`/`PoolMigrationRead`/`PoolMigrationWrite`
+/// error type. Everything is folded into `anyhow::Error` (matching the rest of this JNI glue's
```
dannywillems, line 48: > `anyhow::Error` (matching the rest of this JNI glue's idiom)

Ah that's the reason!

backend-lib/src/main/rust/migration_engine.rs
```diff
@@ -0,0 +1,286 @@
...
+/// The destination is the account's own internal Ironwood receiver. Ironwood shares the Orchard
+/// receiver encoding end to end (confirmed in `zcash_keys::address::Address::can_receive_as`:
+/// `PoolType::Shielded(ShieldedPool::Orchard | ShieldedPool::Ironwood)` both match an Orchard
+/// receiver) — there is no separate "Ironwood address" type, so deriving
+/// `orchard_fvk.address_at(0u32, Scope::Internal)` and wrapping it as `Receiver::Orchard` before
+/// encoding to a `ZcashAddress` is both correct and exactly how
+/// `zcash_pool_migration_backend::build::build_transfer_pczt` derives a migration transfer's own
+/// crossing destination (its `recipient = orchard_fvk.address_at(0u32, Scope::Internal)`) — this
+/// reuses that same derivation, not a second one.
+pub fn propose_immediate_send_max(
```
dannywillems, line 267: This kind of "pure" method should go upstream.

nuttycom:
## Review follow-up checklist

Consolidated from a deep review of this PR's diff against `prerelease/snapshot-v2.6.6` (as of `64aa849d`; file:line references are to that revision). Checked items are already addressed on this branch.

### Funds- / privacy-critical

- [x] **Approve-X-sign-Y commit hazard** — commit calls ignored the caller-echoed schedule and signed whatever plan the process-global cache last held; read-only-looking calls silently replaced that plan. *Fixed in 79aca498: plans are cached Rust-side under an opaque `PlanHandle`; commits pass only the handle and error on Missing/Superseded.*
- [ ] **Permanent `InvalidNote` on any non-gRPC submit rejection** (`OrchardMigrationSdkImpl.kt:630-644`) — process death between broadcast and `recordTransferResult` means the resubmit's "already have transaction" rejection permanently invalidates a *successful* transfer and pushes the user into `restartCurrentMigrationStep` (double-spend hazard vs the in-mempool tx). Transient mempool/fee rejections likewise made permanent.
- [ ] **Clearnet leak on import with Tor enabled** (`SlipstreamSynchronizer.kt:295-304`) — `importAccountByUfvk` calls `restoreAnchor(torDir = null)` even when Tor is on (cold-start provisioning correctly passes `engineTorDir`), linking the user's IP to their wallet at import.
- [ ] **Expired-transfer recovery unreachable end-to-end** (`migration.rs:1011` + `:431-441`) — Expired tag is a no-op, nothing calls upstream `rebuild_expired_transfer`, and no production path sets `MigrationStatus::Failed`; a device offline past all broadcast windows is permanently stuck `InProgress` with `hasInvalidTransfers` false forever.
- [ ] **`MIGRATION_DIAG` logs defeat ZIP 318 privacy** (`migration.rs:283-330, 1399-1414`) — per-note zatoshi values plus every transfer's broadcast/expiry height reconstruct the randomized schedule the delay-drawing exists to hide (debug builds/logcat/bugreports).

### Major correctness

- [ ] **Seconds-vs-height unit confusion in `rescheduleOverdueTransfer`** (`OrchardMigrationSdkImpl.kt:450-456`) — `minOf(nowSeconds + interval, expiryHeight - 1)` always yields `expiryHeight - 1`; the reschedule-one-cadence-later branch is dead code.
- [ ] **Unguarded `isSyncBlocked()` flow can kill sync permanently** (`OrchardMigrationSdkImpl.kt:414-426, 506-525`) — one exception (e.g. DB file not yet created, transient SQLite lock) terminates `WalletCoordinator`'s `combine→stateIn` pipeline until process restart.
- [ ] **Voting: multi-bundle delegation wedges** (`voting/delegation.rs:147`) — `update_round_phase_forward(HotkeyGenerated)` errors with "refusing to regress round phase" once bundle 0 advances the round, so bundle ≥1 (or any rebuild) permanently fails.
- [ ] **`chp-voting` feature never enabled by the Gradle build** (`backend-lib/build.gradle.kts:133-140` vs `lib.rs:111`) — every voting native symbol is absent from the shipped `.so` (`UnsatisfiedLinkError`), and the rewritten `VotingRustBackendTest` suite can't pass as configured. If intentional for this cycle, document it.
- [ ] **Ironwood proving key rebuilt on every proof** (`lib.rs:2619-2626`) — use `cached_orchard_proving_key(PostNu6_3)` like the adjacent Orchard branch; currently a multi-second halo2 key build per transaction on mobile.
- [ ] **Slipstream: Pending→Confirmed never re-evaluated** (`TransactionsController.kt:29-43`, `PollGate.kt:21-28`) — chain-tip advance doesn't trigger a re-query, so a mined tx in an idle wallet shows Pending forever.
- [ ] **Slipstream lifecycle gaps** (`SlipstreamSynchronizer.kt`) — rewind/`deleteAccount` skip the engine-restart invariant and the quiescence check before destructive writes (:559-582, :722-729); construction failure after `engine.open()` leaks the native handle while releasing the instance guard (:974-1049, WAL-corruption scenario); `erase` lacks shutdown-await/mutex (:1104-1118); `onForeground`/`close` race can crash the process (:686-694).
- [ ] **`remainingOrchardValueZatoshi` is a disguised constant 0** (`migration.rs:453-457, 1110`), and `next_transfer_ready_at_height` never reports the actual future scheduled height (:459-461) — progress UI shows 0 remaining and no ETA.
- [ ] **Slipstream host reads decode nullable columns non-optionally** (`host_read.rs:274-295` `getTransactionRaw`; `:430-451` `listResubmissionCandidates`) — one un-enhanced/unmined row throws; the resubmission case disables resubmission for *all* candidates.
- [ ] **`recordTransferResult` semantics don't exist** (`migration.rs:975-1017`) — tags 1/2/3 are no-ops, `retryable` is ignored (KDoc promises otherwise), and tag 0 can regress `Mined`→`Broadcast` on a stale callback.

### Supply chain / resources

- [ ] **Pin `ur-registry` by rev** (`Cargo.toml:187` — floats on the default branch of KeystoneHQ/keystone-sdk-rust, inside the hardware-wallet signing path); `ur` pinned only by mutable tag (:186).
- [ ] **Migration Tor runtime never disposed** (`OrchardMigrationSdkImpl.kt:99-121`) — per-instance `RustBackend` + `TorClient` with no dispose API and a fresh-instance-per-call-site doc; repeated Tor broadcasts accumulate live runtimes on one data dir. Related: `LazyTorClient.ifCreated` invokes the action outside the mutex, racing `dispose()` (`LazyTorClient.kt:45-55`).
- [ ] **String-typed error discrimination at the FFI** — retryable-import marker string-matched on both sides with no pinning test (`lib.rs:504-536` ↔ `ImportAccountErrors.kt:20-22`); `InsufficientFunds` retry via `e.message?.contains(...)` (`OrchardMigrationSdkImpl.kt:161-176`); upstream typed errors (`StalePlan` etc.) flattened to `{:?}` strings.

### API surface / packaging

- [ ] **Source-incompatible public changes need release-note treatment** — `AccountBalance` gains `ironwood` mid-parameter-list; `Synchronizer` gains abstract suspend members; `WalletCoordinator` gains required `isSyncBlocked` param.
- [ ] **Shrink the `com.zodl.slipstream` public surface** — everything except the `SlipstreamSynchronizer` facade can be `internal` (JNI reflection ignores Kotlin visibility; `@Keep` + proguard already protect carriers); `SlipstreamWalletDb` currently exposes free-form SQL over the wallet DB publicly. Document the fork-only packaging.
- [ ] **`MigrationSdk` raw `Long`s → `Zatoshi`/`BlockHeight` value types** (this gap is what let the seconds/height bug compile); `ByteArray`-holding data classes lack `equals`/`hashCode`.

### Tests / cleanups

- [ ] `OrchardMigrationSdkImplTest` has exactly one test — cover `mapSubmitResult`, `rescheduleOverdueTransfer`, `isSyncBlockedNow`, `loggedRetryLoop`, the null-account matrix.
- [ ] `singleton_id_collision_between_accounts` (`migration.rs`) asserts a bug the pinned upstream has fixed — it now spuriously fails and its doc misinforms.
- [ ] Pre-existing broken test `utils::exception::tests::unknown_any` — `panic!("{}", 1)` always produces a String payload, so the asserted "Unknown error occurred" branch is unreachable.
- [ ] `summary.rs:379-381` opens the wallet DB read-write (with CREATE) contrary to the module's read-only rule; use `open_read_only`.
- [ ] Slipstream JNI silently coerces invalid `network_id`/negative heights (`slipstream/mod.rs:300-305, 667, 841-847`) — fail fast instead.
- [ ] Voting positive-path coverage regression — delegation-submission and vote-recovery round trips now only assert that calls throw.
- [ ] `SubmitPlanStore`: no version field, unbounded growth, write-only (no production reader), survives `erase`.
- [ ] `EncryptedPreferenceKeys` sync-resume key: KDoc says millis but value is seconds; not namespaced per wallet/alias; no version tag.
- [ ] Proto Ironwood field numbers (`ironwoodActions = 9`, `ironwoodTree = 7`, `ironwood = 2`) — link the lightwalletd commit that assigns them; `CompactBlockUnsafe.ironwoodOutputsCount` is a self-described temporary diagnostic in a public class.
- [ ] Stale docs: slipstream "never built via cargo-ndk / see README" block (no README exists; if descriptors are hand-derived, a `javap` pass should gate merge); obsolete "Cargo cycle" rationale for the duplicated raw SQL in `slipstream/mod.rs:377-387`; stale `Cargo.toml` comments (migration_finalize.rs, pczt 0.7.0, zcash_voting 0.5.9).
- [ ] `gradle.properties`: verify `org.gradle.tooling.parallel` is a real property (the standard flag is `org.gradle.parallel`).

---
*Filed with Claude Code assistance.*

nuttycom:
Closing in favor of https://github.com/zcash/zcash-android-wallet-sdk/pull/2049

PR #2025, opened 2026-07-24
refactor: move the JNI export surface into crate::jni (stage 1)

dannywillems:
Follow-up to review comment [r3640431929](https://github.com/zcash/zcash-android-wallet-sdk/pull/2021#discussion_r3640431929) on #2021. Targets that PR's branch, since it rewrites files #2021 also touches.

This is **stage 1 of 5**. It establishes the `crate::jni` module tree and moves the two non-feature-gated areas into it. Later stages move `lib.rs` (76 exports), `voting/` (60) and `slipstream/` (17).

## What and why

The comment asked for the JNI glue in `migration.rs` to move to its own module, suggesting `rust::jni::migration`. Two adjustments:

- That path is not reachable. The crate root is `src/main/rust/lib.rs`, so `rust` is a directory, not a module. The reachable spelling is `crate::jni::migration`.
- Applied to `migration.rs` alone it would have made the tree *less* consistent. `lib.rs` holds 76 exports to `migration.rs`'s 32, and the existing `voting.rs` / `slipstream/` modules use the opposite idiom (JNI at the module root, helpers pushed into submodules). Doing it repo-wide is what makes the suggestion an improvement.

## The boundary

The rule is whether a symbol *speaks JNI*, not whether it is an export. Exports, anything whose signature mentions `JNIEnv` or a `J*`/`j*` type, Java class descriptors, and the object encoders/decoders all move. Where a helper mixed both (`open`, `plan` decoded a `JString` and then did real work), it was split rather than moved wholesale, leaving the pure half behind.

Marshalling lives beside the exports it serves, not in a shared `helpers` module. `crate::utils` remains the home for genuinely domain-neutral helpers, which is where `decode_byte_array_list` went, closing [r3640449715](https://github.com/zcash/zcash-android-wallet-sdk/pull/2021#discussion_r3640449715).

This gives a checkable invariant, now documented in `AGENTS.md`: **outside `#[cfg(test)]`, a logic module must not reference the `jni` crate at all.**

## Result

| | before | after |
|---|---|---|
| `migration.rs` | 3214 | 1423 (logic only) |
| `eip681.rs` | 266 | 37 (logic only) |
| `jni/` | - | 2115 across 4 files |

`migration.rs`'s 858 lines of live-wallet tests stayed put and now exercise ordinary Rust.

## Verification

- `cargo check --lib --all-features` green, zero warnings. `--all-features` matters: `chp-voting` and `slipstream` are off by default, so a plain `cargo check` compiles only 110 of the 187 exports.
- **Exported symbol set unchanged: 187 before, 187 after, empty `nm -gU` diff.** Kotlin binds by symbol name and resolves lazily, so a rename would be a runtime crash rather than a compile error. Verified at each refactor commit independently, so both are safe to cherry-pick.
- `cargo fmt --check` and the rustdoc intra-doc link check clean on all touched files.

## Rebased onto `Remove dead MigrationProgress.remainingOrchardZatoshi field`

That commit modified `encode_migration_progress` and `derive_migration_state`, which this PR moves, so it conflicted. Resolution re-applies the field removal in the new locations: the `(IIJJ)V` -> `(IIJ)V` signature change and the dropped `wallet` / `account` parameters are preserved in `jni/migration/encode.rs`, with both call sites updated in `jni/migration.rs` and the two imports that became dead removed. Verified afterwards that the set of functions and constants across the three files still matches the base's `migration.rs`, with `decode_byte_array_list` the only intentional departure.

## Notes for reviewers

- Because the new module is named `jni`, a bare `jni::` path **inside lib.rs** now resolves to it rather than the crate; 11 paths there are qualified `::jni::`. Submodules are unaffected. Documented in `AGENTS.md`.
- Visibility widening is the real cost: 12 items in `crate::migration` and 2 in `crate::eip681` go private to `pub(crate)`, and 15 marshalling items are `pub(super)`, scoped to the migration exports only. Every one has a call site.
- Two pre-existing defects surfaced while verifying, both untouched here: `cargo fmt --check` already fails on `slipstream/mod.rs`, and `lib.rs` has three broken rustdoc intra-doc links, one referencing a zcash-swift-wallet-sdk symbol that does not exist in this crate.
- CI on this branch will also show the 75 detekt issues inherited from the base; this PR changes no Kotlin.

Commits are split so each is independently cherry-pickable.


nuttycom:
Changed to draft - this PR should be rebased after everything else lands because it's code motion we don't need to take on right now.

AGENTS.md
```diff
@@ -87,11 +87,117 @@ The project uses ticket-prefixed commit messages for tracked work, e.g.
 prefix is acceptable (`fix: ...`, `chore: ...`). Keep the first line under
 72 characters; include context in the body.
 
+## JNI module layout (Rust)
+
+All `Java_*` JNI exports live under `backend-lib/src/main/rust/jni/`, one
```
nesence-m, line 92: This section describes the stage-5 end state in the present tense — "All `Java_*` JNI exports live under `backend-lib/src/main/rust/jni/`", with a tree listing `jni/wallet.rs`, `jni/derivation.rs`, `jni/tor.rs`, `jni/slipstream.rs`, `jni/voting/` — but at this commit only `jni/eip681.rs` and `jni/migration.rs` exist; 76 exports are still in `lib.rs` and the voting/slipstream exports are still in their logic modules. Since this PR is now parked to rebase after everything else lands, the window in which the doc describes a layout that doesn't exist could be long, and AGENTS.md is exactly the file an agent or new contributor takes literally (e.g. it would look for `jni/wallet.rs` to add a wallet export). Suggest either wording the table as the target layout with an explicit "stages 2–5 pending; currently only migration and eip681 have moved" marker, or splitting the AGENTS.md commit out so it lands with the final stage. The hardcoded per-class counts have the same shelf-life problem, though "as of the split" at least dates them.

AGENTS.md
```diff
@@ -87,11 +87,117 @@ The project uses ticket-prefixed commit messages for tracked work, e.g.
...
+This yields a mechanically checkable invariant: **outside `#[cfg(test)]`, a
+logic module must not reference the `jni` crate at all.**
+
+```bash
+cd backend-lib/src/main/rust
+grep -n 'jni::\|JNIEnv\|JString\|JByteArray\|JObject\|jlong\|jint' \
+  migration.rs eip681.rs tor.rs
+```
+
+Any hit outside a test block means the boundary is in the wrong place.
```
nesence-m, line 158: "Any hit outside a test block means the boundary is in the wrong place" is already false on this PR's own output: the grep as written matches doc comments, and the moved files legitimately contain `[crate::jni::migration]` / `[crate::jni::eip681]` intra-doc links plus a prose mention of `JString` in `open_at`'s doc — three hits today, all fine. A checker that fires on the known-good state tends to get ignored (or worse, "fixed" by deleting the doc links). Suggest either scoping the command to code references (filter out comment lines) or rewording to "any hit in code, as opposed to doc comments, means the boundary is in the wrong place".

backend-lib/src/main/rust/jni/migration/encode.rs
```diff
@@ -0,0 +1,326 @@
...
+pub(super) const JNI_MIGRATION_TRANSFER_STATES: &str =
+    "cash/z/ecc/android/sdk/internal/model/migration/JniMigrationTransferStates";
+pub(super) const JNI_UNSIGNED_TRANSFER_PCZT: &str =
+    "cash/z/ecc/android/sdk/internal/model/migration/JniUnsignedTransferPczt";
+pub(super) const JNI_KEYSTONE_BATCH_DECODE_RESULT: &str =
+    "cash/z/ecc/android/sdk/internal/model/migration/JniKeystoneBatchDecodeResult";
+pub(super) const JNI_KEYSTONE_BATCH_SIGNED_PCZTS: &str =
+    "cash/z/ecc/android/sdk/internal/model/migration/JniKeystoneBatchSignedPczts";
+
+pub(super) fn open(
```
nesence-m, line 51: Placement nit: `open` and `plan` aren't marshalling — they decode two arguments and then orchestrate real work (wallet open, plan computation) — yet they live in `encode.rs`, which `jni/mod.rs` defines as the home of "Java class descriptors, object encoders, argument decoders". By the PR's own boundary table these are the "mixed" helpers, and their JNI half reads more like shared plumbing for the exports than like an encoder. Consider keeping them in `jni/migration.rs` beside the exports that call them (they're `pub(super)` anyway), so `encode.rs` stays exactly what its name and the module doc claim. Not blocking — just worth settling before stages 2–4 copy the pattern.

nesence-m (commented):
Reviewed as a draft, with the understanding (per the parking comment) that this rebases after the rest lands. Independently verified the core invariant: the `Java_*` set is identical between base and head — 187 exports (plus the same commented-out `setAlternateServers`), same names, `lib.rs` 76 / `jni/migration.rs` 32 / voting 60 / slipstream 17 / eip681 2. A function-by-function body comparison of everything moved out of `migration.rs` and `eip681.rs` (including the constants and `open`/`plan`) shows only visibility widening (`pub(crate)`/`pub(super)`) and rustfmt re-wrapping — no behavior change smuggled into the motion. `Cargo.toml`/`Cargo.lock` are untouched, and the direction is right: this keeps the JNI surface as modules inside the one `backend-lib` crate rather than re-growing a separate crate, which is exactly the shape the #2021 rework established. Feature gating is unaffected (both moved families were ungated in base and remain so; the `slipstream`/`chp-voting` gates are untouched).

Remaining points are small and mostly about the AGENTS.md documentation landing ahead of the code it describes — see inline comments.

One non-anchored note, pre-existing and non-blocking: the anchor-boundary doc comments that stayed in `migration.rs` (around `natural_anchor_height`) still describe the separate `slipstream-jni` crate and its `path = "slipstream-jni"` dependency, which the base branch has since folded into this crate (`backend-lib/Cargo.toml` even documents the removal). Since this PR is already a documentation-and-motion pass over that file, the eventual rebase would be a natural moment to update those references.

PR #2026, opened 2026-07-24
Add Ironwood (NU6.3) pool support

dannywillems:
Exposes the Ironwood shielded pool through the SDK.

Extracted from `feature/2.5.x_migration`, which carries Ironwood support
alongside the Orchard-to-Ironwood migration engine, voting, and slipstream.
**None of those are included here** — this is the Ironwood pool surface only.

## No dependency changes

`maint/v2.5.x` already carries the dependency bump (4843902c), and the
crates.io releases it pins already ship the full Ironwood API:

| crate | version | provides |
|---|---|---|
| `zcash_client_backend` | 0.24.0-rc.1 | `put_ironwood_subtree_roots`, `ironwood_balance`, `next_ironwood_subtree_index` |
| `zcash_client_sqlite` | 0.22.0-rc.1 | `put_ironwood_subtree_roots`, pool code `4` |
| `zcash_protocol` | 0.10 | `ShieldedPool::Ironwood` |

So `Cargo.toml` and `Cargo.lock` are untouched. In particular this drops the
source branch's `[patch.crates-io]` git-rev pin of librustzcash, along with
`zcash_pool_migration_backend`, `ur`, `ur-registry`, `shardtree`,
`incrementalmerkletree`, and `bip0039` — all of which exist on that branch to
serve migration and Keystone batch signing, not Ironwood.

## What changed

Ironwood is Orchard note-version V3: it shares Orchard's commitment-tree
machinery but is a distinct pool with its own note commitment tree, subtree
roots, and balance.

**Rust (`lib.rs`)**

- `putSubtreeRoots` takes an Ironwood start index and root array, parsed with
  the Orchard hash type, forwarded to `put_ironwood_subtree_roots`.
- `encode_account_balance` and `encode_wallet_summary` carry the Ironwood
  balance triple and `next_ironwood_subtree_index`; both JNI constructor
  signatures updated in lockstep with the Kotlin `Jni*` classes.
- `parse_protocol` maps pool code 4 to `ShieldedPool::Ironwood`.

**Kotlin**

- `ZcashProtocol.IRONWOOD` (pool code 4), `TransactionPool.IRONWOOD`,
  `ShieldedProtocolEnum.IRONWOOD`.
- `JniAccountBalance` / `AccountBalance` gain the Ironwood balance;
  `JniWalletSummary` / `WalletSummary` gain `nextIronwoodSubtreeIndex`.
- `Backend.putSubtreeRoots` and `TypesafeBackend.putSubtreeRoots` take the
  Ironwood start index and roots. `CompactBlockProcessor` fetches Ironwood
  subtree roots alongside Sapling and Orchard and passes them through
  `GetSubtreeRootsResult.SpendBeforeSync`. As with Orchard, an empty Ironwood
  root list is not treated as a spend-before-sync failure, since NU6.3
  activates after Sapling.

**Protobuf**

- `ChainMetadata.ironwoodCommitmentTreeSize`, `CompactTx.ironwoodActions`
  (field 9), `TreeState.ironwoodTree`, and `ShieldedProtocol.ironwood`.

Verified field-for-field against the canonical
[`zcash/lightwallet-protocol`](https://github.com/zcash/lightwallet-protocol/blob/main/walletrpc/compact_formats.proto)
(which `lightwalletd` now symlinks to) and against `zcash_client_backend`'s
generated bindings. Both type `ironwoodActions` as `repeated
CompactOrchardAction` at field 9, and neither defines a
`CompactIronwoodAction` — the compact wire encoding is unchanged from Orchard,
so reusing the message is upstream's decision, not a shortcut here.

## Commits

| commit | |
|---|---|
| `e6a10834` | Add Ironwood (NU6.3) pool support |
| `cf187eae` | AGENTS.md: document how to compile locally |
| `d0c88dcb` | Give the Ironwood fixtures distinct, non-zero values |
| `e7b761c8` | Prefix the per-pool counters in getOutputsCounts |
| `3e0a21d2` | Cite lightwalletd's bindings for the TreeState field numbers |
| `7c821bd1` | Cite zcash_client_sqlite's encoding for the pool codes |

### On the fixture values (`d0c88dcb`)

Worth a reviewer's attention, because it is not cosmetic.

`JniAccountBalance` is built from Rust by `encode_account_balance`, which
passes ten positional `jlong`s under a JVM descriptor, `([BJJJJJJJJJJ)V`, that
cannot tell them apart; the Kotlin side only checks each is non-negative. With
`JniAccountBalanceFixture` setting all ten to `0L`, **any permutation of the
ten fields passed every test using it** — including one reporting Orchard funds
as Ironwood, which matters because the two pools have different spend rules.
The values are now `1L`–`10L`.

The Ironwood fixture values inherited from the source branch were `0`
throughout, encoding "the server does not send these yet" rather than a chosen
value. `lightwalletd` disproves that premise: it parses ZIP 229 v6
transactions, exposes `Transaction.IronwoodActionsCount`, filters on
`PoolType_IRONWOOD`, and its own tests build non-empty `IronwoodActions` lists.
The specific replacement values (`3u`, and `(9, 6, 3)`) are arbitrary markers
following the existing Sapling/Orchard fixture convention, not upstream
constants.

## Deviations from the source branch

Two things needed authoring rather than a straight cherry-pick, because
`feature/2.5.x_migration` is not green on either count:

1. **Unit tests did not compile there.** `JniWalletSummaryTest` and
   `JniAccountBalanceFixture` were never threaded through with the new
   required constructor parameters.
2. **Detekt/ktlint fail there.** Five issues were suppressions and line
   wrapping. The sixth: a third copy of the ~50-line per-pool subtree-root
   fetch loop tripped `ThrowsCount`, so the three near-identical loops are
   extracted into one pool-parameterized local helper. The only visible
   behaviour change is the pool name in the log messages.

Also dropped as debug scaffolding rather than Ironwood support:
`RustBackend.rustLogging = Debug`, and the `IRONWOOD_DIAG` temporary logging
in `CompactBlockDownloader` (the branch marks it "safe to remove once
confirmed").

## Validation

Run locally against this branch:

- `cd backend-lib && cargo check --lib` — clean
- `cargo fmt --check` — clean
- `./scripts/ci-local.sh fast` (detekt + ktlint) — pass
- `./scripts/ci-local.sh quick` (compile + unit tests) — pass

Not run: `./scripts/ci-local.sh full` (androidTest on a Gradle managed device).

### Two caveats

**`cargo test` is red on this branch, for a pre-existing reason.**
`utils::exception::tests::unknown_any` fails on `maint/v2.5.x` itself; this
branch does not touch `exception.rs` (`git diff maint/v2.5.x..HEAD --
backend-lib/src/main/rust/utils/exception.rs` is empty). The test's helper
panics via `panic!("{}", val)`, which always yields a `String` payload, so the
`"Unknown error occurred"` fallback it claims to cover is unreachable.
#2016 fixes it and adds the `cargo test` CI job that would have caught it; the
clean ordering is to merge that first and rebase this.

**Slow CI jobs on this branch may be cancelled rather than run.** The workflow
sets `cancel-in-progress: true`, and `static_analysis_android_lint` needs
~25-28 minutes because it cross-compiles Rust for four ABIs before linting. On
a branch pushed to frequently, a green tick can mean the fast jobs passed and
the slow ones never finished. Worth leaving the branch idle before trusting a
full green.

### DIFF
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/ZcashProtocol.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/ZcashProtocol.kt
@@ -13,18 +13,22 @@ enum class ZcashProtocol {
     },
     ORCHARD {
         override val poolCode = 3
+    },
+    IRONWOOD {
+        override val poolCode = 4
     };
 
     abstract val poolCode: Int
 
-    fun isShielded() = this == SAPLING || this == ORCHARD
+    fun isShielded() = this == SAPLING || this == ORCHARD || this == IRONWOOD
 
     companion object {
         fun validate(poolTypeCode: Int): Boolean =
             when (poolTypeCode) {
                 TRANSPARENT.poolCode,
                 SAPLING.poolCode,
-                ORCHARD.poolCode -> true
+                ORCHARD.poolCode,
+                IRONWOOD.poolCode -> true
 
                 else -> false
             }
@@ -34,6 +38,7 @@ enum class ZcashProtocol {
                 TRANSPARENT.poolCode -> TRANSPARENT
                 SAPLING.poolCode -> SAPLING
                 ORCHARD.poolCode -> ORCHARD
+                IRONWOOD.poolCode -> IRONWOOD
                 else -> error("Unsupported pool type: $poolCode")
             }
     }
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -1430,6 +1430,8 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_putSubtre
     sapling_roots: JObjectArray<'local>,
     orchard_start_index: jlong,
     orchard_roots: JObjectArray<'local>,
+    ironwood_start_index: jlong,
+    ironwood_roots: JObjectArray<'local>,
     network_id: jint,
 ) {
     let res = catch_unwind(&mut env, |env| {
@@ -1470,6 +1472,15 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_putSubtre
             orchard::tree::MerkleHashOrchard::read(n)
         })?;
 
+        let ironwood_start_index = if ironwood_start_index >= 0 {
+            ironwood_start_index as u64
+        } else {
+            return Err(anyhow!("Ironwood start index must be nonnegative."));
+        };
+        let ironwood_roots = parse_roots(env, ironwood_roots, |n| {
+            orchard::tree::MerkleHashOrchard::read(n)
+        })?;
+
         db_data
             .put_sapling_subtree_roots(sapling_start_index, &sapling_roots)
             .map_err(|e| anyhow!("Error while storing Sapling subtree roots: {}", e))?;
@@ -1477,6 +1488,10 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_putSubtre
         db_data
             .put_orchard_subtree_roots(orchard_start_index, &orchard_roots)
             .map_err(|e| anyhow!("Error while storing Orchard subtree roots: {}", e))?;
+
+        db_data
+            .put_ironwood_subtree_roots(ironwood_start_index, &ironwood_roots)
+            .map_err(|e| anyhow!("Error while storing Ironwood subtree roots: {}", e))?;
 
         Ok(())
     });
@@ -1579,11 +1594,17 @@ fn encode_account_balance<'a>(
     let orchard_value_pending =
         ZatBalance::from(balance.orchard_balance().value_pending_spendability());
 
+    let ironwood_verified_balance = ZatBalance::from(balance.ironwood_balance().spendable_value());
+    let ironwood_change_pending =
+        ZatBalance::from(balance.ironwood_balance().change_pending_confirmation());
+    let ironwood_value_pending =
+        ZatBalance::from(balance.ironwood_balance().value_pending_spendability());
+
     let unshielded = ZatBalance::from(balance.unshielded_balance().total());
 
     env.new_object(
         JNI_ACCOUNT_BALANCE,
-        "([BJJJJJJJ)V",
+        "([BJJJJJJJJJJ)V",
         &[
             (&env.byte_array_from_slice(account_uuid.expose_uuid().as_bytes())?).into(),
             JValue::Long(sapling_verified_balance.into()),
@@ -1592,6 +1613,9 @@ fn encode_account_balance<'a>(
             JValue::Long(orchard_verified_balance.into()),
             JValue::Long(orchard_change_pending.into()),
             JValue::Long(orchard_value_pending.into()),
+            JValue::Long(ironwood_verified_balance.into()),
+            JValue::Long(ironwood_change_pending.into()),
+            JValue::Long(ironwood_value_pending.into()),
             JValue::Long(unshielded.into()),
         ],
     )
@@ -1637,7 +1661,7 @@ fn encode_wallet_summary<'a>(
     Ok(env.new_object(
         "cash/z/ecc/android/sdk/internal/model/JniWalletSummary",
         format!(
-            "([L{};JJJJLjava/lang/Long;Ljava/lang/Long;JJ)V",
+            "([L{};JJJJLjava/lang/Long;Ljava/lang/Long;JJJ)V",
             JNI_ACCOUNT_BALANCE
         ),
         &[
@@ -1650,6 +1674,7 @@ fn encode_wallet_summary<'a>(
             JValue::Object(&recovery_progress_denominator),
             JValue::Long(summary.next_sapling_subtree_index() as i64),
             JValue::Long(summary.next_orchard_subtree_index() as i64),
+            JValue::Long(summary.next_ironwood_subtree_index() as i64),
         ],
     )?)
 }
@@ -3707,6 +3732,7 @@ fn parse_protocol(code: i32) -> anyhow::Result<ShieldedPool> {
     match code {
         2 => Ok(ShieldedPool::Sapling),
         3 => Ok(ShieldedPool::Orchard),
+        4 => Ok(ShieldedPool::Ironwood),
         _ => Err(anyhow!("Shielded protocol not recognized: {code}")),
     }
 }
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -3729,6 +3729,8 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_model_TorWalletClient_fet
 //
 
 fn parse_protocol(code: i32) -> anyhow::Result<ShieldedPool> {
+    // The codes below must follow zcash_client_sqlite's own pool-type encoding:
+    // https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/wallet/encoding.rs#L42
     match code {
         2 => Ok(ShieldedPool::Sapling),
         3 => Ok(ShieldedPool::Orchard),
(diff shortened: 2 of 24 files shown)


backend-lib/src/main/rust/lib.rs
```diff
@@ -1579,11 +1596,17 @@ fn encode_account_balance<'a>(
     let orchard_value_pending =
         ZatBalance::from(balance.orchard_balance().value_pending_spendability());
 
+    let ironwood_verified_balance = ZatBalance::from(balance.ironwood_balance().spendable_value());
```
dannywillems, line 1597: Following Orchard/Sapling above. LGTM.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/ZcashProtocol.kt
```diff
@@ -13,18 +13,22 @@ enum class ZcashProtocol {
     },
     ORCHARD {
         override val poolCode = 3
+    },
+    IRONWOOD {
```
dannywillems, line 17: See https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/wallet/encoding.rs#L42

backend-lib/src/main/rust/lib.rs
```diff
@@ -3707,6 +3732,9 @@ fn parse_protocol(code: i32) -> anyhow::Result<ShieldedPool> {
     match code {
         2 => Ok(ShieldedPool::Sapling),
         3 => Ok(ShieldedPool::Orchard),
+        // Must match ZcashProtocol.kt's poolCode and zcash_client_sqlite's own pool-type
+        // encoding (wallet/encoding.rs) exactly.
+        4 => Ok(ShieldedPool::Ironwood),
```
dannywillems, line 3737: See https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/wallet/encoding.rs#L42

backend-lib/src/test/java/cash/z/ecc/android/sdk/internal/fixture/JniAccountBalanceFixture.kt
```diff
@@ -8,13 +8,22 @@ import cash.z.ecc.android.sdk.internal.model.JniAccountBalance
...
     val ACCOUNT_UUID: ByteArray = "random_uuid_16_b".toByteArray()
-    const val SAPLING_VERIFIED_BALANCE: Long = 0L
-    const val SAPLING_CHANGE_PENDING: Long = 0L
-    const val SAPLING_VALUE_PENDING: Long = 0L
-    const val ORCHARD_VERIFIED_BALANCE: Long = 0L
-    const val ORCHARD_CHANGE_PENDING: Long = 0L
-    const val ORCHARD_VALUE_PENDING: Long = 0L
-    const val UNSHIELDED_BALANCE: Long = 0L
+
+    // Every value is distinct on purpose. [JniAccountBalance] is constructed from Rust by
```
dannywillems, line 12: This is a test file.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/model/CompactBlockUnsafe.kt
```diff
@@ -28,25 +30,29 @@ class CompactBlockUnsafe(
                 time = compactBlock.time,
                 saplingOutputsCount = outputCounts.saplingOutputsCount,
                 orchardOutputsCount = outputCounts.orchardActionsCount,
+                ironwoodOutputsCount = outputCounts.ironwoodActionsCount,
                 compactBlockBytes = compactBlock.toByteArray()
             )
         }
 
         private fun getOutputsCounts(vtxList: List<CompactFormats.CompactTx>): CompactBlockOutputsCounts {
             var outputsCount: UInt = 0u
```
dannywillems: We should add orchard/sapling as a prefix.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/fixture/SingleCompactBlockFixture.kt
```diff
@@ -11,6 +11,7 @@ internal object SingleCompactBlockFixture {
     internal const val DEFAULT_TIME = 0
     internal const val DEFAULT_SAPLING_OUTPUT_COUNT = 1u
     internal const val DEFAULT_ORCHARD_OUTPUT_COUNT = 2u
+    internal const val DEFAULT_IRONWOOD_OUTPUT_COUNT = 3u
```
dannywillems, line 14: We're in the fixture files, so it is fine.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/model/CompactBlockUnsafe.kt
dannywillems: Done in e7b761c8.

Renamed the locals after the fields they populate, so all three carry their pool:

```kotlin
var saplingOutputsCount: UInt = 0u
var orchardActionsCount: UInt = 0u
var ironwoodActionsCount: UInt = 0u
```

I took `sapling` along with `orchard` since it is the same defect one line up, and leaving it would have made the block less consistent than before, not more.

Also switched the `CompactBlockOutputsCounts(...)` call to named arguments while there: all three values are `UInt`, so a positional call cannot catch a swapped argument. Happy to drop that part if you would rather keep the diff to the rename.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1470,6 +1472,15 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_putSubtre
             orchard::tree::MerkleHashOrchard::read(n)
         })?;
 
+        let ironwood_start_index = if ironwood_start_index >= 0 {
```
dannywillems, line 1475: Follow Orchard code above.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1478,6 +1489,10 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_putSubtre
             .put_orchard_subtree_roots(orchard_start_index, &orchard_roots)
             .map_err(|e| anyhow!("Error while storing Orchard subtree roots: {}", e))?;
 
+        db_data
```
dannywillems, line 1492: Follow Orchard code above.

backend-lib/src/main/rust/lib.rs
```diff
@@ -1579,11 +1594,17 @@ fn encode_account_balance<'a>(
...
+        ZatBalance::from(balance.ironwood_balance().change_pending_confirmation());
+    let ironwood_value_pending =
+        ZatBalance::from(balance.ironwood_balance().value_pending_spendability());
+
     let unshielded = ZatBalance::from(balance.unshielded_balance().total());
 
     env.new_object(
         JNI_ACCOUNT_BALANCE,
-        "([BJJJJJJJ)V",
+        "([BJJJJJJJJJJ)V",
```
dannywillems, line 1607: One thing I'd like to check later is if the JNI has a limit number of parameters. I know in OCaml, there is a limit for the bytecode because of old semantic from the 90s, and only 5 parameters can be given when writing C bindings. Spawning an agent for an audit about this.

nesence-m:
@zodl-review

nuttycom (approved):
utACK 7c821bd1238d34eb097b014beb7c35a34b04f0b6

lightwallet-client-lib/src/main/proto/compact_formats.proto
```diff
@@ -15,6 +15,7 @@ option swift_prefix = "";
 message ChainMetadata {
     uint32 saplingCommitmentTreeSize = 1; // the size of the Sapling note commitment tree as of the end of this block
     uint32 orchardCommitmentTreeSize = 2; // the size of the Orchard note commitment tree as of the end of this block
+    uint32 ironwoodCommitmentTreeSize = 3; // the size of the Ironwood note commitment tree as of the end of this block
```
pacu, line 18: Are these changes introduced manually? 

Can we actually use the https://github.com/zcash/lightwallet-protocol as intended or at least mark it as a TODO for later?

lightwallet-client-lib/src/main/proto/service.proto
pacu, line 1: same as above

pacu (approved):
LGTM

dannywillems:
Fixing [Pull Request / test_android_modules_wtf (pull_request)](https://github.com/zcash/zcash-android-wallet-sdk/actions/runs/30091553694/job/89475851268?pr=2026) in a follow-up, cherry-picking changes from `main`.

PR #2029, opened 2026-07-24
refactor: move the JNI export surface into crate::zcash_jni

dannywillems:
Moves the JNI export surface into a `crate::zcash_jni` module tree, so the logic modules stop carrying `extern "C"` functions.

This is an **independent implementation** of the same idea as #2025, not a port of it. That PR targets `android-slipstream-ironwood-chp`, where the file set is different: it carries `migration.rs` (32 exports), `voting/` (60) and `slipstream/` (17), none of which exist on `maint/v2.5.x`. Nothing was cherry-picked; the design rationale is shared, the code is not.

## Restructured after review

This branch was originally 7 commits, one of which moved 2512 lines. The
objection was fair: nothing let a reviewer confirm the code was *moved* rather
than edited, and four helpers had in fact been changed along the way, invisibly.

It is now 11 commits, and `check-jni-move.sh` (commit 2) exists to make the
"pure move" claim checkable. The rewrite changed **only the history**: the tree
at the tip is byte-identical to the pre-restructure tip, verified with
`git diff` between the two.

## How to review this (start here)

The branch is deliberately split so that **no commit asks you to trust that a large diff is a move**. Two scripts land first, before any code moves, so they are runnable at every later commit.

```bash
git checkout chore/jni-module-split-2.5.x
./scripts/check-jni-symbols.sh          # 78 exported names match the baseline
./scripts/check-jni-move.sh d65b4d2c    # all 78 export BODIES byte-identical to the base
./scripts/check-jni-move.sh d65b4d2c --diff   # ...and show any that are not
```

`check-jni-move.sh` splits every `Java_*` function out of the base revision and out of the working tree and compares them byte for byte. It exits non-zero if any body differs, so "this commit only moves code" is a claim you can check rather than one you have to trust.

Run it at **any** commit on the branch, not just the tip:

```bash
for c in $(git rev-list --reverse d65b4d2c..HEAD); do
  git checkout -q "$c"
  echo "$(git log -1 --format='%h %s' "$c")"
  ./scripts/check-jni-move.sh d65b4d2c | grep byte-identical
done
```

It reports `byte-identical: 78` at every commit from the second onward (the first is where `check-jni-move.sh` itself is added).

## Commit structure

| # | commit | what to look for |
|---|---|---|
| 1 | Add a regression check for the JNI export surface | `check-jni-symbols.sh` + baseline + CI job |
| 2 | Add a byte-equality check for moved JNI functions | `check-jni-move.sh` |
| 3 | refactor: establish `crate::zcash_jni`, move the EIP-681 exports | git reports it as a **rename** |
| 4 | refactor: move the derivation exports | pure move |
| 5 | refactor: move the Tor exports | pure move |
| 6 | refactor: move the wallet exports and shared marshalling | pure move (the big one) |
| 7 | helper: split `wallet_db` into a JNI half and a pure half | +18 / -4 |
| 8 | helper: split `block_db` into a JNI half and a pure half | +11 / -6 |
| 9 | helper: move `chain_id_string_to_u64` out of the JNI module | +19 / -13 |
| 10 | helper: move `hex_string_to_u256` out of the JNI module | +21 / -20 |
| 11 | AGENTS.md: document the JNI module layout and the symbol check | docs only, and the only commit touching AGENTS.md |

**Commits 3-6 are pure moves.** Every function they relocate, exported or not, keeps the body it had at the base commit `d65b4d2c`. In particular the four helpers this refactor genuinely changed are carried through those commits in their *original* form, and edited afterwards in commits 7-10, one helper per commit. That is why the 2512-line wallet move contains no edit to read: the edits are the four small commits after it.

Of the non-exported helpers involved, **all are byte-identical to `d65b4d2c` except four**, each of which has its own commit:

| helper | what changed | commit |
|---|---|---|
| `wallet_db` | split: pure `(path: PathBuf, params)` half in `lib.rs`, JNI half delegates | 7 |
| `block_db` | split: pure `(path: PathBuf)` half in `lib.rs`, JNI half delegates | 8 |
| `chain_id_string_to_u64` | relocated to `crate::eip681`, widened to `pub(crate)`; body unchanged | 9 |
| `hex_string_to_u256` | relocated to `crate::eip681`, widened to `pub(crate)`; body unchanged | 10 |

## Result

| file | before | after |
|---|---|---|
| `lib.rs` | 3812 (76 exports) | **169** (logic only) |
| `eip681.rs` | 266 (2 exports) | **36** (logic only) |
| `tor.rs` | 273 (0 exports) | 273 (untouched) |
| `utils.rs` + `utils/` | 386 (0 exports) | 386 (untouched) |
| `zcash_jni/` | - | **4069** across 5 files |

Exports per submodule, by Java class:

| submodule | class | exports |
|---|---|---|
| `zcash_jni/wallet.rs` | `RustBackend` | 50 |
| `zcash_jni/tor.rs` | `TorClient` + `TorWalletClient` | 8 + 9 |
| `zcash_jni/derivation.rs` | `RustDerivationTool` | 9 |
| `zcash_jni/eip681.rs` | `RustEip681Tool` | 2 |
| | | **78** |

Both Tor classes share one submodule because they share their marshalling and the `TorRuntime` handle; the split is by area, not strictly one file per class.

## The boundary

The rule is whether a symbol *speaks JNI*, not whether it is an export. Exports, anything whose signature mentions `JNIEnv` or a `J*`/`j*` type, Java class descriptors, and the object encoders/decoders all move. Anything expressible in plain Rust types stays.

Where a helper mixed both, it was **split** rather than moved wholesale. `wallet_db` and `block_db` each decoded a `JString` path and then opened a database, so each became two functions: the pure half takes a `PathBuf` and stays in `lib.rs`, and the `zcash_jni` half decodes the path and calls it.

Marshalling lives beside the exports it serves. `zcash_jni/mod.rs` holds only what two or more submodules need (`path_from_jni`, `secret_from_jni`, `parse_txid`, `parse_ufvk`, `parse_ua`, `decode_usk`, ...). Anything with a single caller stayed with that caller, which is why `encode_usk` is in `wallet.rs` and `encode_transaction` in `tor.rs`.

What stays in `lib.rs` is logic: the two database openers, `ReceiverFlags`, `zip317_helper`, four `parse_*` conversions over plain integers, and `UnifiedAddressParser`.

**Invariant, now documented in `AGENTS.md`: outside `#[cfg(test)]`, a logic module must not reference the `jni` crate at all.** Verified:

```console
$ grep -n 'jni::\|JNIEnv\|jlong\|JString\|JObject\|jbyte' lib.rs eip681.rs tor.rs
eip681.rs:5://! [`crate::zcash_jni::eip681`].
```

The single hit is a rustdoc link to our own module. `crate::utils` is the deliberate exception: it is the domain-neutral plumbing (`catch_unwind`, byte-array/string conversion, exception handling) that every submodule builds on, not a logic module.

## Why `zcash_jni` and not `jni`

#2025 names the module `jni`, which shadows the `jni` crate for every bare `jni::` path in the crate root; that PR respells 11 paths in `lib.rs` as `::jni::`. Doing the same here also perturbed rustfmt: `::jni` sorts differently in the import block, and it pushed `encode_blockmeta`'s signature over the line limit, so `cargo fmt --check` failed on otherwise untouched code.

The `zcash_` prefix keeps `jni::` meaning the crate everywhere, needs no qualification, and produced no formatting churn. It is the one intentional divergence from #2025's design.

## Visibility

**Zero items were widened in `lib.rs`.** A private item is already in scope for descendant modules, and `zcash_jni` is a descendant of the crate root, so the helpers it calls needed no change. Only `crate::eip681`'s two pure functions went private -> `pub(crate)`, because `crate::eip681` is a *sibling* of `crate::zcash_jni`, not an ancestor. Those two widenings are commits 9 and 10.

## Verification

Every one of the 11 commits was checked independently:

| check | result |
|---|---|
| `cargo check --lib` | compiles, **zero warnings**, at all 11 commits |
| `cargo fmt --check` | clean at all 11 commits |
| `./scripts/check-jni-symbols.sh` | 78 symbols, baseline matches, at all 11 commits |
| `./scripts/check-jni-move.sh d65b4d2c` | 78 byte-identical / 0 changed, at every commit where the script exists (2 onward) |
| `./scripts/ci-local.sh quick` | detekt, ktlint and Kotlin unit tests pass at the tip |

- **Exported symbol set unchanged: 78 before, 78 after, empty `nm` diff.** Kotlin binds by symbol name and resolves lazily, so a rename would be a runtime crash rather than a compile error. `check-jni-symbols.sh` automates this and is wired into CI; it compares only the symbol *name*, because `nm`'s address column shifts on every rebuild and Mach-O prefixes names with an underscore that ELF does not.
- **Export bodies unchanged: all 78 byte-identical to the base**, proven mechanically by `check-jni-move.sh` rather than by reading.
- No feature-gated modules in this crate, so a plain `--lib` check covers all 78 exports (no `--all-features` caveat, unlike #2025).
- **No Kotlin file changed**, and none needed to: the `external fun` declarations bind to symbol names, which are byte-identical.

## What I did NOT verify

- **`cargo test --lib` is red on this branch, and was already red on the base.** 3 pass, 1 fails: `utils::exception::tests::unknown_any`. Its helper panics via `panic!("{}", val)`, which always produces a `String` payload, so the `"Unknown error occurred"` branch is unreachable. That is fixed by #2016 and is deliberately untouched here. `utils/` has a zero-line diff in this PR, so the failure cannot be attributable to it. Confirmed identical (3 pass, 1 fail) at `d65b4d2c` and at the tip.
- **No Android build, no emulator, no device run.** Only the Rust library was compiled, on macOS/aarch64 as a host `.dylib` — not cross-compiled for any Android ABI. Symbol-name and byte-level body equality are strong evidence the JNI contract is intact, but they do not exercise a single actual JNI call, so nothing here proves runtime behaviour.
- `check-jni-move.sh` covers the `Java_*` exports. The non-exported helpers were compared the same way out-of-band, which is where the "four changed helpers" list above comes from, but that comparison is not committed as a script.

## Known gap

`AGENTS.md` (commit 11) documents `check-jni-symbols.sh` but **not**
`check-jni-move.sh`, and its subject says "the symbol check" for that reason.
Documenting the second script would have meant editing `AGENTS.md`, which would
have broken the byte-identical-tree property that made this restructure
reviewable. It is a one-paragraph follow-up once this lands, not something to
fold in here.

## Downstream

[#2030](https://github.com/zcash/zcash-android-wallet-sdk/pull/2030) (the
Orchard to Ironwood migration proposal) targets this branch, because it adds a
JNI export and belongs in the new module tree. It is rebased on the restructured
history and adds exactly one export; `check-jni-move.sh` run against this branch
reports 78 byte-identical, 0 changed, 1 added.


backend-lib/src/main/rust/lib.rs
```diff
@@ -116,645 +41,18 @@ fn print_debug_state() {
...
-/// When `log_level` defines how the Rust layer logs its events. These values are supported:
-/// `trace` - Logs very low priority, often extremely verbose, information
-/// `debug` - Logs lower priority information
-/// `off` - The logs are completely disabled
-///
-/// # Panics
-///
-/// This method panics if called more than once.
-#[unsafe(no_mangle)]
-pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_initOnLoad<'local>(
```
nuttycom, line 142: Where did this move to?

backend-lib/src/main/rust/zcash_jni/wallet.rs
```diff
@@ -0,0 +1,2512 @@
...
+/// When `log_level` defines how the Rust layer logs its events. These values are supported:
+/// `trace` - Logs very low priority, often extremely verbose, information
+/// `debug` - Logs lower priority information
+/// `off` - The logs are completely disabled
+///
+/// # Panics
+///
+/// This method panics if called more than once.
+#[unsafe(no_mangle)]
+pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_initOnLoad<'local>(
```
dannywillems, line 95: It is here, https://github.com/zcash/zcash-android-wallet-sdk/pull/2029#discussion_r3646408616

backend-lib/src/main/rust/lib.rs
dannywillems: See https://github.com/zcash/zcash-android-wallet-sdk/pull/2029#discussion_r3646460009

dannywillems:
Postponing. Moving into a draft, as discussed online.

PR #2030, opened 2026-07-24
Add an Orchard to Ironwood pool migration proposal

dannywillems:
Adds `Synchronizer.proposeOrchardToIronwoodMigration(account)`.

Targets `maint/v2.5.x` directly. It does not depend on
[#2029](https://github.com/zcash/zcash-android-wallet-sdk/pull/2029) (the
`zcash_jni` split); the new export lives in `lib.rs` alongside the others, and
will move with the rest when that split lands.

## Why

NU6.3 introduces the Ironwood pool and the Orchard turnstile: value may *leave*
the Orchard pool but never re-enter it (`proposal.rs`: "Payments may never be
directed to the Orchard pool after Ironwood activation"). Orchard funds have to
be moved across once, and moved in full, since anything left behind is stranded
in a pool the wallet is leaving.

## What it does

Spends every Orchard note the account holds and sends the maximum to the
account's own internal receiver, with the fee computed so no change returns to
Orchard. Sapling and transparent funds are untouched — this migrates one pool,
it is not a sweep. The proposal is then built, signed and submitted through the
ordinary path.

**The generic capability is upstream and is not reimplemented.**
`propose_send_max_transfer` already spends the whole balance of a caller-chosen
set of pools with the fee computed to leave nothing over. `migration.rs` only
pins the parameters that make that call a migration, and only that instance
crosses the JNI boundary — there is no general send-max in the SDK's API.

Send-max is what a migration *needs*, because the caller cannot compute the
right amount themselves: the fee depends on which inputs the selector picks.

## Three details that are easy to get wrong

Each was verified against the crate source rather than the changelog prose:

- **The recipient is a unified address carrying only an Orchard receiver.**
  `input_selection.rs:898` matches `Address::Unified(addr)` with
  `addr.has_orchard()`, and represents such a payment as an Ironwood-pool output
  once Ironwood is active. Ironwood needs no address type of its own; it shares
  the Orchard receiver encoding. A receiver encoded another way would not take
  that path.
- **The internal scope is used**, so funds return to the account as change
  rather than appearing as an external payment.
- **The proposal is refused unless NU6.3 is active at the chain tip.** Before
  activation the same call is accepted but builds an *Orchard* output — a
  self-send that migrates nothing and costs a fee.

## All-or-nothing, deliberately

`MaxSpendMode::Everything`, not `MaxSpendable`. For a note that is not yet
spendable, `zcash_client_sqlite` (`wallet/common.rs:524`) maps the two modes to
`Err(IneligibleNotes)` and `Ok(None)` respectively — so `MaxSpendable` would
silently skip such notes and report a successful migration that left funds
behind. The eligibility check covers only the pools being spent, so an
unconfirmed *Sapling* note does not block this.

## Every argument is a named binding

`propose_send_max_transfer` takes nine positional arguments. Rather than
annotate them at the call site, each is bound to a named local with its
rationale above the binding, so the reasoning is found by a reader who asks
what a value is, not only by one reading this particular call:

```rust
// Orchard only. Sapling and transparent funds are deliberately left where
// they are: this migrates one pool, it is not a sweep of the wallet.
let spend_pools = [ShieldedPool::Orchard];
...
let mode = MaxSpendMode::Everything;
```

## Privacy

The migration reveals the account's Orchard balance on-chain: it is a single
transaction of exactly that value. Splitting a crossing into less-identifying
amounts is a separate mechanism and is deliberately not attempted here. This is
recorded in the doc comments so it is not mistaken for a private migration path.

## Validation

- `cargo check --lib` — 0 errors, 0 warnings
- `cargo fmt --check` — clean
- `nm` on the built cdylib — 79 exported `Java_*` symbols, 78 from `maint`
  plus `..._RustBackend_proposeOrchardToIronwoodMigration`
- `./scripts/ci-local.sh quick` — detekt, ktlint and Kotlin unit tests pass

`scripts/check-jni-symbols.sh` and `check-jni-move.sh` are not available on this
base; they arrive with #2029. The `nm` count above is the manual equivalent.
Once #2029 lands, this branch inherits them and the baseline needs a one-symbol
`--update`.

## No tests yet, and what they should be

**Nothing here is covered by a test, and nothing exercises a real JNI call.**
The Rust compiles, the symbol is exported, and the Kotlin chain compiles against
it. That is the whole of the evidence.

The gap is worth closing, and there is a usable harness for it.
`zcash_client_sqlite` ships a `test-dependencies` feature exposing
`TestBuilder`/`TestDsl`, which builds a real SQLite wallet with synthetic blocks
and notes at chosen heights — no network and no device. Upstream tests its own
send-max that way (`data_api/testing/pool.rs`), including a
`propose_send_max_transfer` helper and `MaxSpendMode::Everything`.

Because the heights are chosen, the chain tip can be placed past NU6.3, which
makes the three behaviours that actually matter testable:

1. **Pre-activation guard** — tip below NU6.3 yields the error, not a proposal.
   Cheapest, and currently unexercised.
2. **The crossing** — tip above NU6.3 with Orchard notes present: assert the
   proposal's output pool is **Ironwood**. This is the claim the feature rests
   on, and it is presently inferred from reading `input_selection.rs` rather
   than demonstrated.
3. **`Everything` semantics** — an unconfirmed Orchard note yields
   `IneligibleNotes` rather than a partial migration.

Two decisions before that can happen: it needs a `zcash_client_sqlite`
dev-dependency with `features = ["test-dependencies"]`, and `cargo test` does
not run in CI on this branch until
[#2016](https://github.com/zcash/zcash-android-wallet-sdk/pull/2016) lands.

Darkside is a possible end-to-end route later — `darksidewalletd` does support
Ironwood (`startIronwoodCommitmentTreeSize` in `darkside.proto`) — but it is not
in CI and the repo pins no revision of it.

### DIFF
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/FakeRustBackend.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/FakeRustBackend.kt
@@ -113,6 +113,10 @@ class FakeRustBackend(
         error("Intentionally not implemented yet.")
     }
 
+    override suspend fun proposeOrchardToIronwoodMigration(accountUuid: ByteArray): ProposalUnsafe {
+        error("Intentionally not implemented yet.")
+    }
+
     override suspend fun proposeShielding(
         accountUuid: ByteArray,
         shieldingThreshold: Long,
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt
@@ -436,6 +436,17 @@ class RustBackend private constructor(
             )
         }
 
+    override suspend fun proposeOrchardToIronwoodMigration(accountUuid: ByteArray): ProposalUnsafe =
+        withContext(SdkDispatchers.DATABASE_IO) {
+            ProposalUnsafe.parse(
+                proposeOrchardToIronwoodMigration(
+                    dataDbFile.absolutePath,
+                    accountUuid,
+                    networkId = networkId,
+                )
+            )
+        }
+
     override suspend fun proposeShielding(
         accountUuid: ByteArray,
         shieldingThreshold: Long,
@@ -877,6 +888,13 @@ class RustBackend private constructor(
             dbDataPath: String,
             accountUuid: ByteArray,
             uri: String,
+            networkId: Int,
+        ): ByteArray
+
+        @JvmStatic
+        private external fun proposeOrchardToIronwoodMigration(
+            dbDataPath: String,
+            accountUuid: ByteArray,
             networkId: Int,
         ): ByteArray
 
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -102,6 +102,7 @@ use crate::utils::{
 };
 
 mod eip681;
+mod migration;
 mod tor;
 mod utils;
 
@@ -2189,6 +2190,36 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeTr
             None,
         )
         .map_err(|e| anyhow!("Error creating transaction proposal: {}", e))?;
+
+        Ok(utils::rust_bytes_to_java(
+            env,
+            Proposal::from_standard_proposal(&proposal)
+                .encode_to_vec()
+                .as_ref(),
+        )?
+        .into_raw())
+    });
+    unwrap_exc_or(&mut env, res, ptr::null_mut())
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeOrchardToIronwoodMigration<
+    'local,
+>(
+    mut env: JNIEnv<'local>,
+    _: JClass<'local>,
+    db_data: JString<'local>,
+    account_uuid: JByteArray<'local>,
+    network_id: jint,
+) -> jbyteArray {
+    let res = catch_unwind(&mut env, |env| {
+        let _span = tracing::info_span!("RustBackend.proposeOrchardToIronwoodMigration").entered();
+        let network = parse_network(network_id as u32)?;
+        let mut db_data = wallet_db(env, network, db_data)?;
+        let account_uuid = account_id_from_jni(env, account_uuid)?;
+
+        let proposal =
+            crate::migration::propose_orchard_to_ironwood(&mut db_data, &network, account_uuid)?;
 
         Ok(utils::rust_bytes_to_java(
             env,
--- a/backend-lib/src/main/rust/migration.rs
+++ b/backend-lib/src/main/rust/migration.rs
@@ -1,0 +1,140 @@
+//! Orchard to Ironwood pool migration.
+//!
+//! The generic capability this builds on lives upstream:
+//! [`propose_send_max_transfer`] already spends the entire spendable balance of
+//! a caller-chosen set of shielded pools to a single recipient, computing the
+//! fee so that nothing is left over. This module does not reimplement any of
+//! that, nor does it expose the general form. It pins every parameter that
+//! makes the call a migration, and that single instance is all the SDK offers.
+//!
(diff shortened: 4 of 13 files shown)


backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/FakeRustBackend.kt
```diff
@@ -113,6 +113,10 @@ class FakeRustBackend(
         error("Intentionally not implemented yet.")
     }
 
+    override suspend fun proposeOrchardToIronwoodMigration(accountUuid: ByteArray): ProposalUnsafe {
```
dannywillems, line 116: LGTM. Follow above.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt
```diff
@@ -436,6 +436,17 @@ class RustBackend private constructor(
             )
         }
 
+    override suspend fun proposeOrchardToIronwoodMigration(accountUuid: ByteArray): ProposalUnsafe =
```
dannywillems, line 439: LGTM. Follow above.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt
```diff
@@ -880,6 +891,13 @@ class RustBackend private constructor(
             networkId: Int,
         ): ByteArray
 
+        @JvmStatic
```
dannywillems, line 894: LGTM.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2201,6 +2202,36 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeTr
...
+    let res = catch_unwind(&mut env, |env| {
+        let _span = tracing::info_span!("RustBackend.proposeOrchardToIronwoodMigration").entered();
+        let network = parse_network(network_id as u32)?;
+        let mut db_data = wallet_db(env, network, db_data)?;
+        let account_uuid = account_id_from_jni(env, account_uuid)?;
+
+        let proposal =
+            crate::migration::propose_orchard_to_ironwood(&mut db_data, &network, account_uuid)?;
+
+        Ok(utils::rust_bytes_to_java(
```
dannywillems, line 2224: This follows above pattern.

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,140 @@
...
+/// Proposes migrating the account's Orchard balance into the Ironwood pool.
+///
+/// Ironwood needs no address type of its own: it shares the Orchard receiver
+/// encoding, and once Ironwood is active `zcash_client_backend` represents a
+/// payment to an Orchard receiver as an Ironwood-pool output, which the
+/// transaction builder places in the Ironwood bundle. Sending the maximum from
+/// Orchard to the account's own internal Orchard receiver is therefore exactly
+/// the pool crossing, with nothing left behind in Orchard.
+///
+/// Uses [`MaxSpendMode::Everything`], so the proposal fails rather than
```
dannywillems, line 46: LGTM, see https://github.com/zcash/librustzcash/blob/main/zcash_client_backend/src/data_api.rs#L212

backend-lib/src/main/rust/migration.rs
```diff
@@ -0,0 +1,140 @@
...
+/// migrating only part of the balance. `MaxSpendable` would silently skip any
+/// Orchard note that is not yet spendable (`zcash_client_sqlite` maps the two
+/// modes to `Ok(None)` and `Err(IneligibleNotes)` respectively for such a
+/// note), leaving funds in a pool the wallet is trying to leave and reporting
+/// success. The eligibility check covers only the pools being spent, so an
+/// unconfirmed Sapling note does not block this.
+///
+/// Fails if NU6.3 is not yet active. Before activation the identical call is
```
dannywillems, line 54: Maybe we should simply raise an error. We want to have it used only when Ironwood is activated.

dannywillems (commented):
Self-ACK, modulo design decision.

nuttycom (approved):
utACK b580c66098c82273eff2633ebff2ba15eaac897f, this looks great.

PR #2036, opened 2026-07-24
Migrate to zcash_client_backend 0.24, zcash_primitives 0.29 et al.

dannywillems:
Moves `prerelease/snapshot-v2.6.6` from the zcash_client_backend 0.23
generation onto the 0.24.0-rc.1 one, pinned by git rev, so Ironwood support and
the Orchard-to-Ironwood migration can be backported afterwards. This is the
dependency migration only; no Ironwood, no migration engine, no Slipstream.

The work is extracted from #2021 (`android-slipstream-ironwood-chp`), where the
same migration is entangled across three feature commits: `3153bd9a`
(noop-sk, migration engine), `6ea9fb2e` (Nesence, Slipstream) and `1185ac9a`
(Kris Nuttycombe, Ironwood pin advance). Credit for the underlying dependency
and API work belongs to them; every commit here carries them as co-authors.

## Status: BLOCKED, do not merge, decision needed

`cargo check --lib` does **not** pass yet. The non-voting surface is fully
migrated and clean; the voting JNI layer does not compile, and finishing it is a
design decision rather than a mechanical adaptation. Details in
"The voting blocker" below.

## What is done

`backend-lib/Cargo.toml`, `Cargo.lock`, `lib.rs`, `tor.rs` and the
rename-level drift in `voting/` are migrated. Zero errors and zero warnings
outside the voting module.

Version requirements are bumped alongside the new `[patch.crates-io]` block
rather than left on the old minor lines. A `[patch]` only replaces the resolved
crate when its version satisfies the requirement declared in `[dependencies]`;
a requirement only compatible with the published generation the patch replaces
makes cargo silently ignore the patch and pull a second, unpatched copy from
crates.io, splitting the graph into two incompatible copies of the same types.

Deliberately **not** brought across from #2021, because nothing on this branch
needs them: `zcash_pool_migration_backend`, `slipstream-core`, `tokio`,
`tracing-android`, `ur`, `ur-registry`, `shardtree`, and the `signer` feature on
`pczt`. `incrementalmerkletree` is kept because `voting/notes.rs` uses it.

### librustzcash API changes adapted

| Change | Resolution |
| --- | --- |
| `TransparentOutputFilter` -> `CoinbaseFilter` | `::All` -> `::AllTransparentOutputs` |
| `WalletDb::rewind_to_height` | -> `truncate_to_height` |
| `WalletTransparentOutput` generic over `AccountId`; `from_parts` gained `recipient_account`, `recipient_key_scope`, `funding_account` | JNI and Tor UTXO paths have none of them; pass `None`. `LwdConn::with_taddress_utxos` becomes generic over `AccountId` |
| `get_spendable_transparent_outputs` gained a `LockFilter` | `LockFilter::Unfiltered`: the one caller is a balance display, not a spend selection, so it must report the true total regardless of any lock an in-flight proposal holds |
| `propose_transfer` gained `&SpendPolicy` and `lock_inputs` | `SpendPolicy::default()` preserves the previous fully-shielded selection; `None` takes no locks |
| `propose_shielding` gained `lock_inputs` | `None` |
| `Proposal::try_into_standard_proposal` gained consensus parameters | pass `&network` |
| `create_proposed_transactions`' last argument | now an optional expiry height rather than a proposed tx version; `None` preserves behaviour either way |
| `create_pczt_from_proposal` gained a target expiry height and an Orchard `BundleType` | `None` and `BundleType::DEFAULT` |
| `Pczt::serialize` returns a `Result` | mapped into the existing `anyhow` error path at all three call sites |
| Orchard `ProvingKey::build` takes an `OrchardCircuitVersion` | derived from the consensus branch carried by the PCZT via `bundle_version_for_branch` |
| Orchard `VerifyingKey::build` likewise | pass `None` so the extractor derives it from the bundle's own `BundleVersion` |
| `ShieldedProtocol` -> `ShieldedPool` | rename |
| `connect_to_lightwalletd` gained `allow_onion_services` | re-derive the same `.onion` host check at the call site, so custom onion lightwalletd servers keep working without changing the wrapper's signature |
| lightwalletd `TreeState` gained `ironwood_tree` | the `android-test-fixtures` fixture leaves it empty and keeps populating `orchard_tree`; this branch's voting notes are still Orchard notes |

Each follows #2021's resolution, so the two branches stay convergent. Most also
match `maint/v2.5.x` `4843902c`; the note-locking arguments, `LockFilter`,
`allow_onion_services` and `ironwood_tree` are newer than that commit's pin and
follow #2021 only.

### zcash_voting 1.0 rename-level drift adapted

`types::ProofProgressReporter` -> `types::ProgressReporter`;
`share_tracking::compute_share_nullifier` -> `share::compute_nullifier` (return
type tightened from `Vec<u8>` to a fixed-size array, so the explicit length
check is now done by the type system); `tree_sync::VanWitness` ->
`vote::VanWitness`; `types::chunk_notes` -> `note_bundling::chunk_notes`.

## The voting blocker

On this branch voting is **not** feature-gated: `zcash_voting` is a hard
dependency and `voting.rs` plus its 12 submodules compile unconditionally. So
unlike #2021 (which made `chp-voting` optional and off by default) the voting
code must compile against the new stack.

`zcash_voting` 1.0 is a deliberate breaking redesign (`0b5fffb feat!: v2 voting
api`), landed *before* the librustzcash realignment, so no fork rev exists that
carries the old 0.11 API on the new librustzcash. Keeping 0.11 unpatched is not
an option either: it pins orchard 0.14 / zcash_protocol 0.9 / pczt 0.7 and would
split the graph.

20 errors remain, all in `voting/`. They are not renames:

- `decompose::decompose_weight` was **deleted** from the crate.
- `vote_commitment::sign_cast_vote` and `hotkey::spending_key_from_hotkey_seed`
  became `pub(crate)`, so `signCastVoteNative` has no public path and cannot be
  reimplemented locally without duplicating hotkey derivation.
- `VotingDb` lost or privatised `setup_bundles`, `generate_hotkey`,
  `store_vote_tx_hash`, `store_commitment_bundle`, `get_delegation_submission`,
  `get_delegation_submission_with_keystone_sig`, `build_vote_commitment` and
  `record_share_delegation`, replaced by a different composition
  (`CommittedVote`, `VoteRecovery`, `VoteSigner`,
  `get_delegation_submission_with_signature`).
- `VotingHotkey` is now opaque (was a struct with `secret_key` / `public_key` /
  `address` fields), `VoteRecord` lost `submitted`, and `VanWitness::auth_path`
  changed element type.

#2021's port of this **does** compile (verified: `cargo check --lib --features
chp-voting` on `android-slipstream-ironwood-chp` is clean and warning-free), so
a working reference exists. But adopting it is a design change, not a
dependency bump:

1. **It changes the JNI ABI.** #2021 removes `decomposeWeightNative`,
   `signCastVoteNative` and `storeCommitmentBundleNative`, adds
   `recordVcPositionNative`, `recoverCommittedVoteNative` and
   `scheduledShareSubmitAtNative`, and changes `buildVoteCommitmentNative`'s
   parameter list. Kotlin binds by symbol name and resolves lazily, so this
   needs lockstep changes in `VotingRustBackend.kt`,
   `TypesafeVotingBackend.kt`, `TypesafeVotingBackendImpl.kt` and the
   androidTest sources.
2. **It changes hotkey key management.** `buildVoteCommitmentNative` moves from
   a wallet-derived `hotkey_seed` to a 64-byte stored hotkey secret produced by
   `generate_random_voting_hotkey`. That is a change to what secret material
   callers must hold and persist, with a migration implication for existing
   users. It must not ride along inside a dependency bump.

### Options

- **A. Port voting to the v2 API following #2021.** Yields a fully working
  branch, keeps convergence with #2021, but pulls in the ABI and hotkey
  key-management changes above plus their Kotlin lockstep, and imports Ironwood
  semantics this branch otherwise does not have.
- **B. Feature-gate voting off by default, as #2021 did.** Makes
  `cargo check --lib` clean immediately and keeps this PR a true dependency
  bump, deferring the voting port to its own reviewed PR. This is the
  recommended option, but it is a design change to this branch's shape and is
  not being done unilaterally.

## Validation

- `cargo check --lib` — 20 errors, all in `voting/`; zero errors and zero
  warnings elsewhere.
- `cargo check --lib --all-features` — same 20; the `android-test-fixtures`
  `TreeState` fixture is fixed and no longer contributes errors.
- `cargo fmt --check` — clean (plain `cargo fmt`, repo-pinned toolchain, never
  `+nightly`).
- `cargo test --lib` — cannot run until the crate compiles.
- **JNI export symbol diff: empty.** All 138 `Java_*` exports are byte-identical
  to `origin/prerelease/snapshot-v2.6.6` at source level, so no Kotlin change is
  required by anything in this PR. (Compared from source rather than from `nm`,
  since the cdylib cannot be linked yet.)

### DIFF
--- a/backend-lib/Cargo.lock
+++ b/backend-lib/Cargo.lock
@@ -159,6 +159,8 @@ dependencies = [
  "tor-dirmgr",
  "tor-error",
  "tor-guardmgr",
+ "tor-hsclient",
+ "tor-hscrypto",
  "tor-keymgr",
  "tor-linkspec",
  "tor-llcrypto",
@@ -1101,7 +1103,7 @@ checksum = "dd5f2b7218a51c827a11d22d1439b598121fac94bf9b99452e4afffe512d78c9"
 dependencies = [
  "heck",
  "indexmap 2.14.0",
- "itertools 0.13.0",
+ "itertools 0.14.0",
  "proc-macro-crate",
  "proc-macro2",
  "quote",
@@ -1215,7 +1217,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1445,8 +1447,7 @@ dependencies = [
 [[package]]
 name = "equihash"
 version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "306286e8dcc39ab3dfceb74c792ce8baffdab90591321d3ffaae64829734c37f"
+source = "git+https://github.com/zcash/librustzcash?rev=3a10e7fec7afd3a9748cb4fa7f7db853f4537eeb#3a10e7fec7afd3a9748cb4fa7f7db853f4537eeb"
 dependencies = [
  "blake2b_simd",
  "corez",
@@ -1465,7 +1466,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1482,8 +1483,7 @@ dependencies = [
 [[package]]
 name = "f4jumble"
 version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0d42773cb15447644d170be20231a3268600e0c4cea8987d013b93ac973d3cf7"
+source = "git+https://github.com/zcash/librustzcash?rev=3a10e7fec7afd3a9748cb4fa7f7db853f4537eeb#3a10e7fec7afd3a9748cb4fa7f7db853f4537eeb"
 dependencies = [
  "blake2b_simd",
 ]
@@ -1582,6 +1582,18 @@ name = "fluid-let"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "749cff877dc1af878a0b31a41dd221a753634401ea0ef2f87b62d3171522485a"
+
+[[package]]
+name = "flume"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "da0e4dd2a88388a1f4ccc7c9ce104604dab68d9f408dc34cd45823d5a9069095"
+dependencies = [
+ "futures-core",
+ "futures-sink",
+ "nanorand",
+ "spin",
+]
 
 [[package]]
 name = "fnv"
@@ -2375,7 +2387,7 @@ version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7a1886916523694cd6ea3d175f03a1e5010699a2a4cc13696d83d7bea1d80638"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


dannywillems:
I'll go on the large branch tomorrow instead. It seems that we're gonna end up merging that one. No need for splitting.

PR #2037, opened 2026-07-24
Port CI managed-device, TreeState proto citation, and compile-locally docs from maint/v2.5.x

dannywillems:
Human comment: simple PR. I've got agents backporting changes from 2.5.x

-----------------

Ports three self-contained changes that landed on `maint/v2.5.x` but are not
yet on this branch, per the standing "maint changes get backported" rule. Each
is its own commit with `Backport-of` / `Backport-method` trailers.

## What's ported

1. **CI: replace the emulator.wtf job with a Gradle Managed Device**
   (`Backport-of: 5b8a5a22`). Replaces `test_android_modules_wtf`
   (`testDebugWithEmulatorWtf`, needs a paid `EMULATOR_WTF_API_KEY`) with
   `test_android_modules_emulator`, which runs the four instrumentation suites
   on the `pixel2Min` Gradle Managed Device already defined in
   `zcash-sdk.android-conventions.gradle.kts`. This is what keeps CI from
   staying red on the emulator.wtf service failure.
2. **Cite lightwalletd's bindings for the TreeState field numbers**
   (`Backport-of: 3e0a21d2`). Two-line comment in `service.proto`.
3. **AGENTS.md: document how to compile locally** (`Backport-of: cf187eae`).
   Adds the "`quick` compiles the Kotlin", the java-not-on-`PATH` export block,
   and the `cargo check --lib` Rust shortcut with the JNI-mismatch-is-a-runtime-
   crash caveat.

## Not a clean cherry-pick (why the trailers say "reimplemented")

- **CI**: this branch's `pull-request.yml` has diverged from maint's, with
  different `actions/checkout` and `actions/upload-artifact` pins and a
  `rust-cache: true` Setup input. The job semantics are transplanted while
  keeping this branch's own pins. Only the wtf job is replaced; the
  `check_emulator_wtf_secrets` job and the Firebase Test Lab job are left
  exactly as maint left them.
- **AGENTS.md**: this branch's `AGENTS.md` has diverged (it carries the
  worktree-layout and commit-convention sections maint lacks), so the guidance
  is applied at matching anchors.
- The proto change is textually identical to maint's, hence `cherry-pick`.

## Known follow-up (deliberately not in this PR)

`AGENTS.md` still references the now-removed `test_android_modules_wtf` job and
"paid services (`emulator.wtf`)" (the intro and the "What cannot be run locally"
section). Those were left untouched here to keep commit 3 a faithful port of
`cf187eae`; happy to fold a consistency update in if preferred.

## Verification

`pull-request.yml` parses as valid YAML and the jobs list resolves to
`test_android_modules_emulator` (wtf job gone, FTL + secret-check unchanged).

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -352,26 +352,11 @@ jobs:
           name: Test Android modules with FTL results
           path: ~/artifacts
 
-  # DISABLED: the emulator.wtf organization has exhausted its credits, so this job fails on every
-  # run with `Http error 402: You've run out of free credits on emulator.wtf`. The plugin then
-  # parses that error body as a test-run response, so what actually surfaces is the misleading
-  # `com.google.gson.JsonSyntaxException: Missing required properties: testRunId`. Nothing about
-  # the build or the tests is wrong.
-  #
-  # To re-enable: configure billing at
-  # https://emulator.wtf/o/005b7bef-ece6-4af6-a3e6-62c245489044/billing, then restore the
-  # `if:` condition recorded below.
-  #
-  # Note that disabling this does NOT fall back to Firebase Test Lab: test_android_modules_ftl is
-  # gated on the emulator.wtf API key being *absent*, and the key is still set — it is the credits
-  # that ran out. Device-test coverage is therefore zero, which is unchanged from the status quo
-  # (the job was already failing rather than testing), but it is zero rather than merely red.
-  # Removing EMULATOR_WTF_API_KEY from the repository secrets would route these tests to FTL
-  # instead, if that tradeoff is preferred over fixing the billing.
-  test_android_modules_wtf:
-    # Original condition: needs.check_emulator_wtf_secrets.outputs.has-secrets == 'true'
-    if: false
-    needs: [ validate_gradle_wrapper, check_emulator_wtf_secrets ]
+  # Runs the Android instrumentation suites on a Gradle Managed Device: a
+  # hardware-accelerated emulator booted inside the runner. This replaces the
+  # former emulator.wtf cloud job and depends on no external test service.
+  test_android_modules_emulator:
+    needs: [validate_gradle_wrapper]
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -383,6 +368,15 @@ jobs:
           # No job here pushes or otherwise re-uses the checkout credential, so keep it
           # out of .git/config where it would be picked up by artifact uploads.
           persist-credentials: false
+      # ubuntu-latest exposes /dev/kvm, but the default udev rules do not grant
+      # the runner user access to it. This rule opens it so the managed-device
+      # emulator can use KVM acceleration instead of falling back to software.
+      - name: Enable KVM
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' \
+            | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
       - name: Setup
         id: setup
         timeout-minutes: 30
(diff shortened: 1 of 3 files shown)


nuttycom (dismissed):
utACK

dannywillems:
Fixing conflicts.

dannywillems:
Rebased onto the updated base and resolved the `pull-request.yml` conflict.

The conflict was semantic: the base had since disabled the emulator.wtf job (`if: false`, out-of-credits) accepting zero device-test coverage, while this PR replaces that job with an in-runner Gradle Managed Device (the same approach already on `maint/v2.5.x`), which restores coverage with no external service. Resolved in favour of the GMD replacement, which supersedes the reason for the disable.

Crucially, the merge preserves the base's two additions inside the new `test_android_modules_emulator` job: `persist-credentials: false` on checkout (zizmor hardening) and the `slipstream-app-id`/`slipstream-app-private-key` Setup inputs (private-slipstream-dep auth from #2040), without which the build can't fetch `slipstream-core`. The proto-citation and AGENTS.md commits replayed unchanged.

.github/workflows/pull-request.yml
```diff
@@ -352,26 +352,11 @@ jobs:
...
-  # Removing EMULATOR_WTF_API_KEY from the repository secrets would route these tests to FTL
-  # instead, if that tradeoff is preferred over fixing the billing.
-  test_android_modules_wtf:
-    # Original condition: needs.check_emulator_wtf_secrets.outputs.has-secrets == 'true'
-    if: false
-    needs: [ validate_gradle_wrapper, check_emulator_wtf_secrets ]
+  # Runs the Android instrumentation suites on a Gradle Managed Device: a
+  # hardware-accelerated emulator booted inside the runner. This replaces the
+  # former emulator.wtf cloud job and depends on no external test service.
+  test_android_modules_emulator:
```
nesence-m, line 358: This job now exists on the base verbatim (same KVM step, emulator install, `pixel2Min` task list, `maxConcurrentDevices=1`, swiftshader flag, artifact globs) — it landed via the maint merge chain from 5b8a5a22. On top of that, the base's `pull-request.yml` has moved further: new `actions/checkout` pins (9c091bb2), the emulator.wtf scaffolding fully removed including `check_emulator_wtf_secrets` (85589edd), the FTL job commented out (27156d3f), Rust build caching/budget adjustments (1100a3a5), and new `shellcheck`/`rustfmt`/`clippy`/`test_rust_unit` jobs. Resolving this PR's conflict in favour of this hunk would regress all of that; resolving in favour of the base makes the commit a no-op. Either way there is nothing left for this file to contribute.

lightwallet-client-lib/src/main/proto/service.proto
```diff
@@ -111,6 +111,8 @@ message Exclude {
 
 // The TreeState is derived from the Zcash z_gettreestate rpc.
 message TreeState {
+    // The field numbers below must follow lightwalletd's generated bindings:
+    // https://github.com/zcash/lightwalletd/blob/master/walletrpc/service.pb.go
```
nesence-m, line 115: Re-adding this citation is now actively wrong on this branch: 3e0a21d2 is already an ancestor of the base, and 3224aa72 ("Derive the lightwallet protos from the vendored v0.5.0 subtree") subsequently turned this file into a generated artifact — it now opens with `// DERIVED FILE -- DO NOT EDIT.` and is produced by `scripts/update-lightwallet-protocol.sh` from the vendored `lightwallet-client-lib/lightwallet-protocol/` subtree. That commit's message explicitly supersedes the pinned-citation edits: the protos now *come from* lightwalletd's canonical repo, so "there is nothing left to reserve or cite." A hand-edit here would violate the banner and be silently clobbered on the next re-vendor.

AGENTS.md
```diff
@@ -92,6 +109,16 @@ prefix is acceptable (`fix: ...`, `chore: ...`). Keep the first line under
 - The SDK and related libraries are Kotlin + Rust. Changes that cross the
   JNI boundary (`backend-lib/src/main/rust/*` and the Kotlin `Jni*` model
   classes) require updating both sides in lockstep.
+- The Rust backend is fastest to check on its own, without Gradle or a JDK:
+
+  ```bash
+  cd backend-lib && cargo check --lib && cargo fmt --check
```
nesence-m, line 115: cf187eae itself is already an ancestor of the base, so this entire commit (the `quick`-compiles-the-Kotlin note, the `JAVA_HOME` export block, and this `cargo check --lib` section) is present there verbatim. One conflict-resolution hazard if this PR were merged anyway: the base has since amended the stage-table line to `quick    # fast + rust + unit tests (~5m)` (the `quick` stage now runs the Rust tests too, per 3a7644e6); this PR's older `# fast + unit tests (~5m)` wording would drop that if a resolver took this side.

nesence-m (commented):
The porting work itself was faithful when it was written, but the branch has since caught up through the regular merge-forward chain, and that changes the verdict: **all three `Backport-of` source commits (5b8a5a22, 3e0a21d2, cf187eae) are now ancestors of `feature/ironwood-slipstream`** — they arrived via `maint/v2.5.x → maint/v2.6.x` (18828168), `maint/v2.5.x → main` (7bd4e7bf), and ultimately `maint/v2.8.x → feature/ironwood-slipstream` (ba9f2c92). Verified with `git merge-base --is-ancestor` for each of the three SHAs against the current base head. The PR is therefore fully superseded and currently unmergeable (`mergeable_state: dirty`) because the base's copies of all three files have since moved past this branch. I'd recommend closing rather than resolving the conflicts — resolution can only reproduce what the base already has, or regress it (details in the line comments).

Non-anchored notes:

1. **Stale approval.** The utACK was submitted against 82008de1; the branch was subsequently rebased and force-pushed to c201213d, so the approval predates the current head. Moot if this closes, but worth flagging if anyone considers merging the current head on the strength of it.
2. **Branch-model observation, not a defect.** This port went straight from `maint/v2.5.x` to the feature branch, and the merge-forward chain delivered the identical commits within about a day — which is exactly why the PR went redundant. Question for the workflow: should self-contained maint changes ever be ported per-commit onto feature branches, or should feature branches only pick them up via the periodic maint→main→feature merges? If the latter, this class of PR can be skipped entirely.
3. **CI on this head is informative but moot.** `test_android_modules_emulator` failed with `UnsatisfiedLinkError` on every `VotingRustBackendTest` case (no `Java_..._VotingRustBackend_*` exports in the built library — the feature-gated voting JNI isn't enabled in the androidTest native build on this branch). Since the base now carries this exact job, that failure is a branch-level issue to fix on the base, not something this PR introduced or can fix here. The `static_analysis_android_lint` failure is likewise unrelated to the three files touched (workflow YAML, markdown, a proto comment — none feed lint).

nesence-m (dismissed):
utACK

PR #2038, opened 2026-07-24
Reschedule-transfers debug tool: use the typed migration API, drop raw SQL

dannywillems:
Removes the last three hand-rolled SQL statements in the migration JNI glue,
routing them through the typed `pool_migration` API instead. This continues the
direction of `e887b63e` ("Remove direct wallet-database schema and row
manipulation from the backend").

## Scope

One function: the debug JNI export `debugRescheduleTransfersNative` in
`backend-lib/src/main/rust/migration.rs`. It was the only production code left
that ran raw SQL against the wallet DB's migration tables:

- `SELECT m.id ... orchard_ironwood_migrations JOIN accounts` (migration id by account)
- `SELECT tx_id ... orchard_ironwood_migration_transactions WHERE kind='transfer' AND state NOT IN ('broadcast','mined') ORDER BY scheduled_height` (pending transfers)
- `UPDATE orchard_ironwood_migration_transactions SET scheduled_height=?, anchor_boundary=COALESCE(?, anchor_boundary)` (per-row reschedule)

## Change

Replaced with the typed store idiom already used everywhere else in this file
(and by the sibling `pendingTransferProposalNative` immediately below):

- `Backend::new(...)` -> `get_migration()` for the read (returns `None` -> return `0`, as before).
- Filter `state.transactions()` to transfer-kind, non-`Broadcast`/`Mined` transactions, sorted by `scheduled_height()` ascending (the old `ORDER BY`).
- Rewrite each pending transfer's `scheduled_height` / `anchor_boundary` through the public `MigrationTransaction::from_parts(...)` + `MigrationState::from_parts(...)`, then `replace_migration(&new_state)`.

Behavior is unchanged: tip+1 packing (`FIRST_DELAY_BLOCKS`/`STRIDE_BLOCKS`), a
fresh `natural_anchor_height` anchor only for the earliest-due transfer (the
others keep their existing boundary, i.e. the old `COALESCE(NULL, anchor_boundary)`),
and the rescheduled count is returned. The three load-bearing comment blocks
(the packing rationale, the `natural_anchor_height` rationale, and the
`is_prove_ready`/anchor-boundary note explaining why only the earliest transfer
gets a fresh anchor) are preserved.

## Notes

- Imports: added `MigrationTransaction` and `std::collections::HashMap`; dropped
  the now-unused `rusqlite::OptionalExtension`.
- No schema or connection-mode change; still the same `store_conn`, now used
  only through the typed backend.
- Verification: `cd backend-lib && cargo check --lib` is green; `cargo fmt --check`
  clean (repo-pinned toolchain).

Background: this is the "route through `zcash_client_sqlite::pool_migration`
instead of raw SQL" item from the SQL review of this branch (the one case where
an upstream typed API already exists for the hand-rolled query).

### DIFF
--- a/backend-lib/src/main/rust/migration.rs
+++ b/backend-lib/src/main/rust/migration.rs
@@ -39,7 +39,8 @@ use jni::{
 };
 use prost::Message;
 use rand::rngs::OsRng;
-use rusqlite::{Connection, OptionalExtension};
+use rusqlite::Connection;
+use std::collections::HashMap;
 use std::ptr;
 
 use zcash_client_backend::data_api::wallet::input_selection::LockFilter;
@@ -53,8 +54,8 @@ use zcash_protocol::value::Zatoshis;
 use zcash_protocol::{PoolType, ShieldedPool};
 
 use zcash_pool_migration_backend::engine::{
-    self, MigrationCrypto, MigrationPlan, MigrationState, MigrationTxId, MigrationTxKind,
-    MigrationTxState, PoolMigrationRead, PoolMigrationWrite, ProveError,
+    self, MigrationCrypto, MigrationPlan, MigrationState, MigrationTransaction, MigrationTxId,
+    MigrationTxKind, MigrationTxState, PoolMigrationRead, PoolMigrationWrite, ProveError,
 };
 use zcash_pool_migration_backend::wallet::{WalletMigrationProver, WalletProveError};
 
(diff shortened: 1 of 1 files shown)


nuttycom (approved):
utACK d5eca9ed400c647da2ebd15a070016c3c09acdf9

nuttycom (approved):
re-utACK 1d253a8ed74dc5a5de4004532deb8c1168c8eafd

PR #2042, opened 2026-07-24
Use a 12-block anchor bucket interval on test networks

### LINKED ISSUE #2041: Use a 12-block anchor bucket interval on test networks
The ZIP 318 pool-migration anchor bucketing interval is 144 blocks, roughly three hours at the target block spacing. Exercising a migration end to end against a test network therefore means waiting out multiple buckets per transfer, which makes testnet testing impractical.

librustzcash is making the interval configurable (zcash/librustzcash#2759). This SDK should select it per network: the ZIP 318 default on mainnet, and a 12-block interval with delay distributions derived from it on test networks.

The interval must be applied wherever the SDK opens a `WalletDb`, not just on the scanning path: it governs both which note commitment tree checkpoints are retained and which boundaries a migration may anchor to, and those must agree.

---

nuttycom:
Closes #2041. Resolves MOB-1522

Depends on zcash/librustzcash#2759, which makes the ZIP 318 anchor bucketing interval configurable. The `[patch.crates-io]` block is pinned to that PR's tip, `cd2b9a6a27dd784df3d245cd047bfeade21d9430`.

## What this does

`anchor_retention_interval(NetworkType)` selects the grid: `AnchorRetentionInterval::ZIP_318` on mainnet, a 12-block interval on Test and Regtest. `wallet_db()` applies it, so every wallet handle the JNI layer hands out is on the same grid.

`migration.rs`'s `open_at()` opens a *second, independent* `WalletDb`, and it gets the interval through the same helper. This matters: the interval is in-memory configuration rather than persisted state, so an unconfigured second open would have reported the ZIP 318 grid back to the migration engine while the scanning path retained a 12-block one — exactly the divergence the upstream change exists to prevent.

No Kotlin changes. `network_id` already reaches every JNI entry point, so the interval is derived on the Rust side; a Kotlin-level knob would be a second source of truth that could drift from the grid the wallet actually retains.

## Why `scheduling_params` had to be implemented here

`MigrationBackend::scheduling_params` is a new required trait method upstream. This SDK does not use `zcash_pool_migration::wallet::WalletMigration` — `migration_engine.rs` is a deliberately thinner adapter, because `WalletMigration` requires a `UnifiedSpendingKey` unconditionally while several JNI entry points only build unsigned PCZTs. The implementation mirrors `WalletMigration`'s: it derives the delay distributions from the wallet's own retention interval, so the grid remains the single configured value.

## Dependency changes

Bumping the pin pulls in a 44-commit delta, so this also carries:

- `zcash_pool_migration_backend` renamed to `zcash_pool_migration` (now published as `0.1.0-alpha.1`)
- version requirements raised to admit the patched versions — `zcash_primitives`/`zcash_proofs` 0.29→0.30, `zcash_transparent` 0.9→0.10, `pczt` 0.8.0-rc.1→0.8. A `[patch]` entry is silently ignored unless the declared requirement admits the patched version, which is the failure mode this file's own comment warns about.
- `MigrationState::from_parts` gained an `AnchorBucketInterval` argument; the cancellation path carries the stored one through, so a rebuild keeps the grid its run was committed under
- `create_pczt_from_proposal` and `BuildConfig::Standard` take `BundlePadding` instead of `orchard::BundleType`
- the dead `zcash_encoding` patch entry is removed

## Feature builds

Two third-party pins had to move onto the 0.30 generation alongside the librustzcash bump. Both previously required `zcash_primitives = "0.29.0"`, which the patched rev does not satisfy, so cargo left them on unpatched crates.io copies and the graph split:

- `slipstream-core` → `zodl-inc/slipstream` `b9811a12` (was 43 errors)
- `zcash_voting` → `zodl-inc/zcash_voting` `e5d885c2` (was a `PcztParts` type mismatch in `build_from_parts`)

All three feature builds — default, `--features slipstream`, `--features chp-voting` — now compile, with no duplicated zcash crates. Keep these two pins in step with the patch block: a rev on the 0.29 generation cannot accept the patched librustzcash and splits the graph again.

## Verification

- `cargo check --all-targets` (default features): clean
- `cargo check --features slipstream`: clean
- `cargo check --features chp-voting`: clean
- `cargo tree -d`: no duplicated zcash crates
- `cargo fmt --check`: clean
- `git rebase --exec 'cargo fmt --check && cargo check --all-targets'`: both commits build independently
- `cargo test`: 8 passed, 1 failed, 12 ignored. The failure (`utils::exception::tests::unknown_any`) is pre-existing and unrelated — it fails identically on the unmodified tree, because on rustc 1.92 `panic!("{}", 1)` yields a `String` payload. The ignored tests require `MIGRATION_TEST_WALLET_DB`.

## Note for reviewers

An in-flight **testnet** migration committed under the 144-block grid will now fail proving with `ProveError::AnchorIntervalMismatch` and must be restarted. The upstream change records the grid a migration was committed under and refuses to prove against a different one, rather than letting it fail later as an unexplained missing checkpoint. Mainnet is unaffected.

The slipstream engine also does its own checkpoint pruning via `EngineConfig::anchor_retention_height`, which is a separate mechanism from `WalletDb`'s retention — worth a look when the slipstream rebase happens.

---
Opened with Claude Code assistance.

### DIFF
--- a/backend-lib/Cargo.toml
+++ b/backend-lib/Cargo.toml
@@ -35,27 +35,12 @@ slipstream = ["dep:slipstream-core", "dep:tokio", "dep:tracing-android", "tracin
 # voting-circuits support the old git pin provided, matching the zcash-swift-wallet-sdk pin.
 # No orchard [patch] entry is needed (and slipstream-core requires ^0.15.4, which it satisfies).
 orchard = { version = "0.15", features = ["unstable-voting-circuits"] }
-# Orchard->Ironwood migration engine (core/upstream implementation) — git dependency pinned to the
-# same librustzcash rev as the `[patch.crates-io]` block below (this crate isn't published to
-# crates.io). Replaces the former hand-rolled `zcash_pool_migration` crate. The SQLite persistence
-# side (formerly the separate `zcash_pool_migration_sqlite` crate) has since been folded into
-# `zcash_client_sqlite::pool_migration` (see that dependency's `[patch.crates-io]` entry below) —
-# no separate dependency needed for it.
-zcash_pool_migration_backend = { version = "0.1.0", features = ["wallet"] }
-pczt = { version = "0.8.0-rc.1", features = ["prover", "orchard", "sapling", "signer"] }
+zcash_pool_migration = { version = "0.1.0-rc.1", features = ["wallet"] }
+pczt = { version = "0.8", features = ["prover", "orchard", "sapling", "signer"] }
 sapling = { package = "sapling-crypto", version = "0.7", default-features = false }
-# zcash_transparent, zcash_address, zcash_client_backend, zcash_client_sqlite, zcash_primitives,
-# zcash_proofs and zcash_protocol are bumped to the minor line the convergence-pin librustzcash
-# checkout actually publishes (0.9/0.13/0.24.0-rc.1/0.22.0-rc.1/0.29/0.29/0.10 respectively) so
-# the [patch.crates-io] block below is actually used. A `[patch]` only replaces the resolved
-# crate when its version satisfies the requirement string declared here; a requirement that's
-# only compatible with the *published* generation the patch replaces makes cargo silently
-# ignore the patch for that crate and pull a second, unpatched copy from crates.io instead,
-# splitting the graph (identical failure mode already hit and documented by slipstream-jni's
-# own Cargo.toml for zcash_protocol).
-transparent = { package = "zcash_transparent", version = "0.9", default-features = false }
+transparent = { package = "zcash_transparent", version = "0.10", default-features = false }
 zcash_address = "0.13"
-zcash_client_backend = { version = "0.24.0-rc.1", features = [
+zcash_client_backend = { version = "0.24.0-rc.2", features = [
     "orchard",
     "lightwalletd-tonic-tls-webpki-roots",
     "tor",
@@ -63,10 +48,10 @@ zcash_client_backend = { version = "0.24.0-rc.1", features = [
     "unstable",
     "pczt",
 ] }
-zcash_client_sqlite = { version = "0.22.0-rc.1", features = ["orchard", "transparent-inputs", "unstable", "serde"] }
+zcash_client_sqlite = { version = "0.22.0-rc.2", features = ["orchard", "transparent-inputs", "unstable", "serde"] }
 zcash_note_encryption = "0.4.1"
-zcash_primitives = "0.29"
-zcash_proofs = "0.29"
+zcash_primitives = "0.30"
+zcash_proofs = "0.30"
 zcash_protocol = "0.10"
 zcash_script = "0.4"
 zip32 = "0.2"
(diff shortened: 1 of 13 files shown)


nuttycom (commented):
self-utACK; depends on https://github.com/zcash/librustzcash/pull/2759

nuttycom:
Draft until https://github.com/zcash/librustzcash/pull/2759 merges

nesence-m (approved):
utACK

backend-lib/Cargo.toml
```diff
@@ -37,23 +37,23 @@ slipstream = ["dep:slipstream-core", "dep:tokio", "dep:tracing-android", "tracin
 orchard = { version = "0.15", features = ["unstable-voting-circuits"] }
 # Orchard->Ironwood migration engine (core/upstream implementation) — git dependency pinned to the
 # same librustzcash rev as the `[patch.crates-io]` block below (this crate isn't published to
-# crates.io). Replaces the former hand-rolled `zcash_pool_migration` crate. The SQLite persistence
-# side (formerly the separate `zcash_pool_migration_sqlite` crate) has since been folded into
-# `zcash_client_sqlite::pool_migration` (see that dependency's `[patch.crates-io]` entry below) —
-# no separate dependency needed for it.
-zcash_pool_migration_backend = { version = "0.1.0", features = ["wallet"] }
-pczt = { version = "0.8.0-rc.1", features = ["prover", "orchard", "sapling", "signer"] }
+# crates.io). Upstream renamed it from `zcash_pool_migration_backend` to `zcash_pool_migration` and
```
dannywillems: We should remove these comments later.

backend-lib/Cargo.toml
```diff
@@ -148,6 +148,11 @@ xz2 = { version = "0.1", features = ["static"] }
 # now. Optional, gated by `chp-voting`: zcash_voting 1.0's API surface has not yet been fully
 # ported past its own rename-level drift, so a default build must not carry it. Revert
 # alongside the rest of the patch block once the librustzcash family's next real release lands.
+#
+# Like `slipstream-core` below, this rides the [patch.crates-io] rev; the pinned fork rev is on
```
dannywillems: We can remove that later.

backend-lib/Cargo.toml
```diff
@@ -167,7 +172,17 @@ ff = { version = "0.13", optional = true }
 # is git-pinned (no machine-local path); backend-lib's `[patch.crates-io]` below is the single
 # source of truth governing the librustzcash rev for `core` and the whole graph. Keep this rev
 # in step with that patch block.
-slipstream-core = { git = "https://github.com/zodl-inc/slipstream.git", rev = "3009b7772bbedd56334af0f113a543c7fe80ecc5", optional = true }
+#
+# This rev is on the 0.30 generation and takes its anchor-retention grid from
```
dannywillems: We can remove that now I think.

dannywillems (approved):
utACK: only affecting testnets.

backend-lib/src/main/rust/migration.rs
```diff
@@ -1851,6 +1865,12 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_MigrationRustBackend_
             state.note_split().clone(),
             state.preparation().clone(),
             new_transactions,
+            match network {
```
dannywillems, line 1868: Nit: could be extracted.

dannywillems (approved):
utACK

PR #2046, opened 2026-07-25
Move backend-lib onto the published post-RC librustzcash crates

dannywillems:
## Summary

`backend-lib` depended on the frozen crates.io release candidates
`zcash_client_backend 0.24.0-rc.1` and `zcash_client_sqlite 0.22.0-rc.1`,
which predate the breaking send-max and Ironwood API changes. Those changes
have since shipped in the published `0.24.0-rc.2` / `0.22.0-rc.2` release
candidates, so this moves the backend onto the published crates and adapts the
JNI glue. Send-max and Ironwood support keep working; no JNI-visible signatures
change.

The work is in three commits:

1. **Use the published crates** - bump `zcash_client_backend` and
   `zcash_client_sqlite` from `rc.1` to `rc.2`, keep the aligned `pczt 0.8.0` /
   `zcash_transparent 0.10` / `zcash_primitives`/`zcash_proofs 0.30`, and drop
   the temporary `[patch.crates-io]` git pin. The committed `Cargo.lock` now
   resolves every former git-sourced crate from crates.io, with no incidental
   transitive churn.

2. **Adapt the JNI backend** to the parameters added after the RCs:
   - `propose_send_max_transfer` / `propose_transfer`: locked-input policy
     and/or input-lock request
   - `propose_shielding`: input-lock request
   - `create_proposed_transactions`: expiry height
   - `create_pczt_from_proposal`: expiry height + `BundlePadding` (was
     `BundleType`)
   - `get_spendable_transparent_outputs`: lock filter
   - Tor `connect_to_lightwalletd`: allow-onion-services flag

   Every new argument is a named `let` binding passing the value that preserves
   prior behavior: no input locking, default padding, and clearnet-only
   lightwalletd connections.

3. **Ignore** the `.claude` local config and agent worktrees.

## Testing

- `cargo check --lib` against the published `rc.2` crates: clean.
- `./scripts/ci-local.sh quick` (detekt, ktlint, Kotlin compile, native-lib
  cross-build, unit tests): passed. The JNI glue is unchanged from that run;
  only the crate versions moved `rc.1` -> `rc.2`.

## Note

`0.24.0-rc.2` / `0.22.0-rc.2` are the latest published librustzcash releases and
carry the post-`rc.1` send-max and Ironwood APIs. Bump to the final `0.24.0` /
`0.22.0` once those are cut.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Updated the librustzcash crates to their published releases,
+  `zcash_client_backend 0.24.0-rc.2` and `zcash_client_sqlite 0.22.0-rc.2`.
+
 ## [2.5.2] - 2026-06-03
 
 ### Changed
--- a/backend-lib/Cargo.toml
+++ b/backend-lib/Cargo.toml
@@ -13,11 +13,11 @@ rust-version = "1.92"
 [dependencies]
 # Zcash dependencies
 orchard = "0.15"
-pczt = { version = "0.8.0-rc.1", features = ["prover", "orchard", "sapling"] }
+pczt = { version = "0.8.0", features = ["prover", "orchard", "sapling"] }
 sapling = { package = "sapling-crypto", version = "0.7", default-features = false }
-transparent = { package = "zcash_transparent", version = "0.9", default-features = false }
+transparent = { package = "zcash_transparent", version = "0.10", default-features = false }
 zcash_address = "0.13"
-zcash_client_backend = { version = "0.24.0-rc.1", features = [
+zcash_client_backend = { version = "0.24.0-rc.2", features = [
     "orchard",
     "lightwalletd-tonic-tls-webpki-roots",
     "tor",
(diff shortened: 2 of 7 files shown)


backend-lib/Cargo.toml
```diff
@@ -77,6 +77,22 @@ xz2 = { version = "0.1", features = ["static"] }
 component = "0.1.1"
 rust-analyzer = "0.0.1"
 
+# Pin the librustzcash crates to the `dw/remove-orchard-pin` branch, which is
```
dannywillems: Remove when the release has been made. Depends on https://github.com/zcash/librustzcash/pull/2764

dannywillems: Crates published. Removing patches.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Updated the librustzcash crates to their published releases, replacing the
```
dannywillems: Changing this. Should not be mentioning `dw/remove-orchard-pin` which is from an old commit.

backend-lib/src/main/rust/tor.rs
```diff
@@ -91,7 +91,14 @@ impl TorRuntime {
     pub(crate) fn connect_to_lightwalletd(&self, endpoint: Uri) -> anyhow::Result<LwdConn> {
         let Self { runtime, client } = self.isolated_client();
 
-        let conn = runtime.block_on(async { client.connect_to_lightwalletd(endpoint).await })?;
+        // lightwalletd endpoints are ordinary clearnet hosts, not Tor onion
+        // services, so onion connections are not permitted for them.
```
nuttycom, line 95: This overstates the case but it's fine for now - we need to allow .onion addresses to be configured, then this can be switched.

nuttycom (approved):
utACK 2a9abba329a30fb5b1c304045815e81efa449ad1

PR #2048, opened 2026-07-25
Reconcile maint/v2.5.x and the 2.6.x release line into main

nuttycom:
Reconciles the two maintenance lines into `main`. Opened as a draft for review of the
integration itself and, especially, the conflict resolutions.

Fast-forward only: `upstream/main` is an ancestor of this branch, so merging cannot
introduce a new merge commit.

## What this merges

    c7345450  Merge maint/v2.6.x into main
    554826fd  Merge maint/v2.5.x into main

`maint/v2.6.x` was cut at `release/snapshot-v2.6.5` (`6949ea93`) and pushed separately;
it mirrors how `maint/v2.5.x` sits on the 2.5 release line.

Two assumptions worth stating, because they inverted the original plan for this work:
`release/v2.6.0` was already an ancestor of `main` (merging it would have been a no-op),
and `main` was *behind* the 2.6.x release line rather than ahead of it — `main` sat at
`2.6.4-SNAPSHOT` while 2.6.5 had shipped.

## Conflict resolutions

Resolved hunk by hunk rather than file by file, so neither side's cleanly-merged work was
discarded. Reviewers may want to focus here:

- **`gradle.properties` / `Cargo.toml` version** — kept `main`'s 2.6.5 over the 2.5.2
  maintenance line.
- **`backend-lib/Cargo.toml` dependencies** — took `maint/v2.5.x` (the librustzcash 0.24
  generation), which is the point of the merge.
- **`.github/workflows/pull-request.yml`** — kept `main`'s newer `actions/checkout` 7.0.0
  pin and its `rust-cache: true` option, both absent on 2.5.x.
- **`CHANGELOG.md`** — 2.5.x's unreleased Ironwood / 0.24 entries folded into `main`'s
  `[Unreleased]`; the released `[2.5.2]` section placed in version order.
- **`backend-lib/Cargo.lock`** — regenerated.

The Ironwood series exists on both branches as patch-equivalent cherry-picks, so `git`
drops the duplicates rather than applying them twice — which is why the textual conflict
count is low.

## Three fixes with no conflict to mark them

Each is an interface or type that `maint/v2.5.x` extends and `main` independently has a user
of. Neither branch carries both halves, so the textual merge is clean and **does not compile**.
This is the defect class to watch for in this integration generally — `main` added to interfaces
and protos, and users living only on the other branch are invisible to a textual merge.

- **`OutboundTransactionManager`** gains `proposeOrchardToIronwoodMigration` on 2.5.x, while
  `main` adds `SdkBroadcasterTest`'s `FakeOutboundTransactionManager` implementing it. The fake
  gains the override, following its own convention of `error("Unused")`.
- **The NU6.3 `TreeState` proto** gains an `ironwood_tree` field, while `main`'s shielded-voting
  fixtures construct `TreeState` literally. The fixture sets it empty, as it already does for the
  Sapling tree it does not exercise.
- **`pczt` 0.7 -> 0.8** makes `Pczt::serialize` return `Result`, while `main`'s delegation test
  passes its output straight to a `&[u8]` parameter.

The last two are behind `#[cfg(feature = "android-test-fixtures")]`, which
`backend-lib/build.gradle.kts` enables for the Android build and CI enables for clippy — so a
default-feature `cargo check --lib` does not compile them, and they surfaced only in CI.

## Shielded voting

Left enabled. `zcash_voting` 0.11 still requires `orchard` 0.14 while the rest of the backend
now resolves `orchard` 0.15, so the dependency graph carries both (and both `zcash_primitives`
0.28 and 0.30). This compiles because the voting JNI surface only uses
`orchard::primitives::redpallas` and never passes orchard types across the `zcash_voting`
boundary. Collapsing onto a single `orchard` needs `zcash_voting` 1.0 and is tracked separately.

## Verification

| Check | Result |
| --- | --- |
| `./gradlew ktlint` | pass |
| `./gradlew detektAll` | pass |
| `./gradlew :sdk-lib:test :sdk-incubator-lib:test` | pass — 195 tests, 0 failures |

Not run locally: Android Lint, instrumented/device tests, and `assembleRelease` — these run in CI.
`LIBRARY_VERSION` is still `2.6.5`; any version bump is separate.

---


dannywillems (dismissed):
utACK. Thank you 🙏

dannywillems:
I looked at the commit history, and it seemed okay. But clippy is unhappy.

nuttycom (commented):
self-ACK

dannywillems (approved):
utACK

PR #2049, opened 2026-07-25
Ironwood/Slipstream: migration engine, Slipstream sync, and main integration

nuttycom:
Replaces `android-slipstream-ironwood-chp` as the integration branch for the Ironwood/Slipstream
work. Opened as a draft against `spec/main` (#2048) so the diff shows only this branch's work
rather than re-showing the maintenance-line reconciliation. GitHub retargets this to `main`
automatically when #2048 merges.

Most of the 138 files here are pre-existing work carried over from
`android-slipstream-ironwood-chp` — the ZIP 318 Orchard→Ironwood migration engine, the Slipstream
sync engine, the Keystone signing path, and the anchor-bucket interval configuration from #2042.
Those have been reviewed on their own PRs. **The four commits worth reviewing here are the ones
this branch adds on top:**

    3a7644e6  CI: run the Rust unit tests instead of only compiling them
    306ce344  backend-lib: fix unreachable assertion in unknown_any test
    3d4e2208  Satisfy clippy on the Slipstream and migration code
    a430e6dc  Merge main into the Ironwood/Slipstream feature branch

## Both migration APIs are preserved, deliberately

They are different products, not duplicates:

- **One-shot** `proposeOrchardToIronwoodMigration` — the cheap, privacy-discarding crossing. A
  single transaction whose value is the account's entire Orchard balance, readable as such by any
  chain observer. Its own doc comment says so.
- **Scheduled** `MigrationSdk` — the denomination-splitting engine.

SDK consumers choose. Because this branch's `migration.rs` is the engine's JNI layer, `main`'s
one-shot module lands beside it as `migration_send_max.rs`, named for the `propose_send_max_transfer`
capability it pins, and `lib.rs` gains the `RustBackend_proposeOrchardToIronwoodMigration` export
pointing at it. The JNI symbol is confirmed present in the linked `.so` for all four ABIs.

## The defect class this merge kept hitting

`main` extends interfaces and protos; this branch independently has implementors and users of them.
Neither branch carries both halves, so **git produces no conflict and the tree does not compile**.
Four instances, none of which a textual merge could have flagged:

| Extended on main | User only on the other branch |
| --- | --- |
| `Synchronizer.proposeOrchardToIronwoodMigration` | `SlipstreamSynchronizer` |
| `OutboundTransactionManager.proposeOrchardToIronwoodMigration` | `FakeOutboundTransactionManager` (fixed in #2048) |
| `TreeState.ironwood_tree` (NU6.3 proto) | voting fixtures (fixed in #2048) |
| `Pczt::serialize` → `Result` (pczt 0.8) | delegation test (fixed in #2048) |

`SlipstreamSynchronizer` now delegates to `SlipstreamSpendService`, which delegates to `Backend` and
wraps with `Proposal.fromUnsafe` — matching every sibling `propose*` method in that file. Worth
assuming more of these on any future `main` → feature merge.

## Conflict resolutions

Resolved hunk by hunk rather than file by file, so neither side's cleanly-merged work was discarded.
The non-obvious calls:

- **`tor.rs`** — kept this branch's `allow_onion_services` host derivation. `main` hardcodes `false`,
  which would drop support for onion lightwalletd servers.
- **`CompactBlockProcessor`** — took *main's* `fetchSubtreeRoots` helper, which collapses three
  copy-pasted per-pool retry loops and is already Ironwood-aware.
- **`lib.rs`** — kept this branch's version. main's additions there are its rc.2 builder-API
  adaptations, which #2042 already provides in this branch's own form, plus a
  `LockFilter::Policy(&LockedInputPolicy::Exclude)` whose justifying comment ("this backend does not
  use input locking") is false here — this branch *does* lock, via `lockRemainingOrchardBalanceNative`.
- **`voting/util.rs`** — kept this branch's `tree_state_fixture`, which populates `ironwood_tree`
  rather than `orchard_tree` deliberately: voting notes are Ironwood/V3 notes, so that is the field
  the fixtures actually read.
- **CI workflows** — kept this branch's `persist-credentials: false` hardening but carried main's
  newer `actions/checkout` 7.0.0 pin onto it, unifying every workflow onto one pin.

## Clippy and the Rust tests

Merging `main` brings its `static_analysis_clippy` job onto this branch for the first time. It
reported ten findings in this branch's own code — not merge fallout, just never checked before. Six
were applied via `cargo clippy --fix` (`slipstream/mod.rs` is a pure 70-line block move). Four needed
a judgement call: a `CommitContext` struct for `commit_or_reuse`'s five ambient arguments (all eight
call sites updated), plus `MigrationCommitOutcome` and `SignedBatchPczts` type aliases.

Separately, nothing ran the Rust unit tests. `static_analysis_clippy` compiles them via `--tests` but
never executes them, so `utils::exception`'s `unknown_any` had been failing indefinitely — its helper
panics via `panic!("{}", val)`, which always produces a `String` payload, making the fallback branch
it asserts on unreachable. Fixed with `panic::panic_any`, and a `test_rust_unit` job now runs
`cargo test --all-features` so this cannot recur. Both cherry-picked from `dw/run-rust-tests-in-ci`
rather than merged — that branch predates changes here and merging it would revert the checkout pin
and drop the shellcheck job.

## Verification

| Check | Result |
| --- | --- |
| `cargo test --all-features` | pass — 30 tests, 0 failures |
| `./gradlew ktlint detektAll` | pass |
| `./gradlew :sdk-lib:test :sdk-incubator-lib:test` | pass — 651 tests, 0 failures |
| `./gradlew :sdk-lib:assembleRelease` (+ 3 library modules) | pass |
| native link, all four ABIs | pass — JNI export verified in each `.so` |
| shellcheck | pass |

Not verified locally: Android Lint, instrumented/device tests, and the demo-app release path (its
signing keystore is not available outside CI). `LIBRARY_VERSION` is still `2.6.5`; the 3.0.0 bump is
separate.

Shielded voting stays gated off here behind `chp-voting`, so `VotingRustBackend`'s natives are absent
from a default build — unchanged from this branch's existing behaviour and intentional.

---


nuttycom:
## Adoption note: zcash/librustzcash#2892 (prove-time wallet persistence)

When this branch's `zcash_pool_migration` pin advances past zcash/librustzcash#2892 (currently on the `feat/migration_unsatisfiability` line), the prove seam changes shape. Summary for the engine JNI layer (`migration.rs`) and product code:

### API changes to adopt

- `ProveOutcome::Proved` now carries a `#[must_use]` `ProvedTransaction` (the proven PCZT bytes + row id), and `prove_transfer` / `prove_preparation` no longer move the state to `Proved` themselves. The only discharge path is the new required `PoolMigrationWrite::store_proved_transaction`:

  ```rust
  match engine::prove_transfer(&params, &mut prover, &mut state, id, scanned_tip, &mut rng)? {
      ProveOutcome::Proved(proven) => {
          // The prover's wallet borrow must end before the store is used — the proof
          // travels as a value precisely so the two borrows stay sequential.
          store.store_proved_transaction(&mut state, proven)?;
      }
      outcome => {
          // NotYetProvable / MarkedUnsatisfiable: persist via replace_migration, as today.
          store.replace_migration(&state)?;
      }
  }
  ```

- `zcash_client_sqlite`'s store constructor gains wallet context: `PoolMigrations::for_account(params, clock, conn, account)` (pass the network parameters and `zcash_client_sqlite::util::SystemClock`). The `PoolMigrationWrite` impl requires real `Parameters`/`Clock` values; `()` is accepted only for read-only use.

- The one-shot `proposeOrchardToIronwoodMigration` path is unaffected: it goes through the standard proposal/build flow, which has always persisted at build time. Only the scheduled `MigrationSdk` engine path changes.

### Behavioral changes product code should expect

- From the moment a transaction is proved, the wallet database holds its finalized record — raw transaction, fee, sent outputs, and input notes marked spent — written atomically with the migration store. Spendable balance therefore drops at *prove* time rather than at mine time, and the wallet's own spends can no longer consume a migration input during the (deliberately long) window before the scheduled broadcast. Balance/note-selection expectations in tests and UI should be checked against this.
- Broadcast-only sessions no longer need to extract the transaction from the stored PCZT: the raw transaction is available from the wallet DB by txid from prove time onward. Relatedly, the submit-crash window in `reconcileMigrationInvalidations`-style probes narrows: the wallet's own transaction table now already knows the txid and raw bytes before any submission attempt.
- Proving surfaces extraction problems early: `store_proved_transaction` re-verifies proofs and signatures while finalizing, so a PCZT that would not survive broadcast fails at the prove step rather than at submission.
- **Do not auto-rebroadcast**: proved migration transfers now sit in the wallet DB as stored, unmined, deliberately not-yet-broadcast transactions, possibly for days. Any generic resubmission logic that treats "stored unmined sent transaction" as "submit now" must exclude them (e.g. skip txids present in the migration store) or it will front-run the ZIP 318 schedule and collapse the timing privacy. See the follow-up note on zcash/librustzcash#2891.

---

PR #2051, opened 2026-07-25
Merge maint/v2.5.x into maint/v2.6.x

nuttycom:
Brings the librustzcash 0.24 generation (`zcash_client_backend` 0.24, `zcash_client_sqlite` 0.22,
`zcash_primitives` 0.30, `orchard` 0.15) and Ironwood (NU6.3) pool support onto the 2.6 maintenance
line.

Opened as a PR because `maint/v2.6.x` requires one; the branch is otherwise just that branch plus
these two commits, and merges fast-forward.

Five textual conflicts and three no-conflict compile breaks were resolved. The result builds and all
tests pass — see Verification below.

    ec8792c2  backend-lib: fix unreachable assertion in unknown_any test
    18828168  Merge maint/v2.5.x into maint/v2.6.x

## Conflict resolutions

- **`gradle.properties` / `Cargo.toml` version** — kept 2.6.5, ahead of the 2.5.2 line.
- **`backend-lib/Cargo.toml` dependencies** — took `maint/v2.5.x`, which is the point of the merge.
  `zcash_voting` stays at `=0.11.0`, so the graph carries both `orchard` 0.14 (for voting) and 0.15
  (everything else), and both `zcash_primitives` 0.28 and 0.30. That resolves and compiles: the
  voting JNI surface only uses `orchard::primitives::redpallas` and never passes orchard types
  across the `zcash_voting` boundary.
- **`.github/workflows/pull-request.yml`** — took 2.5.x's `test_android_modules_emulator` job, which
  replaces the emulator.wtf cloud job with an in-runner Gradle Managed Device, including its KVM
  enablement and emulator-install steps. Kept 2.6.x's newer `actions/checkout` 6.0.3 pin (2.5.x
  still had 6.0.2) and its `rust-cache: true` setup input, and dropped the now-dead emulator.wtf
  build step and its API-key env. All ten checkout invocations end up on one pin, and the resulting
  job set matches 2.5.x's exactly.
- **`CHANGELOG.md`** — 2.5.x's unreleased Ironwood / 0.24 entries into `[Unreleased]`; the released
  `[2.5.2]` section placed in version order below the 2.6.x sections.
- **`backend-lib/Cargo.lock`** — regenerated.

## Three breaks git could not flag — found and fixed here

These are the substantive part of the merge. Each is an interface or type that `maint/v2.5.x`
extends and `maint/v2.6.x` independently has a user of. Neither branch carries both halves, so git
reports **no conflict** and the naive merge result does not compile. All three are fixed in this PR;
the tree builds and the tests pass.

| Extended by 2.5.x | User only on 2.6.x | Fix applied |
| --- | --- | --- |
| `OutboundTransactionManager.proposeOrchardToIronwoodMigration` | `SdkBroadcasterTest`'s `FakeOutboundTransactionManager` | added the override as `error("Unused")`, matching the fake's existing convention for methods it never exercises |
| `TreeState.ironwood_tree` (NU6.3 proto) | shielded-voting `tree_state_fixture` | set to `String::new()`, as the fixture already does for the Sapling tree it does not exercise |
| `Pczt::serialize` now returns `Result` (pczt 0.7 → 0.8) | delegation test passing it straight to a `&[u8]` parameter | `.expect("serialize signed PCZT")`, matching the `.expect(...)` style of the surrounding test lines |

The latter two live behind `#[cfg(feature = "android-test-fixtures")]`, which
`backend-lib/build.gradle.kts` enables for the Android build. A default-feature `cargo check --lib`
does not compile them, so they surface only under the Android build or `--all-features` — worth
knowing for anyone merging into this line again.

## The `unknown_any` fix

Cherry-picked from `dw/run-rust-tests-in-ci`. Pre-existing on `maint/v2.6.x`, unrelated to the merge:
the helper panics via `panic!("{}", val)`, which always produces a `String` payload, so
`any_to_string`'s fallback branch the test asserts on was unreachable. Nothing caught it because this
line has no job that runs `cargo test`.

## Verification

| Check | Result |
| --- | --- |
| `cargo test --all-features` | pass — 18 tests, 0 failures (was 17 passed / 1 failed before the `unknown_any` fix) |
| `./gradlew ktlint` | pass |
| `./gradlew :sdk-lib:test :sdk-incubator-lib:test :backend-lib:test` | pass — 243 tests, 0 failures |

Two things deliberately left alone, both pre-existing and non-gating on this line, which has neither
a clippy nor a `cargo test` job:

- Six clippy findings in `eip681.rs` and `voting/delegation.rs` (three `needless_borrow`, two
  `unnecessary to_vec`, one 10-argument function).
- `detektAll` is clean on this tree. If you run it in a working directory containing nested
  worktrees, it globs those too and reports their findings as if they were this branch's.

---


dannywillems (dismissed):
utACK

pacu (approved):
**Non-Blocking comments**

1. This is not a patch-compatible merge, and [Unreleased] doesn't say so. AccountBalance is a public data class gaining a required ironwood: WalletBalance in the middle of its parameter list — source- and binary-breaking for positional construction, copy(), and destructuring. Synchronizer gains an abstract method; ZcashProtocol gains IRONWOOD, breaking exhaustive when. The next release off this line should be 2.7.0, not 2.6.6. Worth stating in the changelog.
For you specifically: zodl-android pins ZCASH_SDK_VERSION=2.6.5-SNAPSHOT and has no AccountBalance(...) construction or ZcashProtocol reference, so nothing breaks at the source level — but it also won't surface Ironwood balances until it reads the new field.

2. The changelog calls release candidates "published releases." crates.io max-stable is zcash_client_backend 0.23.0 and zcash_client_sqlite 0.21.1; this pins 0.24.0-rc.2 / 0.22.0-rc.2. Both the commit title ("Use the published librustzcash crates past the 2.5.x RCs") and the changelog line are misleading. The substantive part: an RC zcash_client_sqlite means the data-DB migration set isn't final, and anything shipped from this line migrates real wallet DBs on a release-candidate schema. That call was made on 2.5.x, but this is the PR that puts it on the line zodl-android consumes. (The requirement is ^0.24.0-rc.2, so it will pick up 0.24.0 final automatically — that part is fine.)

3. The body understates the duplicated graph. It names orchard 0.14/0.15 and zcash_primitives 0.28/0.30. The lock actually carries six duplicated Zcash crates — also zcash_address 0.12+0.13, zcash_protocol 0.9+0.10, zcash_transparent 0.8+0.10, pczt 0.7+0.8. Mitigating: halo2_proofs and sapling-crypto remain single-version, which caps the damage. Still worth measuring the .so size before release.

4. The voting-boundary claim is inaccurate as written. "the voting JNI surface only uses orchard::primitives::redpallas" — it also uses orchard::keys::Scope, orchard::note::{Note, ExtractedNoteCommitment}, orchard::tree::MerkleHashOrchard, and orchard::NOTE_COMMITMENT_TREE_DEPTH. The conclusion still holds — I checked the zcash_voting boundary and it's byte-oriented (fvk_bytes, hotkey_raw_address, raw note data), so no orchard type crosses — but the sentence should be corrected, because the real invariant ("orchard types never cross; witness/FVK data crosses as bytes") is stronger and is what a future merger needs. It also names the latent risk: MerkleHashOrchard from 0.15 produces tree data that voting-circuits reconstructs via 0.14. Compiles either way; only the voting instrumentation tests exercise it.

5. Dead emulator.wtf gate (pre-existing on 2.5.x — I verified it's inherited verbatim, not a resolution error). The wtf job is gone but check_emulator_wtf_secrets remains, and test_android_modules_ftl is still gated on ...has-secrets == 'false' under the now-false comment "Emulator.wtf is preferred if it has an API key." Net effect: if EMULATOR_WTF_API_KEY is still set on the repo, FTL silently never runs; if it isn't, FTL and the new managed-device job duplicate coverage. Cheap to fix while you're in this file.

6. Ironwood subtree-root fetch degrades silently against a non-Ironwood lightwalletd. retryUpToAndContinue swallows after N retries, and result is then overwritten by the saplingSubtreeRootList.isNotEmpty() block — so an upstream lightwalletd answering "unrecognized shielded protocol" costs N retries with exponential backoff on every spend-before-sync init, then proceeds with an empty Ironwood list. Identical to the pre-existing Orchard behavior, so the refactor isn't at fault. Your own lightwalletd fork already handles ShieldedProtocol_ironwood with explicit fail-open/fail-closed logic, so your infra is covered.

7. Nit: the body says "just that branch plus these two commits" but there are now three — 097962a1 (raise warm_proving_caches_smoke to a 5-minute runTest timeout for the slower in-runner emulator) landed after the body was written. The commit itself is well-justified; just update the description.

Issue #1405, opened 2024-02-29
Fix/Remove broken `SdkSynchronizer.validateConsensusBranch` function


HonzaR:
## Is your feature request related to a problem? Please describe.
This function works internally with `ConsensusBranchId` that contains values that are not correct or should not be used in the current SDK state.

Issue #679, opened 2022-08-19
ConsensusBranchId - improve implementation


HonzaR:
## Is your feature request related to a problem? Please describe.
Filling an additional issue to provide a link to the code base TODO occurrence.

We'd like to check if we can find a way to not rely on branch id definitions as a separate source of truth in `ConsensusBranchId` (either stop converting from hex to display name in the apps or use Rust to get this info).

PR #2053, opened 2026-07-25
Expose mismatch details on server-compatibility exceptions

pacu:
## What

`MismatchedConsensusBranch`, `MismatchedNetwork` and `MismatchedSaplingActivationHeight` take their mismatched values as plain constructor params, so the values exist only inside the exception's message string. Promote them to public `val`s.

```kotlin
class MismatchedConsensusBranch(
    val clientBranchId: String,
    val serverBranchId: String
) : CompactBlockProcessorException(...)
```

## Why

A wallet that wants to tell the user *why* a server is incompatible currently has two bad options: regex the IDs out of the message (the wording is not an API contract, so it would silently stop matching), or show the raw English sentence verbatim in every locale.

This came from a real report: a server on NU6.3/Ironwood (`0x37a5165b`) against a client expecting NU6.2 (`0x5437f330`). Sync can never progress, and the remedy is a decision the user has to make — update, or switch servers — so the specifics are worth showing them.

The KDoc is explicit that branch IDs are opaque unordered constants: neither the SDK nor a consumer can tell from them which side is stale. In the observed case the *client* was behind, not the server.

## Compatibility

Source and binary compatible. Constructor signatures are unchanged; the classes gain getters. All existing call sites (`CompactBlockProcessor.kt`, `SdkSynchronizer.kt`) are untouched.

## Scope

Deliberately does **not** touch `Synchronizer.validateConsensusBranch()` (#1405) or the `ConsensusBranchId` enum (#679, hardcoded and stopping at CANOPY, so `fromHex("37a5165b")` returns null). Those are separate decisions.

## Target branch

Targeting `maint/v2.5.x` per the merge-forward convention — oldest maintenance branch that should carry the fix, then merged forward. Verified that `Exceptions.kt` cherry-picks onto `main` with no conflict (only `CHANGELOG.md` conflicts, as expected).

## Verification

`./scripts/ci-local.sh quick` — **passing**, exit code 0, "all requested stages passed":

```
==> [1/6] detekt (static_analysis_detekt)      BUILD SUCCESSFUL
==> [2/6] ktlint (static_analysis_ktlint)      BUILD SUCCESSFUL
> Task :sdk-lib:testDebugUnitTest
> Task :sdk-lib:testReleaseUnitTest
> Task :sdk-lib:test
BUILD SUCCESSFUL in 9m 34s
483 actionable tasks: 160 executed, 38 from cache, 285 up-to-date
```

Not run: `full` (androidTest on a managed device) and the `emulator.wtf` / Firebase Test Lab cloud stages. Per AGENTS.md's table this change is "logic or refactor" tier, for which `quick` is the stated minimum.

### Note for other macOS contributors

The Rust link step fails on a clean macOS machine with `linker-wrapper.sh: line 4: python: command not found`, because macOS ships only `python3`. `docs/Setup.md:34` covers it — set `rust.pythonCommand` in `local.properties`. Unrelated to this change, but it's the first thing you'll hit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Synchronizer.proposeOrchardToIronwoodMigration`, which builds a proposal that
   moves the account's entire Orchard balance across the NU6.3 turnstile into the
   Ironwood pool.
+- `CompactBlockProcessorException.MismatchedConsensusBranch`, `MismatchedNetwork`
+  and `MismatchedSaplingActivationHeight` now expose their constructor arguments
+  as public `val`s (`clientBranchId`/`serverBranchId`,
+  `clientNetwork`/`serverNetwork`, `clientHeight`/`serverHeight`). Previously the
+  mismatched values were reachable only by parsing the exception's `message`, so
+  consumers had to either scrape English prose or surface it verbatim. Wallets
+  can now render a localized, structured explanation of why a server is
+  incompatible. Note that consensus branch IDs are opaque unordered constants:
+  neither the SDK nor a consumer can infer from them alone which side is stale.
 
 ### Changed
 - Migrated to the `zcash_client_backend 0.24` / `zcash_client_sqlite 0.22` API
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK

PR #2059, opened 2026-07-25
Release zcash-android-wallet-sdk v2.7.0-rc.1

nuttycom:
> **Note**

# Author

# Reviewer


pacu (commented):
Reviewed the full diff (checkpoints and `Cargo.lock` skipped). Overall this reads well — the commenting discipline is unusually good, and I verified the things it claims rather than taking them on faith:

- **Proto field numbers are all correct** against `zcash/lightwallet-protocol`: `CompactTx.ironwoodActions = 9`, `TreeState.ironwoodTree = 7`, `ShieldedProtocol.ironwood = 2`, `ChainMetadata.ironwoodCommitmentTreeSize = 3`. ✅
- `IS_SNAPSHOT=true` alongside an rc version initially looked wrong to me; it matches the v2.5.2 and v2.1.0 tags, so I withdrew that.

Comments are inline. Four of them are questions I could not settle from outside the code — most importantly the `ValuePool::Orchard` one in the PCZT proving path, which may well be correct today but appears correct *by coincidence*.

Leaving this as a comment rather than a change request; the calls are yours.

One piece of praise worth stating explicitly: the `JniAccountBalanceFixture` switch to distinct values is exactly the right instinct. Ten positional `jlong`s under `([BJJJJJJJJJJ)V` with only non-negativity checks means identical fixtures would let *any* permutation of the ten fields pass — including reporting Orchard funds as Ironwood. Same for `panic_any_error`, which fixes a test that never reached the branch it claimed to cover.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2485,11 +2591,21 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_addProofs
 
         let pczt = parse_pczt(env, pczt)?;
 
+        // The Orchard circuit version is fixed by the consensus branch under which the
+        // transaction will be mined; derive it from the branch id carried by the PCZT.
+        let orchard_circuit_version = BranchId::try_from(*pczt.global().consensus_branch_id())
+            .ok()
+            .and_then(|branch_id| bundle_version_for_branch(branch_id, orchard::ValuePool::Orchard))
```
pacu, line 2608: The pool is hardcoded to `Orchard` here, but the PCZT being proved may hold an **Ironwood** bundle post-NU6.3.

`BundleVersion` is documented as *"a shielded ValuePool (Orchard or Ironwood) together with the ProtocolVersion under which a bundle in that pool is built"* — so this is definitively the Orchard-pool bundle version, and its `circuit_version()` then builds the proving key for whatever bundle the PCZT actually contains.

From the `zcash_primitives` docs, at NU6.3+ the Orchard pool maps to the post-NU6.3 circuit with cross-address *disabled*, and Ironwood to the post-NU6.3 circuit with cross-address *enabled*. I could not determine whether `circuit_version()` collapses that distinction — docs.rs 404s on `BundleVersion` itself.

If it does collapse, this is correct today but correct by coincidence, and silently becomes wrong if the two pools' circuit versions ever diverge. If it doesn't, the Keystone/PCZT signing path proves against the wrong circuit once Ironwood is live.

Either way, deriving the pool from the bundle being proved rather than pinning it would remove the question. Note the pre-NU6.3 case is safe regardless, since `bundle_version_for_branch` returns `None` for Ironwood there.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -1327,132 +1337,92 @@ class CompactBlockProcessor internal constructor(
...
         }
 
-        // Intentionally omitting [orchardSubtreeRootList], e.g., for Mainnet usage, we could check it, but on
-        // custom networks without NU5 activation, it wouldn't work. If the Orchard subtree roots are empty, it's
-        // technically still ok (as Orchard activates after Sapling, so on a network that doesn't have NU5
-        // activated, this would behave correctly). In contrast, if the Sapling subtree roots are empty, we
-        // cannot do SbS at all.
+        val saplingSubtreeRootList = fetchSubtreeRoots(saplingStartIndex, ShieldedProtocolEnum.SAPLING)
+        val orchardSubtreeRootList = fetchSubtreeRoots(orchardStartIndex, ShieldedProtocolEnum.ORCHARD)
+        val ironwoodSubtreeRootList = fetchSubtreeRoots(ironwoodStartIndex, ShieldedProtocolEnum.IRONWOOD)
```
pacu: This fetches Ironwood subtree roots unconditionally on every `start()`, including against lightwalletd instances that predate NU6.3 support.

When that request fails, the shared helper burns all `GET_SUBTREE_ROOTS_RETRIES` retries and sets `result` to `FailureConnection`/`OtherFailure` — which the trailing `if (saplingSubtreeRootList.isNotEmpty())` block then overwrites, so the failure is masked rather than surfaced. Correct outcome, but it costs a full retry round of latency on every sync start until the entire server population is upgraded.

Worth gating the Ironwood fetch on the NU6.3 activation height, in the same spirit as the existing comment about NU5 and custom networks. This matters disproportionately for wallets whose users point at arbitrary self-selected servers.

.github/workflows/pull-request.yml
```diff
@@ -275,26 +323,63 @@ jobs:
           name: Test Android modules with FTL results
           path: ~/artifacts
 
-  test_android_modules_wtf:
-    if: needs.check_emulator_wtf_secrets.outputs.has-secrets == 'true'
-    needs: [ validate_gradle_wrapper, check_emulator_wtf_secrets ]
+  # Runs the Android instrumentation suites on a Gradle Managed Device: a
+  # hardware-accelerated emulator booted inside the runner. This replaces the
+  # former emulator.wtf cloud job and depends on no external test service.
+  test_android_modules_emulator:
```
pacu, line 361: Removing the emulator.wtf job leaves live scaffolding behind, and one piece of it silently disables a different job.

`check_emulator_wtf_secrets` is still defined (line 50), and the Firebase Test Lab job still gates on it (line 271):

```yaml
if: needs.check_firebase_secrets.outputs.has-secrets == 'true' && needs.check_emulator_wtf_secrets.outputs.has-secrets == 'false'
```

So on any repo or fork where the `EMULATOR_WTF_API_KEY` secret is still configured, **FTL now skips itself in deference to a job that no longer exists** — instrumentation coverage disappears with no failure and no warning. The comment above it ("Emulator.wtf is preferred if it has an API key") no longer describes reality.

Also still present on this branch: `ZCASH_EMULATOR_WTF_API_KEY` and `EMULATOR_WTF_GRADLE_PLUGIN_VERSION` in `gradle.properties`, the `wtf.emulator` plugin and maven repo in `settings.gradle.kts`, `build.gradle.kts:54`, and `build-conventions/src/main/kotlin/zcash-sdk.emulator-wtf-conventions.gradle.kts`.

CHANGELOG.md
```diff
@@ -6,6 +6,21 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  (balance, subtree roots, sync) alongside Sapling and Orchard.
+- `Synchronizer.proposeOrchardToIronwoodMigration`, which builds a proposal that
+  moves the account's entire Orchard balance across the NU6.3 turnstile into the
+  Ironwood pool.
+
+### Changed
+- Migrated to the `zcash_client_backend 0.24` / `zcash_client_sqlite 0.22` API
+  line, adapting the backend to the send-max and builder API changes.
+- Updated the librustzcash crates to their published releases,
+  `zcash_client_backend 0.24.0-rc.2` and `zcash_client_sqlite 0.22.0-rc.2`.
```
pacu, line 70: This release breaks source compatibility for consumers, and the changelog doesn't say so. From a downstream wallet's perspective the following all require code changes:

- `AccountBalance` gains a required `ironwood` field **in the middle** of the constructor — positional construction silently rebinds arguments or fails to compile.
- `TransactionPool` and `ShieldedProtocolEnum` gain `IRONWOOD`, so exhaustive `when` blocks downstream stop compiling.
- `Synchronizer` gains an abstract `proposeOrchardToIronwoodMigration`, breaking any implementer or test fake.
- `CompactBlockUnsafe` gains a required constructor parameter.

A `### Breaking changes` (or `### Removed`/`### Changed`) section listing these would save every consumer of this rc the same rediscovery.

Separately: the privacy property of the migration is currently documented only in the rustdoc and KDoc. That one transaction reveals the account's entire Orchard balance on-chain is a *user-facing* characteristic, and wallet teams will design the confirmation UI from the changelog, not from `migration.rs`. It belongs here too.

lightwallet-client-lib/src/main/proto/compact_formats.proto
```diff
@@ -49,9 +50,12 @@ message CompactTx {
     //    valueBalanceSapling + valueBalanceOrchard + sum(vPubNew) - sum(vPubOld) - sum(tOut)
     uint32 fee = 3;
 
+    // The field numbers below must follow lightwalletd's generated bindings:
+    // https://github.com/zcash/lightwalletd/blob/master/walletrpc/compact_formats.pb.go
     repeated CompactSaplingSpend spends = 4;
     repeated CompactSaplingOutput outputs = 5;
     repeated CompactOrchardAction actions = 6;
+    repeated CompactOrchardAction ironwoodActions = 9;
```
pacu, line 82: Field number verified correct against upstream. ✅ Two notes on the surrounding bookkeeping:

1. The citation points at `lightwalletd/blob/master/walletrpc/compact_formats.pb.go`, but in lightwalletd that path is now a **symlink** to `../lightwallet-protocol/walletrpc/`. The canonical source is `zcash/lightwallet-protocol` — and pinning a tag or commit rather than `master` would keep the citation from drifting out from under the field numbers it justifies.

2. This copy claims field 9 without reserving 7 and 8, which upstream uses for `repeated CompactTxIn vin = 7` and `repeated TxOut vout = 8`. Adding `reserved 7, 8;` documents why the numbering skips and prevents a future edit from colliding with upstream.

.github/actions/setup/action.yml
```diff
@@ -21,10 +21,17 @@ runs:
...
+      # Use the channel pinned in rust-toolchain.toml (the single source of truth) rather than a
+      # hardcoded version. rustup honors the toolchain file in-repo regardless, so a hardcoded
+      # version here installs components onto a toolchain the build never runs: that is why
+      # `cargo clippy` failed with "'cargo-clippy' is not installed for the toolchain 1.92.0"
+      # while clippy had been installed for 1.84.1.
       run: |
-        rustup install 1.84.1
-        rustup default 1.84.1
-        rustup target add armv7-linux-androideabi aarch64-linux-android i686-linux-android x86_64-linux-android
+        CHANNEL="$(grep -oP 'channel\s*=\s*"\K[^"]+' rust-toolchain.toml)"
```
pacu: `grep -oP` is GNU-only; it fails on macOS/BSD grep. Every job here is `ubuntu-latest` so this works today, but a composite action under `.github/actions/` is the sort of thing that gets reused from a macOS runner later. `sed -n 's/^channel *= *"\([^"]*\)".*/\1/p'` is portable and no less readable.

.gitignore
```diff
@@ -88,6 +88,9 @@ fastlane/readme.md
 # Rust / Cargo
 fraget/
 
+# Claude Code local config and agent worktrees
+.claude/
```
pacu, line 92: This ignores the entire `.claude/` directory, which also excludes `.claude/settings.json` and any agent or skill definitions the team might want to share — the same reasoning that put `AGENTS.md` under version control.

If the intent is to ignore local state rather than shared config, `.claude/settings.local.json` plus `.claude/worktrees/` is the narrower form.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/Backend.kt
```diff
@@ -35,6 +35,14 @@ interface Backend {
         memo: ByteArray? = null
     ): ProposalUnsafe
 
+    /**
+     * Proposes migrating the account's entire Orchard balance into the Ironwood pool.
+     *
+     * @throws RuntimeException if NU6.3 is not active, if any Orchard note is not yet
+     * spendable, or as a common indicator of the operation failure
+     */
+    suspend fun proposeOrchardToIronwoodMigration(accountUuid: ByteArray): ProposalUnsafe
```
pacu, line 45: Missing `@Throws(RuntimeException::class)`, which every sibling method on this interface carries — including the ones whose KDoc makes the same `@throws RuntimeException` claim as the doc comment directly above.

Also worth noting: nothing exercises this new path anywhere. `FakeRustBackend` throws, and there's no Rust-side test for the NU6.3 gate in `propose_orchard_to_ironwood`. Since this PR adds a `cargo test --all-features` job, a test asserting the gate rejects pre-activation would be close to free and would cover the one branch most likely to regress silently.

pacu (dismissed):
I think that claude found some interesting points that are worth looking into

backend-lib/src/main/rust/lib.rs
nuttycom: Not a bug; Orchard is correct here both before and after activation. This is the *circuit* version, after NU6.3 they're identical.

nuttycom: @-

.github/actions/setup/action.yml
nuttycom: @-

lightwallet-client-lib/src/main/proto/compact_formats.proto
nuttycom: @-

.github/workflows/pull-request.yml
nuttycom: @-

CHANGELOG.md
nuttycom: @-

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/Backend.kt
nuttycom: @-

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
nuttycom: @-

.gitignore
nuttycom: @-

pacu (approved):
Ack with subsequent PR

nullcopy (approved):

Issue #2062, opened 2026-07-26
Add Rust test coverage for the NU6.3 activation gate in propose_orchard_to_ironwood


nuttycom:
## Context

`backend-lib/src/main/rust/migration.rs::propose_orchard_to_ironwood` refuses to build a proposal unless NU6.3 is active at the chain tip:

```rust
match network.activation_height(NetworkUpgrade::Nu6_3) {
    Some(h) if chain_tip >= h => (),
    _ => return Err(anyhow!("Ironwood (NU6.3) is not active yet; ...")),
}
```

That gate is load-bearing. Before activation, `input_selection.rs` picks the output pool on `ironwood_active_at`, so the identical call is still accepted by the input selector but constructs an *Orchard*-pool output — a self-send that migrates nothing and costs a fee. The gate is what turns that silent no-op into an error.

## The problem

Nothing exercises it. Raised in review on #2059:

- `FakeRustBackend` throws for `proposeOrchardToIronwoodMigration`, so no Kotlin test reaches the Rust path.
- `backend-lib` has essentially no Rust test scaffolding — `utils/exception.rs` is the only module with a `#[cfg(test)]` block, and `Cargo.toml` has no `[dev-dependencies]`.

#2059 also added a `cargo test --all-features` CI job, so a test here would actually run.

## What to do

Add a test asserting the gate rejects a pre-activation chain tip. This needs:

- `zcash_client_sqlite` with its `test-dependencies` feature under `[dev-dependencies]` in `backend-lib/Cargo.toml`.
- An in-memory `WalletDb` fixture with a chain tip below the NU6.3 activation height (or on a network where `activation_height(Nu6_3)` is `None`).
- An assertion that `propose_orchard_to_ironwood` returns `Err`, and — ideally — a companion post-activation case asserting the proposal's output lands in the Ironwood pool rather than Orchard, which is the failure mode the gate exists to prevent.

Deferred from #2059 because standing up the fixture is a meaningful addition to land on a release branch.

---

_Filed with the assistance of Claude Code._

Issue #2061, opened 2026-07-26
Stop tolerating missing Ironwood subtree roots once lightwalletd support is widespread


nuttycom:
## Context

`CompactBlockProcessor.getSubtreeRoots` fetches subtree roots for Sapling, Orchard, and Ironwood on every `start()`. A fetch failure for Sapling or Orchard is recorded into the returned `GetSubtreeRootsResult`; the Ironwood fetch's failure is **deliberately swallowed**.

This was introduced in #2059 as a review follow-up. The reason is transitional: no currently deployed lightwalletd answers `GetSubtreeRoots` for `ShieldedProtocol.ironwood` (enum value `2`, added in `zcash/lightwallet-protocol` v0.5.0), so until the server population is upgraded that request fails for every wallet on every sync start. Recording it would surface a `FailureConnection`/`OtherFailure` that the trailing `saplingSubtreeRootList.isNotEmpty()` block discards anyway.

## The problem

This tolerance is intended to be temporary, and nothing currently forces it to be revisited. Once servers do support Ironwood, a genuine Ironwood subtree-root fetch failure — a real server or network fault — will be silently ignored, and the wallet will proceed to spend-before-sync with an incomplete Ironwood commitment tree rather than reporting a failure.

There is also a per-sync latency cost while the tolerance is in place: the shared helper still burns all `GET_SUBTREE_ROOTS_RETRIES` (3) attempts with exponential backoff before giving up, roughly 1.5s added to every sync start.

## What to do

Restore Ironwood to the same failure handling as Sapling and Orchard:

```kotlin
val ironwoodSubtreeRootList =
    fetchSubtreeRoots(ironwoodStartIndex, ShieldedProtocolEnum.IRONWOOD) { failure -> result = failure }
```

and drop the explanatory comment above the three `fetchSubtreeRoots` call sites in `sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt`.

Alternatively, if the fetch should be skipped rather than tolerated before NU6.3 activation, gate it on the activation height. That needs an activation-height source plumbed into `CompactBlockProcessor` (from the Rust backend or `GetLightdInfo`), which does not exist in the Kotlin layer today — see the existing comment in `getSubtreeRoots` about NU5 and custom networks for why activation-height gating was avoided there.

## Trigger

[...]

PR #2063, opened 2026-07-26
Address review comments on #2059

nuttycom:
Addresses the inline review comments from @pacu on #2059. Branched off `release/v2.7.0-rc.1` and targeted back at it, so the release commit stays at the tip of the release branch.

Deliberately opened without a closing keyword: this is follow-up to review on another PR, and the two issues it references (#2061, #2062) are follow-ups that it does **not** close.

Resolves MOB-1543, MOB-1542

## Commits

| Commit | Comment addressed |
|---|---|
| `85589edd` | CI: emulator.wtf scaffolding + `grep -oP` portability |
| `6e55c4ea` | Proto `reserved 7, 8` + citation drift |
| `61a4ed59` | Missing `@Throws` + the `ValuePool::Orchard` question |
| `d81477fe` | Ironwood subtree-root fetch masking |
| `9d4a1b65` | Changelog breaking changes + migration privacy |

## The one real bug

Removing the emulator.wtf job left `check_emulator_wtf_secrets` defined and still referenced by the Firebase Test Lab job:

```yaml
if: needs.check_firebase_secrets.outputs.has-secrets == 'true'
    && needs.check_emulator_wtf_secrets.outputs.has-secrets == 'false'
```

On any repo or fork where `EMULATOR_WTF_API_KEY` is still configured, FTL skipped itself in deference to a job that no longer exists — instrumentation coverage gone with no failure and no warning.

The Gradle-side scaffolding (the `wtf.emulator` plugin, its maven repo, `zcash-sdk.emulator-wtf-conventions.gradle.kts`, the `gradle.properties` keys) is left in place rather than touching every module's build file on a release branch. None of it disables a job the way the workflow gate did.

## The `ValuePool::Orchard` question: not a bug

The review asked whether pinning `orchard::ValuePool::Orchard` in the PCZT proving path could prove an Ironwood bundle against the wrong circuit. It cannot. `orchard 0.15.3` documents `BundleVersion::circuit_version` as many-to-one, and the body derives only from `protocol_version`:

```rust
pub fn circuit_version(&self) -> OrchardCircuitVersion {
    match self.protocol_version {
        ProtocolVersion::InsecureV1 => OrchardCircuitVersion::InsecurePreNu6_2,
        ProtocolVersion::V2 => OrchardCircuitVersion::FixedPostNu6_2,
        ProtocolVersion::V3 => OrchardCircuitVersion::PostNu6_3,
    }
}
```

`Orchard` is also the only *total* choice: `bundle_version_for_branch(_, ValuePool::Ironwood)` is `None` for every branch before NU6.3, so deriving the pool from the bundle — the suggested fix — would have broken proving for every pre-NU6.3 Orchard PCZT. No behavioural change; the contract is now recorded in a comment so a future divergence surfaces as a documented-contract break rather than as this question again.

## Ironwood subtree roots

The shared fetch helper takes an explicit `onFailure`, so which pools record a failure is stated at the call site:

```kotlin
val saplingSubtreeRootList =
    fetchSubtreeRoots(saplingStartIndex, ShieldedProtocolEnum.SAPLING) { failure -> result = failure }
val orchardSubtreeRootList =
    fetchSubtreeRoots(orchardStartIndex, ShieldedProtocolEnum.ORCHARD) { failure -> result = failure }
val ironwoodSubtreeRootList =
    fetchSubtreeRoots(ironwoodStartIndex, ShieldedProtocolEnum.IRONWOOD) { /* tolerated; see above */ }
```

Sapling and Orchard record exactly as they did on `maint/v2.5.x` before Ironwood existed — checked against the pre-refactor code, where both per-pool loops assigned `result`. Only Ironwood is tolerated, because only Ironwood fails universally today: no deployed lightwalletd answers for `ShieldedProtocol.ironwood`.

That tolerance is transitional and tracked in #2061, referenced from the code comment. Left in past the server rollout it would ignore a genuine Ironwood fault and let spend-before-sync proceed on an incomplete commitment tree.

Activation-height gating (the reviewer's suggestion) was considered and not taken: there is no activation-height concept in the Kotlin layer, so it would mean plumbing one in from the Rust backend or `GetLightdInfo` — more than a release branch should carry. Recorded as the alternative in #2061.

## Found while in there

The subtree-root failure branch called `traceScope.end()` on *every* failed attempt, while `retryUpToAndContinue` swallows the throw and control always reaches the trailing `traceScope.end()`. Each retry closed an already-closed scope and logged `TraceScope ... ended more than once!`. Removed — routine Ironwood failures would have made it constant.

## Not done

- **`.gitignore`**: keeping `.claude/` ignored wholesale. Nothing under it is currently intended to be shared, and the narrower form invites committing local state by accident.
- **Rust test for the NU6.3 gate**: filed as #2062. `backend-lib` has no Rust test scaffolding to hang it on — `utils/exception.rs` is the only module with a `#[cfg(test)]` block and `Cargo.toml` has no `[dev-dependencies]`. A real test needs `zcash_client_sqlite`'s `test-dependencies` feature plus an in-memory `WalletDb` below the NU6.3 activation height.

## Verification

- `cargo check --manifest-path backend-lib/Cargo.toml --all-features`
- `./gradlew ktlint :sdk-lib:compileDebugKotlin :lightwallet-client-lib:compileDebugKotlin` — protoc accepts the `reserved` statement, codegen unchanged
- Every `needs:` in `pull-request.yml` resolves to a defined job
- `sed -n 's/^channel *= *"\([^"]*\)".*/\1/p' rust-toolchain.toml` → `1.92.0`, matching the `grep -oP` it replaces

---

_Opened with the assistance of Claude Code._

### DIFF
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
@@ -1345,12 +1345,24 @@ class CompactBlockProcessor internal constructor(
 
         var result: GetSubtreeRootsResult = GetSubtreeRootsResult.Linear
 
-        // Fetching the subtree roots of a pool differs only in which pool is targeted, so the
-        // three per-pool retry loops share one implementation. `result` and `traceScope` are
-        // captured so a failure reports the same way it did when each loop was written out.
+        // Fetching the subtree roots of a pool differs only in which pool is targeted and in
+        // whether a failure is recorded, so the three per-pool retry loops share one implementation.
+        //
+        // Sapling and Orchard record into [result] exactly as they did before Ironwood existed.
+        // Only the Ironwood fetch tolerates failure: no deployed lightwalletd answers for a pool
+        // that activates at NU6.3, so until the server population is upgraded this request fails
+        // for every wallet, on every sync start. Recording that would set a value the trailing
+        // `saplingSubtreeRootList.isNotEmpty()` block discards anyway; tolerating it keeps the
+        // absence of Ironwood roots from reading as a fetch failure.
+        //
+        // This tolerance is transitional and must not outlive the server rollout, or a genuine
+        // Ironwood fetch fault will be ignored and spend-before-sync will proceed on an
+        // incomplete Ironwood commitment tree. Tracked in
+        // https://github.com/zcash/zcash-android-wallet-sdk/issues/2061.
         suspend fun fetchSubtreeRoots(
             startIndex: UInt,
-            shieldedProtocol: ShieldedProtocolEnum
+            shieldedProtocol: ShieldedProtocolEnum,
+            onFailure: (GetSubtreeRootsResult) -> Unit
         ): List<SubtreeRoot> {
             var roots: List<SubtreeRoot> = emptyList()
             retryUpToAndContinue(GET_SUBTREE_ROOTS_RETRIES) {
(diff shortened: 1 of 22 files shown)


pacu (dismissed):
utACK

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -1345,12 +1345,24 @@ class CompactBlockProcessor internal constructor(
...
+        // whether a failure is recorded, so the three per-pool retry loops share one implementation.
+        //
+        // Sapling and Orchard record into [result] exactly as they did before Ironwood existed.
+        // Only the Ironwood fetch tolerates failure: no deployed lightwalletd answers for a pool
+        // that activates at NU6.3, so until the server population is upgraded this request fails
+        // for every wallet, on every sync start. Recording that would set a value the trailing
+        // `saplingSubtreeRootList.isNotEmpty()` block discards anyway; tolerating it keeps the
+        // absence of Ironwood roots from reading as a fetch failure.
+        //
+        // This tolerance is transitional and must not outlive the server rollout, or a genuine
```
nullcopy, line 1358: Is this TODO tracked somewhere?

nuttycom: See 3 lines below: https://github.com/zcash/zcash-android-wallet-sdk/issues/2061

nullcopy: :facepalm:

nullcopy (approved):
utACK a88745a412de09e5d17e2aa24959679059b2666a

PR #2066, opened 2026-07-26
fix: create and redact Ironwood proofs in the PCZT prover/signer glue

dannywillems:
## Development

Two Ironwood parity fixes in the PCZT JNI glue, each in its own commit (plus a dedicated CHANGELOG commit).

### 1. `addProofsToPczt` now creates Ironwood proofs

The prover glue only handled Orchard and Sapling, so any PCZT with Ironwood Actions (e.g. a Keystone-signed spend from the Ironwood pool) sailed through proving without an Ironwood proof and failed at extraction with `Pczt(Extraction(Ironwood(Extract(MissingProof))))`.

Mirrors the Orchard branch: derives the Ironwood circuit version from the PCZT's consensus branch id (`bundle_version_for_branch` with `ValuePool::Ironwood`) and calls `Prover::create_ironwood_proof` with a proving key built for that circuit. The stale comment claiming `ValuePool::Orchard` is correct for Ironwood bundles is updated to reflect the per-pool derivation.

### 2. `redactPcztForSignerRole` now redacts the Ironwood bundle

While auditing pool parity, the Signer-role redaction was found to clear the spend witness and the internal `zcash_client_backend:output_info` metadata for Orchard and Sapling but leave the Ironwood bundle untouched. A Keystone-signed Ironwood spend therefore shipped its Merkle witnesses and wallet output metadata to the external signer, which does not need them. Fixed by mirroring the Orchard redaction for Ironwood.

## Verification

- `cargo check` on `backend-lib` is clean; the crate is formatted with the repo toolchain.
- Confirmed against the pinned `pczt 0.8.0` that `requires_ironwood_proof`/`create_ironwood_proof`/`redact_ironwood_with` exist and share the Orchard shape, and that the extractor and Signer role authorize Ironwood identically to Orchard.

Other Orchard sites (subtree roots, balance, sync index, pool mapping, migration proposal) already handle Ironwood.

### DIFF
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -79,7 +79,10 @@ use zcash_client_sqlite::{
 use zcash_primitives::{
     block::BlockHash,
     merkle_tree::HashSer,
-    transaction::{Transaction, TxId, components::orchard::bundle_version_for_branch},
+    transaction::{
+        Transaction, TxId, builder::cached_orchard_proving_key,
+        components::orchard::bundle_version_for_branch,
+    },
 };
 use zcash_proofs::prover::LocalTxProver;
 use zcash_protocol::{
@@ -2621,7 +2624,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_addProofs
                 anyhow!("PCZT requires an Orchard proof but its consensus branch does not support Orchard")
             })?;
             prover = prover
-                .create_orchard_proof(&orchard::circuit::ProvingKey::build(circuit_version))
+                .create_orchard_proof(cached_orchard_proving_key(circuit_version))
                 .map_err(|e| anyhow!("Failed to create Orchard proof for PCZT: {:?}", e))?;
         }
         assert!(!prover.requires_orchard_proof());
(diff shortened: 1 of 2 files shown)


backend-lib/src/main/rust/lib.rs
```diff
@@ -2620,6 +2626,16 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_addProofs
         }
         assert!(!prover.requires_orchard_proof());
 
+        if prover.requires_ironwood_proof() {
+            let circuit_version = circuit_version_for(orchard::ValuePool::Ironwood).ok_or_else(|| {
+                anyhow!("PCZT requires an Ironwood proof but its consensus branch does not support Ironwood")
+            })?;
+            prover = prover
+                .create_ironwood_proof(&orchard::circuit::ProvingKey::build(circuit_version))
```
dannywillems: This could be lazily loaded, I think.

dannywillems: See [a552ac5](https://github.com/zcash/zcash-android-wallet-sdk/pull/2066/commits/a552ac50d0bedc7b0162f86ead8e051c8786911b)

nuttycom (approved):
utACK edb9f5dd8771058115d700b3b405b19768728d0c

PR #2068, opened 2026-07-26
Carry migration transfer ids as Long, not decimal strings

nuttycom:
The Swift SDK counterpart is zcash/zcash-swift-wallet-sdk#1868; this is the same defect on the JNI boundary.

## The change

The engine's transfer id is a `u32`. The boundary rendered it as a Java `String` outbound and parsed it back inbound, which buys nothing — every id is engine-issued, round-tripped verbatim by the app, and never displayed — while costing a string allocation per schedule row, a parse-failure lane on every inbound call, and a Kotlin API where an opaque-looking `String` invites callers to construct one.

Ids now cross as **`Long`**: the width this boundary already uses for every unsigned 32-bit value (heights included), with the same `isInUIntRange()` guard, so an out-of-range id fails loudly at the boundary rather than silently becoming a different transaction.

- **Outbound:** `JniTransferProposal`, `JniPreparedTransfer`, `JniMigrationTransferState`, `JniUnsignedTransferPczt`, `JniAttentionReason.InvalidTransfer` (constructor signatures `(Ljava/lang/String;…)V` → `(J…)V`), and their `MigrationSdk` counterparts.
- **Inbound:** `recordTransferResult(transferId:)`, and `storeSignedSchedulePczts`' `ids` array — now a `LongArray`, read as a primitive region rather than element-by-element.

## Which identifier this is

It names the **transfer**, not one broadcast attempt. `rebuild_expired_transfer` keeps the transfer id while producing a new transaction with a new `TxId`, so this id — not the `TxId` — is what an app correlates its own records on across an expiry rebuild. zcash/librustzcash#2767 renames the engine type `MigrationTxId` → `MigrationTransferId` for the same reason; when the pin moves to that rc, the Rust references here rename with it.

## Testing

`:sdk-lib:compileDebugKotlin` and `:sdk-lib:testDebugUnitTest` both pass, and the JNI crate cross-compiles (`cargo ndk -t arm64-v8a check`, plus the full `cargoBuildArm` that the Kotlin build drives). The JNI signature changes are compile-checked on the Kotlin side and hand-matched on the Rust side — #2017's JNI ABI check would be the mechanical guard for exactly this class of change, if it lands.

Not run: instrumented tests (no emulator in this environment).

---

This PR was prepared with Claude Code assistance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,15 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   line, adapting the backend to the send-max and builder API changes.
 - Updated the librustzcash crates to their published releases,
   `zcash_client_backend 0.24.0-rc.2` and `zcash_client_sqlite 0.22.0-rc.2`.
+- Migration transfer ids are `Long` (the engine's `u32`, widened as this JNI boundary widens every
+  unsigned 32-bit value) rather than decimal strings, across `JniTransferProposal`,
+  `JniPreparedTransfer`, `JniMigrationTransferState`, `JniUnsignedTransferPczt`,
+  `JniAttentionReason.InvalidTransfer`, their `MigrationSdk` counterparts
+  (`TransferProposal`, `MigrationTransferState`, `AttentionReason.InvalidTransfer`),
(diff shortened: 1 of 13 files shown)


Cosmos-Harry (approved):
utACK

PR #2075, opened 2026-07-27
Merge maint/v2.8.x back to `main`, restoring shielded voting on the Ironwood stack

### LINKED ISSUE #2056: backend-lib: the zcash_voting =0.11.0 pin duplicates six Zcash crates in the native library
Follow-up from the review of #2051.

Because `zcash_voting` is pinned to `=0.11.0` while everything else moved to the librustzcash 0.24 generation, the merged `backend-lib/Cargo.lock` carries **two versions of six Zcash crates**:

| crate | versions in the lock |
| --- | --- |
| `orchard` | 0.14.0, 0.15.4 |
| `zcash_primitives` | 0.28.0, 0.30.0 |
| `zcash_protocol` | 0.9.0, 0.10.1 |
| `zcash_address` | 0.12.0, 0.13.0 |
| `zcash_transparent` | 0.8.0, 0.10.0 |
| `pczt` | 0.7.0, 0.8.0 |

#2051's description mentions the first two; the other four are duplicated as well.

Mitigating: `halo2_proofs` (0.3.4) and `sapling-crypto` (0.7.0) each resolve to a single version, which caps how bad the size impact can be.

### Asks

**1. Measure the cost.** Record the stripped `.so` size per ABI before and after #2051. This is an Android library, so native size lands directly in consumer APKs.

**2. Collapse the duplication** once `zcash_voting` supports the orchard 0.15 line, and drop the `=` pin at that point.

**3. Write down the invariant that makes the split safe.** #2051's description states that "the voting JNI surface only uses `orchard::primitives::redpallas` and never passes orchard types across the `zcash_voting` boundary". The first half is not accurate — the voting code also uses `orchard::keys::Scope`, `orchard::note::{Note, ExtractedNoteCommitment}`, `orchard::tree::MerkleHashOrchard`, and `orchard::NOTE_COMMITMENT_TREE_DEPTH`.

The conclusion still holds: the `zcash_voting` boundary is byte-oriented (`fvk_bytes`, `hotkey_raw_address`, raw note data), so no orchard type crosses it. But that is the stronger and more useful statement, and it belongs in a comment next to the pin in `Cargo.toml`, because it is what keeps the two-version graph sound.

Worth calling out explicitly in that comment: witness data computed with `MerkleHashOrchard` from **orchard 0.15** is consumed by voting-circuits via **orchard 0.14**. That crossing is by bytes, so the compiler cannot catch a divergence — only the voting instrumentation tests exercise it.

nuttycom:
Merges the 2.8.x maintenance line back into `main` and brings the shielded voting surface onto the Ironwood (NU6.3) dependency stack in the process.

Closes #2056.

## Why the dependency had to move

`zcash_voting` was pinned to `=0.11.0` from crates.io, which requires the pre-Ironwood librustzcash family. Since this crate is on the Ironwood family, cargo resolved **both**, and the native library carried two copies each of eight Zcash crates: `orchard`, `pczt`, `shardtree`, `zcash_address`, `zcash_keys`, `zcash_primitives`, `zcash_protocol` and `zcash_transparent`. That is what #2056 reports.

The dependency now comes from [`zodl-inc/zcash_voting@bde9ee02`](https://github.com/zodl-inc/zcash_voting/commit/bde9ee0283ff3e23995bd6996dff3ba1f8969076) (version `2.0.0-rc.1`), whose pins match this crate's. Every Zcash crate resolves exactly once. Voting is compiled unconditionally again rather than gated behind `cfg(zcash_voting)`, and the ERROR-level deprecation on `VotingRustBackend` is gone.

**Reviewers: check the version requirement and the patched revision together.** A `[patch.crates-io]` entry only applies when it satisfies the dependency requirement. If the two drift apart, cargo silently ignores the patch, resolves the crates.io release, and rebuilds the duplicated graph — while still compiling successfully. It reports this only as a warning. `backend-lib/Cargo.toml` carries a comment to that effect.

## API changes

Upstream absorbed orchestration this SDK used to drive step by step and made the intermediate steps private, so the JNI surface goes from 60 native entry points to 55: three added, eight removed (seven public API, the eighth a test fixture).

Four vote-casting entry points collapse into a single `commitVote`. `storeVoteTxHash` merges into the conflict-checked `markVoteSubmitted`. `decomposeWeight` has no replacement. Full detail is in `CHANGELOG.md`.

### The change most likely to affect a shipped wallet

Voting hotkeys are no longer derived from the wallet seed. They are app-owned random values, so **the application must persist the hotkey's stored secret**. It cannot be re-derived: restoring a wallet from its seed phrase does not restore the ability to vote, and voting power already delegated to a hotkey whose secret was lost is forfeit. Hotkeys created by an earlier SDK version do not carry over.

### One API changed meaning without changing signature

`deriveHotkeyRawAddress(ByteArray, Int)` is byte-identical before and after, but its first parameter went from a wallet-seed-derived `hotkeySeed` to an app-owned `hotkeyStoredSecret`. **Every existing call site compiles unchanged.**

This is worse than a normal rename: BIP-39 seeds are exactly 64 bytes, which is exactly the required stored-secret length, so a wallet that keeps passing its wallet seed is *accepted at run time* and silently returns a different address — while its wallet seed becomes per-round hotkey material in possibly weaker storage. There is no compile error to catch it.

## Two pre-existing bugs the port surfaced

Neither was caused by this work; adapting to the new API exposed them, and neither could be left for a follow-up because the port does not function with them in place.

- The consensus branch id was hardcoded to `BranchId::Nu6`. Upstream now validates the supplied branch against the round's stored snapshot, and governance PCZT construction accepts only `Nu6_3`, so every real Ironwood round would have failed at run time. It is now derived from the round.
- Two PCZT readers looked in `pczt.orchard()`, which is **empty** for an NU6.3 governance PCZT — its actions are Ironwood/V3. Signature extraction and the recipient fixture read the wrong bundle.

## Test plan

Run locally on macOS with JDK 17:

- `cargo check --lib --all-targets`, `cargo clippy --all-targets`, `cargo fmt --check` — clean, no warnings
- `cargo test --lib` — 14 passed, 0 failed
- `./scripts/ci-local.sh quick` and `fast` — both pass (detekt, ktlint, unit tests)
- `:backend-lib:compileDebugAndroidTestKotlin` and `:sdk-lib:compileDebugAndroidTestKotlin` — both compile

The JNI boundary was cross-checked mechanically rather than by inspection, because a Kotlin `external fun` whose native counterpart changed compiles cleanly and fails only when invoked:

- 55 Kotlin `external fun` declarations against 55 Rust exports — set-identical, with zero arity, positional-type or return-type mismatches
- `javap -p -s` on the compiled `VotingRustBackend` — all 55 JVM descriptors match the Rust parameter lists
- `nm -gU` on the built `libzcashwalletsdk.so` — 47 exported voting symbols; the 8 absent are exactly the `android-test-fixtures`-gated ones
- 18/18 JNI model constructor descriptors verified against the compiled classes, which rules out `NoSuchMethodError`

**What the static checks did not cover, and what CI has now confirmed:** none of the above executes a JNI call. `VotingRustBackendTest` (47 tests) was `@Ignore`d on the maintenance line while voting was gated off, with the stated condition "re-enable with the voting module". This branch re-enables the module and deletes the `@Ignore`, so `test_android_modules_emulator` has exercised that suite against the ported native library — the first time it has run since voting was switched off.

That run found nine real failures, in three groups, each fixed in a commit of its own:

- Six from the round snapshot height. The port derives the consensus branch from the round rather than a hardcoded branch id, and governance PCZTs are Ironwood-only, so the old fixture height of `123_456` resolved to a pre-NU6.3 branch and was rejected. The fixture now sits at testnet's NU6.3 activation height, which is above mainnet's and so covers both networks the suite drives.
- Two from `share::record`, which now derives the share nullifier from the vote's own recovery bundle rather than taking one from the caller. A vote row alone is no longer enough, so the `android-test-fixtures`-gated `storeVoteFixture` helper now also writes the bundle a real `vote::commit` would have.
- One stale mock in the typesafe suite. `commitVote` checks that the commitment it receives names the bundle that was requested; the canned result named a different one. The check is correct production behaviour and is covered separately, so only the fixture moved.

A tenth surfaced on the following run, inverted: a test asserting that `extractSpendAuthSig` throws on an unsigned governance PCZT, which now succeeds. That is upstream behaviour — `extract_spend_auth_sig` falls back to scanning every action for a signature, and the Ironwood builder's IO finalizer signs the zero-value padding action, so the scan finds that one. The crate's comment on the fallback ("only the real governance spend gets signed") predates the finalizer doing so. The signature returned is real but wrong, and is caught a step later when verified against `rk` and the sighash — a containment that has its own passing test. The assertion now pins the shape of what extraction returns rather than being deleted, so an upstream change to the fallback surfaces rather than passing silently.

**All 16 checks now pass, including `test_android_modules_emulator`.** Notably, no `UnsatisfiedLinkError` or `NoSuchMethodError` appeared in any run: every failure was semantic, reached through a working JNI boundary, which is consistent with the static cross-check above.

## Follow-ups, not blocking

- `BundleLayout.dropped_count` is discarded, so notes in dust bundles below `BALLOT_DIVISOR` silently reduce a user's voting weight with no count surfaced to the caller.
- `round_consensus_branch_id` derives the branch from the caller-supplied network rather than the round's stored network. It fails closed today — upstream validates the network and then the branch id — but the stored value is available in the same query.
- `JniVotingHotkey`'s constructor and `copy()` are public where they were `internal`. Restricting them would break the only test of the length validation the restriction exists to back up, and the bypass is nominal since every native entry point takes the raw secret bytes. Documented in the type's KDoc and the changelog.
- The semgrep rule `kotlin-typesafe-returns-jni-model` has no definition or scan step anywhere in this repository, so the placement of its suppressions cannot be validated locally. The convention predates this branch; it needs whoever owns the rule.

---
Prepared with Claude Code assistance.


Cosmos-Harry (approved):
utACK [097a7d9](https://github.com/zcash/zcash-android-wallet-sdk/pull/2075/commits/097a7d919e3192fbda0501a9fb7181c6b93639c7)

PR #2081, opened 2026-07-27
Merge maint/v2.8.x into feature/ironwood-slipstream

nuttycom:
Merges `maint/v2.8.x` (`d1a42c8f`) into `feature/ironwood-slipstream`. Eight files conflicted; this describes how each was resolved and what the merge forced downstream.

## CI

**`.github/actions/setup/action.yml`** — took maint's `sed`-based `rust-toolchain.toml` parse (`grep -oP` is a GNU extension, and this is a composite action that could be reused from a macOS/BSD runner) and its `rustfmt` component. Kept this branch's slipstream token-minting and cargo-authorization steps.

**`.github/workflows/pull-request.yml`** — the two branches disabled the emulator.wtf/FTL jobs differently. Adopted maint's decision (2026-07-25) over this branch's (2026-07-24): `check_emulator_wtf_secrets` is gone and `test_android_modules_ftl` is commented out with maint's rationale. The commented body was rewritten to carry this branch's pinned checkout, `persist-credentials: false`, and slipstream setup inputs, so whoever restores that job does not silently undo the zizmor hardening.

`test_rust_unit`, `static_analysis_clippy` and `static_analysis_rustfmt` existed on both sides. Kept this branch's versions — its rustfmt job installs a minimal toolchain directly instead of running the whole setup action — and took maint's rationale comments onto them.

**`AGENTS.md`, `gradle.properties`** — both sides' additions are independent; both kept.

## Voting stays off

maint's shape wins in full. This branch had ported the voting sources to `zcash_voting` 1.0 and gated them behind a `chp-voting` cargo feature that `build.gradle.kts` switched on for instrumentation builds. That is not the intended state here — voting is re-enabled separately, against `main`, by #2075 — so:

- `mod voting` is gated behind `cfg(zcash_voting)` again; the `chp-voting` feature is gone, and maint's `[lints.rust] unexpected_cfgs` declaration for that cfg comes back with it.
- `zcash_voting`, `pasta_curves` and `ff` — the only dependencies the voting sources alone need — are commented out, as is the `zcash_voting` entry in `[patch.crates-io]` (cargo warns about a patch for a crate absent from the graph).
- `build.gradle.kts` no longer adds `chp-voting` to instrumentation builds; it still adds `android-test-fixtures`.
- `VotingRustBackendTest` is `@Ignore`d again, but keeps this branch's fixture updates for the zcash_voting 1.0 / NU6.3 API, so #2075 should not have to touch it.
- The `orchard` dependency stops asking for `unstable-voting-circuits`. Only the voting sources needed it; Ironwood pool support does not. Because cargo features are additive, dropping it here was only half the job — `slipstream-core` enabled it too — so its pin moves from `35faa0c2` to `dd11d093`, which drops it upstream. `cargo tree -e features --features slipstream | grep -c unstable-voting-circuits` is now 0, i.e. the circuits are genuinely out of the shipped library rather than merely undeclared here. The rev bump is a one-line lockfile change; `slipstream-core` stays at 0.6.4.

Building with `--cfg zcash_voting` now fails on exactly those three imports and nothing else, so re-enabling is precisely an uncomment.

One deliberate departure from maint: **`incrementalmerkletree` stays uncommented.** maint's comment says it serves voting "and nothing else", which is stale here — `migration_engine.rs` and `migration_keystone.rs` both use it.

## backend-lib, otherwise

**`addProofsToPczt`.** Both branches independently added branch-derived circuit versions and a cached proving key. Took maint's: it derives the circuit version per value pool from the PCZT's consensus branch, where this branch hardcoded `PostNu6_3` for the Ironwood proof and built that key uncached.

**Dependency work the merge forced.** `zcash_client_sqlite 0.22.0-rc.4` requires `zcash_pool_migration ^0.1.0-rc.3`, so the rc.1 pin here could not be kept. rc.3 renamed `MigrationTxId` to `MigrationTransferId` and the note-split plan accessors to `denominations()`. `migration.rs` and `migration_engine.rs` are adapted; the JNI method names (`prepareNoteSplitNative` and friends) and the Kotlin-facing surface are unchanged.

`pczt` moves to 0.9.0, keeping the `signer` feature this branch's Keystone batch signing needs. With voting gated off, `zcash_voting` no longer drags pczt 0.8 into the graph, so a single pczt resolves in every configuration.

## Verification

- `cargo check` — default, `slipstream`, and `--all-features --tests`
- `cargo clippy --tests --all-features -- -W clippy::all -D warnings`
- `cargo fmt --all --check`
- `cargo test --all-features` — 20 passed (the voting tests are no longer compiled)
- `./gradlew ktlint` and the Android native + Kotlin build

detekt reports findings only from unrelated local worktrees, none in this tree.

---

Prepared with Claude Code assistance.


Cosmos-Harry (approved):
utACK [ba9f2c9](https://github.com/zcash/zcash-android-wallet-sdk/pull/2081/commits/ba9f2c92e7dc3b2248c4cd6cd175347d992398c6)

Will have to test Keystone with this though. just to be safe.

PR #2092, opened 2026-07-30
Sync the vendored proposal.proto with zcash_client_backend 0.24.0-rc.6

dannywillems:
The Kotlin copy of `proposal.proto` had drifted from the definition the pinned
`zcash_client_backend` encodes against. Rust builds the proposal with the
upstream schema, so every field the vendored copy lacked was unreadable on the
Kotlin side.

## What was missing

- `ValuePool.Ironwood = 4`. A proposal carrying an Ironwood output or change
  value decoded that pool as `UNRECOGNIZED`, which is exactly the
  misattribution the upstream comment on the variant warns about.
- `Proposal.confirmationsPolicy = 5`, plus the `ConfirmationsPolicy` message.
- `Proposal.proposedVersion = 6`.
- `TransactionBalance.dummyOutputs = 3`, plus the `DummyOutputs` message. This
  one is Ironwood-relevant: it carries the per-pool dummy-output counts that
  determine Ironwood bundle padding.
- `ChangeValue.isEphemeral = 4`.
- Two comment typos and the change/ephemeral wording upstream has since fixed.

All are additive, so previously serialized proposals keep decoding unchanged.
Round-tripping a proposal through Kotlin already preserved these values as
unknown fields, so no proposal was ever mis-built; they were simply unreadable.

## Verification

- `protoc --descriptor_set_out` compiles the file (protoc 35.1 locally;
  `PROTOC_VERSION=4.34.0` in `gradle.properties`, both well above the 3.15
  needed for proto3 `optional`).
- Decoding the descriptor set confirms `name: "Ironwood" number: 4`, matching
  the upstream discriminant.
- `diff` against `zcash_client_backend/proto/proposal.proto` at librustzcash
  `6c07e5f329` (the `zcash_client_sqlite-0.22.0-rc.6` release commit this
  branch pins) is empty apart from the `option java_package` line. That is the
  same single local deviation `scripts/update-lightwallet-protocol.sh:11`
  documents for the other vendored protos.
- No Kotlin or Rust source references `ValuePool`, and the generated Java and
  Kotlin is not committed, so the next build picks this up with no other edits.

## Follow-up worth doing separately

Unlike `compact_formats.proto` and `service.proto`, this file has no provenance
header and no sync tooling. Nothing in `scripts/`, `.github/`, or the Gradle
build references `proposal.proto` at all, so there is nothing to tell a reader
where it comes from or to catch it drifting again. Its upstream is a different
repository from the lightwallet protos, so it needs its own equivalent of
`scripts/update-lightwallet-protocol.sh` plus a CI check that diffs it against
the pinned `zcash_client_backend`. This PR only repairs the current drift.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,15 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   note exists.
 
 ### Fixed
+- The vendored `proposal.proto` is back in sync with the definition in
+  `zcash_client_backend 0.24.0-rc.6`. It was missing the `Ironwood` value pool,
+  so a proposal with an Ironwood output or change value decoded that pool as
+  `UNRECOGNIZED`, and it was missing the `Proposal.confirmationsPolicy`,
+  `Proposal.proposedVersion`, `TransactionBalance.dummyOutputs`, and
+  `ChangeValue.isEphemeral` fields, which decoded into unknown fields and so
+  could not be read. Round-tripping a proposal through Kotlin preserved all of
+  these, so no proposal was mis-built; they were simply unreadable.
+
 All of the following were picked up from the librustzcash update:
 
 - A wallet whose database was upgraded by a build using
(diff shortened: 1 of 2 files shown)


dannywillems:
Closing in favor of https://github.com/zcash/zcash-android-wallet-sdk/pull/2093

PR #2093, opened 2026-07-30
Vendor proposal.proto from librustzcash by script, and check it in CI

dannywillems:
Resolves MOB-1612

Follow-up to #2092, which repaired the drift in `proposal.proto` by hand. This
adds the mechanism that should have prevented it.

`backend-lib/src/main/proto/proposal.proto` is a hand-copied snapshot of
`zcash_client_backend/proto/proposal.proto`. Nothing recorded where it came
from and nothing checked it, so it drifted across four upstream changes without
anyone noticing. Rust keeps encoding proposals with the pinned crate schema, so
every field the copy lacks is one Kotlin cannot read.

## What this adds

- **`scripts/update-proposal-proto.sh`** re-derives the file. It resolves the
  upstream ref from `backend-lib/Cargo.lock` rather than `Cargo.toml`, so it
  follows the *resolved* dependency: a crates.io release maps to the
  `<crate>-<version>` tag, and a git dependency maps to the exact revision
  Cargo recorded. That keeps it correct while the librustzcash git overrides at
  the bottom of `backend-lib/Cargo.toml` are in use.
- **`--check`** re-derives to a temp file and diffs, printing the offending
  hunks and failing on drift. A new `check_proposal_proto` job runs it.
- **A `DERIVED FILE` banner** on the proto naming the upstream path, the ref,
  and the script, matching what `update-lightwallet-protocol.sh` writes for the
  lightwallet protos.

As with those protos, the sole deviation from upstream is the injected
`option java_package`; without it protoc would generate into
`cash.z.wallet.sdk.ffi` rather than the `cash.z.wallet.sdk.internal.ffi` this
SDK compiles against. The check needs only curl, awk and git, so the job skips
the Gradle and Rust setup action.

## Verification

Run locally against this branch and the companion branch:

- `shellcheck` clean.
- `--check` on this branch **fails**, exit 1, printing exactly the four missing
  fields, and resolving the ref to `zcash_client_backend-0.24.0-rc.6` from
  `Cargo.lock`. That is the check catching the real drift.
- Re-deriving and stripping the banner produces a file **byte-identical** to
  the proto in #2092, which independently confirms both that hand-sync and this
  script.
- Simulating the post-merge tree (this banner plus #2092 body) makes `--check`
  **pass**.
- The workflow YAML parses and the new job registers with both steps.

## Merge order

The `check_proposal_proto` job is red on this branch until #2092 merges, since
this branch is based on `maint/v2.7.x` where the proto is still stale. That red
is the check working, not a defect. **Merge #2092 first**, then this. I kept
the two disjoint on purpose: the proto diff here is the banner only, zero
deleted lines, so there is no overlap with #2092.


scripts/update-proposal-proto.sh
```diff
@@ -0,0 +1,190 @@
...
+readonly CRATE="zcash_client_backend"
+readonly LOCK="backend-lib/Cargo.lock"
+readonly DEST="backend-lib/src/main/proto/proposal.proto"
+readonly JAVA_PACKAGE="cash.z.wallet.sdk.internal.ffi"
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+readonly REPO_ROOT
+cd "${REPO_ROOT}"
+
+usage() {
```
dannywillems, line 42: I like this way of defining `usage`. Fragile from the top level doc, but ok IMO for something easy to parse as a human.

scripts/update-proposal-proto.sh
```diff
@@ -0,0 +1,190 @@
...
+    case "${source}" in
+        registry+*)
+            echo "${CRATE}-${version}"
+            ;;
+        git+*)
+            if [ "${source}" = "${source#*#}" ]; then
+                echo "error: git source for ${CRATE} pins no revision: ${source}" >&2
+                exit 1
+            fi
+            echo "${source##*#}"
```
nesence-m, line 87: The `git+` arm takes the rev from the URL fragment but ignores the repo in the URL itself — the fetch always goes to the hardcoded `UPSTREAM_REPO="zcash/librustzcash"`. If a patch ever pins a fork (fork revs have entered this dependency graph before, e.g. the zcash_voting fork pin), the raw fetch targets `zcash/librustzcash` at a sha that may only exist in the fork: usually a 404 whose hint (`Is <sha> pushed to zcash/librustzcash?`) points at the wrong repo, and when it *does* succeed it is silently relying on GitHub serving fork-network commits through the parent path. Content can never be wrong (a sha is content-addressed), so this fails confusingly rather than dangerously — but suggest parsing `owner/repo` out of the `git+https://github.com/...` source and using that in the raw URL, or at least asserting the source URL matches `UPSTREAM_REPO` and printing the full source in the error.

scripts/update-proposal-proto.sh
```diff
@@ -0,0 +1,190 @@
...
+# Writes the derived proto for the given upstream ref to stdout.
+derive() {
+    local ref="$1"
+    local url="https://raw.githubusercontent.com/${UPSTREAM_REPO}/${ref}/${UPSTREAM_PATH}"
+    local upstream
+    upstream="$(mktemp)"
+    # shellcheck disable=SC2064 # expand the path now, while it is still in scope
+    trap "rm -f '${upstream}'" RETURN
+
+    if ! curl -fsSL "${url}" -o "${upstream}"; then
```
nesence-m, line 105: Nit: the `exit 1` paths inside `derive` bypass the `RETURN` trap (`RETURN` traps don't run on `exit`; only `EXIT` traps do), so on a curl failure the `${upstream}` temp file leaks — and in `update` mode there is no `EXIT` trap at all. Note that adding an `EXIT` trap here would clobber the one `main` sets for `${derived}` in check mode, so the cleaner fix is to `return 1` from `derive` (the `RETURN` trap then fires, and `set -e` still aborts the script). Harmless on ephemeral CI runners; just tidiness for local runs.

.github/workflows/pull-request.yml
```diff
@@ -243,6 +243,24 @@ jobs:
...
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout
+        timeout-minutes: 1
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10
+      - name: Check proposal.proto matches the pinned zcash_client_backend
+        timeout-minutes: 5
+        run: |
+          ./scripts/update-proposal-proto.sh --check
```
nesence-m, line 262: This check trusts the committed `backend-lib/Cargo.lock`, but nothing in CI holds that lock to the manifest: `test_rust_unit`, clippy, and fmt all run cargo without `--locked`, so cargo silently re-resolves in the runner when the two disagree. Failure scenario: a pin-bump PR edits the `[patch.crates-io]` rev in `Cargo.toml` but forgets to regenerate the committed lock — the Rust jobs build the *new* rev and pass, this job vendors from the *old* lock and passes, and the vendored proto is out of sync with what actually links, which is precisely the drift this job exists to prevent. Suggest adding `--locked` to the cargo invocations in the Rust jobs (that keeps this job free of the Rust toolchain, as intended) so lock staleness fails loudly. Pre-existing gap, but this check's correctness now depends on it.

nesence-m (commented):
Solid mechanism, and the core design choice is right: resolving the upstream ref from `backend-lib/Cargo.lock` rather than `Cargo.toml` tracks the *resolved* dependency, and both states work — the current released state (registry entry maps to the `zcash_client_backend-0.24.0-rc.6` tag, which is exactly what the banner records today) and a future `[patch.crates-io]` state (the lock records the resolved sha in the URL fragment, so even a `branch =` patch vendors reproducibly). I re-derived the file independently (fetched `zcash_client_backend/proto/proposal.proto` at the `zcash_client_backend-0.24.0-rc.6` tag, injected `option java_package` after the `package` line) and it is byte-identical to the vendored copy here minus the banner, matching the green `check_proposal_proto` run. The script is `set -euo pipefail` and I found nothing bash-3.2/BSD-incompatible for local macOS runs (`$'\t'`, the `RETURN` trap, and the awk/sed usages are all fine there).

One item that doesn't anchor to a diff line: **the PR description no longer matches the pushed head.** The "Merge order" section says the proto diff here is "the banner only, zero deleted lines", that `check_proposal_proto` is red until #2092 merges, and to merge #2092 first. In fact commit 841d10ee ("Run ./scripts/update-proposal-proto.sh") regenerated the whole file (53 insertions, 9 deletions), the check is green on this branch, and #2092 is closed unmerged — this PR now carries the full schema sync itself. That end state is fine (arguably better), but please refresh the description so the merge record doesn't assert the opposite of what actually lands.

Line comments below cover two robustness gaps (fork pins, a stale-lockfile blind spot) and a temp-file nit.

scripts/update-proposal-proto.sh
dannywillems: >hardcoded UPSTREAM_REPO="zcash/librustzcash"

I think it is fine.

nesence-m (approved):
utACK

PR #2094, opened 2026-07-30
Add a Makefile and route every CI job through it

dannywillems:
Each CI job repeated its own gradle or cargo invocation, so reproducing a
failure locally meant reading the workflow and retyping the command, and the
two drifted. This adds a Makefile that owns every command the project needs,
and has `pull-request.yml` call the targets.

Independent of, and not stacked on, the `proposal.proto` PRs (#2092, #2093).

## Eleven jobs now run a target

`check-properties`, `detekt`, `ktlint`, `lint-android`, `test-unit`,
`build-release`, `test-rust`, `lint-rust`, `check-format-rust`,
`test-instrumented`, `test-robo`. No raw `./gradlew` or `cargo` invocation is
left in the workflow.

I diffed each target expansion against the command the job ran before: **all
eleven are byte-identical**, module order and gradle `-P` flags included. The
three cargo jobs drop their `working-directory: backend-lib`, since make runs
from the root and changes directory itself.

## A real divergence this surfaced

CI runs the instrumentation tests on the **`pixel2Min`** managed device, while
`scripts/ci-local.sh` uses **`pixel2Target`**. Both are defined in
`zcash-sdk.android-conventions.gradle.kts`, so "running CI locally" was
quietly exercising a different API level. The Makefile defaults to CI's
`pixel2Min`; override with `make test-instrumented MANAGED_DEVICE=pixel2Target`.

I left `scripts/ci-local.sh` alone. Unifying it is a follow-up, and worth
doing.

## What is deliberately not converted

- `validate_gradle_wrapper` is a GitHub Action (checksum validation of the
  wrapper jar), not a command.
- `check_firebase_secrets` is CI logic testing whether secrets are defined.
- Artifact `mkdir`/`zip`, the debug `keytool` keystore, the `sdkmanager`
  emulator install and the KVM udev rules are runner plumbing, not project
  commands.
- `test-robo` has a target for symmetry, but it hits Firebase Test Lab, so it
  needs credentials and cannot run offline.

## Portability and ergonomics

Targets GNU Make **3.81**, the version macOS ships, so none of the 4.x-only
features are used. `JAVA_HOME` and `ANDROID_HOME` are probed across both macOS
and Linux layouts (Homebrew both prefixes, Android Studio JBR, `/usr/lib/jvm`,
sdkman); an explicit value in the environment always wins, which is what makes
it correct on a runner as well as a laptop.

detekt scans the whole tree, so a git worktree left under `.claude/worktrees/`
makes it report other branches' code as your failures. The scanning targets
warn when that directory is non-empty.

## Verification

- 0 lines over 80 chars; parses and runs under GNU Make 3.81; all 40 targets
  register; `make help` and `make info` work.
- `make info` correctly resolved JDK 21 and the Android SDK on macOS.
- Dry-ran every converted target and compared against the original commands.
- Ran `make check-format-rust` for real: passes.
- Workflow YAML still parses with all 13 jobs intact.

I did not run the full Gradle builds, only verified the commands they issue.

No CHANGELOG entry: this is developer tooling, with no public API change.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -62,7 +62,7 @@ jobs:
       - name: Check properties
         timeout-minutes: 4
         run: |
-          ./gradlew checkProperties
+          make check-properties
 
   static_analysis_detekt:
     needs: validate_gradle_wrapper
@@ -80,7 +80,7 @@ jobs:
       - name: Detekt
         timeout-minutes: 4
         run: |
-          ./gradlew detektAll
+          make detekt
       - name: Collect Artifacts
         timeout-minutes: 1
         if: ${{ always() }}
@@ -115,7 +115,7 @@ jobs:
       - name: Ktlint
         timeout-minutes: 4
         run: |
-          ./gradlew ktlint
+          make ktlint
       - name: Collect Artifacts
         timeout-minutes: 1
         if: ${{ always() }}
@@ -153,7 +153,7 @@ jobs:
           # Disable minify, since it makes lint run faster
           ORG_GRADLE_PROJECT_IS_MINIFY_APP_ENABLED: false
         run: |
-          ./gradlew :sdk-lib:lintRelease :demo-app:lintZcashmainnetRelease
+          make lint-android
       - name: Collect Artifacts
         if: ${{ always() }}
         timeout-minutes: 1
@@ -190,9 +190,8 @@ jobs:
           rust-cache: true
       - name: Test
         timeout-minutes: 15
-        working-directory: backend-lib
         run: |
-          cargo test --all-features
+          make test-rust
 
   # --all-features matters: code behind a cargo feature (for example the
   # android-test-fixtures gate the Android build enables) is invisible to a
@@ -215,9 +214,8 @@ jobs:
           rust-cache: true
       - name: Clippy
         timeout-minutes: 15
-        working-directory: backend-lib
         run: |
-          cargo clippy --tests --all-features -- -W clippy::all -D warnings
+          make lint-rust
 
   # Checks that the Rust sources are rustfmt-clean. `--check` only reports; it never
   # rewrites the tree, so this job cannot mask an unformatted commit by fixing it.
@@ -239,9 +237,8 @@ jobs:
         uses: ./.github/actions/setup
       - name: Rustfmt
         timeout-minutes: 5
-        working-directory: backend-lib
         run: |
-          cargo fmt --all --check
+          make check-format-rust
 
   test_android_modules_unit:
     needs: [ validate_gradle_wrapper ]
@@ -259,7 +256,7 @@ jobs:
       - name: Build and test
         timeout-minutes: 30
         run: |
-          ./gradlew test
+          make test-unit
       - name: Collect Artifacts
         if: ${{ always() }}
         timeout-minutes: 1
@@ -405,13 +402,7 @@ jobs:
       - name: Build and test
         timeout-minutes: 45
         run: |
-          ./gradlew \
-            :sdk-lib:pixel2MinDebugAndroidTest \
-            :lightwallet-client-lib:pixel2MinDebugAndroidTest \
-            :sdk-incubator-lib:pixel2MinDebugAndroidTest \
-            :backend-lib:pixel2MinDebugAndroidTest \
-            -Pandroid.experimental.testOptions.managedDevices.maxConcurrentDevices=1 \
-            -Pandroid.testoptions.manageddevices.emulator.gpu=swiftshader_indirect
+          make test-instrumented
       - name: Collect Artifacts
         if: ${{ always() }}
         timeout-minutes: 1
@@ -464,7 +455,7 @@ jobs:
           ORG_GRADLE_PROJECT_ZCASH_RELEASE_KEY_ALIAS: androiddebugkey
           ORG_GRADLE_PROJECT_ZCASH_RELEASE_KEY_ALIAS_PASSWORD: android
         run: |
-           ./gradlew assembleRelease
+           make build-release
       - name: Collect Artifacts
         timeout-minutes: 1
         env:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -0,0 +1,350 @@
...
+# Build variant
+# -------------
+# The demo app has a network dimension (Zcashmainnet, Zcashtestnet). The
+# library modules do not. Override per invocation:
+#   make lint-android NETWORK=Zcashtestnet
+#
+# Detekt and git worktrees
+# ------------------------
+# detekt scans the whole tree, including any git worktrees left inside it
+# (for example under .claude/worktrees/). Those hold other branches' sources,
```
dannywillems: We want only the one of the current branch.

Makefile
```diff
@@ -0,0 +1,350 @@
...
+	-Pandroid.testoptions.manageddevices.emulator.gpu=swiftshader_indirect
+
+# Android targets the Rust JNI backend is cross-compiled for.
+RUST_ANDROID_TARGETS := \
+	armv7-linux-androideabi \
+	aarch64-linux-android \
+	i686-linux-android \
+	x86_64-linux-android
+
+JAVA_HOME ?= $(shell \
```
dannywillems, line 70: Trying to be complete...

Makefile
```diff
@@ -0,0 +1,350 @@
...
+		/opt/android-studio/jbr "$$HOME/android-studio/jbr" \
+		"/Applications/Android Studio.app/Contents/jbr/Contents/Home"; \
+	do \
+		[ -n "$$d" ] || continue; \
+		[ -x "$$d/bin/javac" ] || continue; \
+		echo "$$d"; break; \
+	done)
+export JAVA_HOME
+
+ANDROID_HOME ?= $(shell \
```
dannywillems, line 90: Trying to be complete here too...

Makefile
```diff
@@ -0,0 +1,350 @@
...
+	do \
+		[ -n "$$d" ] || continue; \
+		[ -d "$$d/platform-tools" ] || continue; \
+		echo "$$d"; break; \
+	done)
+export ANDROID_HOME
+
+# Gradle and Cargo parallelize internally; running two of them at once only
+# contends on their own locks, so never run these targets concurrently.
+.NOTPARALLEL:
```
dannywillems, line 102: LGTM.

Makefile
```diff
@@ -0,0 +1,350 @@
...
+
+.DEFAULT_GOAL := help
```
dannywillems, line 104: Good practice to avoid relying on the first target.

nesence-m (approved):
lgtm

Makefile
```diff
@@ -0,0 +1,328 @@
...
+	-Pandroid.testoptions.manageddevices.emulator.gpu=swiftshader_indirect
+
+# Android targets the Rust JNI backend is cross-compiled for.
+RUST_ANDROID_TARGETS := \
+	armv7-linux-androideabi \
+	aarch64-linux-android \
+	i686-linux-android \
+	x86_64-linux-android
+
+JAVA_HOME ?= $(shell \
```
nesence-m, line 70: `?=` creates a recursively expanded variable, so this whole probe — including the two `/usr/libexec/java_home` invocations — re-runs every time `$(JAVA_HOME)` is expanded, and because of the `export JAVA_HOME` below, that is at least once per recipe line make executes (plus each `$(JAVA_HOME)` reference in `info`). On a machine without `JAVA_HOME` set, an aggregate like `make check` pays the probe over and over, and a probe whose answer can change mid-run (the sdkman `current` symlink, for instance) could hand different recipes different JDKs. A 3.81-safe fix that keeps the environment-wins behavior is to freeze the value right after the conditional:

```make
JAVA_HOME ?= $(shell ...)
JAVA_HOME := $(JAVA_HOME)
```

The `ANDROID_HOME` probe below has the same shape. CI is unaffected either way, since the runners export both variables.

.github/workflows/pull-request.yml
```diff
@@ -153,7 +153,7 @@ jobs:
           # Disable minify, since it makes lint run faster
           ORG_GRADLE_PROJECT_IS_MINIFY_APP_ENABLED: false
         run: |
-          ./gradlew :sdk-lib:lintRelease :demo-app:lintZcashmainnetRelease
+          make lint-android
```
nesence-m, line 156: `ORG_GRADLE_PROJECT_IS_MINIFY_APP_ENABLED: false` stays workflow-side, so `make lint-android` on a developer machine lints with minify enabled (`gradle.properties` defaults it to `true`) — a different and slower configuration than CI, which is exactly the local/CI drift this PR set out to remove. It isn't a secret, so consider folding it into the target instead:

```make
lint-android:
	$(GRADLE) -PIS_MINIFY_APP_ENABLED=false :sdk-lib:lintRelease :demo-app:lint$(NETWORK)Release
```

and dropping the env line here, so the target reproduces the CI job byte for byte.

nesence-m (commented):
Post-merge review, for the record. The conversion is faithful: I diffed each of the eleven rewritten jobs against the old run steps, and the commands, module lists and `-P` flags are identical, with the dropped `working-directory: backend-lib` on the three cargo jobs correctly replaced by `cd backend-lib` in the recipes. `.PHONY` covers every target (including `build`/`clean`, which collide with real directory names), the `MANAGED_DEVICE=pixel2Target` override works despite the `:=` module list because command-line variables are bound before the makefile is parsed, and I checked ktlint 1.8.0's glob handling (`expandDoubleStarPatterns`): `!**/.claude/**` also matches a top-level `.claude`, so the agent-directory exclusion covers the motivating worktree case. CI on the head commit was green.

Two small items as line comments. One note that doesn't anchor to a diff line: a few jobs still shape the Gradle build via step-level `env` (the `build-release` signing properties, the robo FTL credentials), so those make targets only reproduce CI when the caller exports the same variables — worth a sentence in the Makefile header stating which targets need environment to match CI.

PR #2100, opened 2026-07-30
Carry the Ironwood tree state through checkpoints

dannywillems:
Found by an audit of where Ironwood should appear but does not. Independent of
the other open PRs; based directly on `maint/v2.7.x`.

## The bug

The lightwalletd `TreeState` message has **`string ironwoodTree = 7`**
(`service.proto:193`). The SDK never read it:

- `Checkpoint.kt` held only `saplingTree` and `orchardTree`
- `TreeState.fromParts` and `TreeStateUnsafe.fromParts` took only those two, so
  the protobuf builder never called `setIronwoodTree`
- `CheckpointExt.kt` parsed only those two

This feeds wallet initialisation (`SdkSynchronizer.kt:902`,
`DerivedDataDb.kt:157`), so a wallet seeded from a checkpoint got **no Ironwood
commitment tree state**.

**Not hypothetical:** testnet checkpoints already ship the field and it was
being read off disk and discarded. Counted in-tree: 7 of 394 on `maint/v2.7.x`,
16 of 394 on `maint/v2.8.x`. The Swift SDK already handles it
(`Checkpoint.swift:34,70,103`), so Android was simply behind.

## Optional, deliberately

`orchardTree` is *required* post-NU5, and the obvious symmetry would be to
require `ironwoodTree` post-NU6.3. **That would break every mainnet wallet.** I
counted: `0` mainnet checkpoints carry an `ironwoodTree` on `maint/v2.7.x`,
`maint/v2.8.x`, or `feature/ironwood-slipstream` (newest checkpoint 3427310).
Requiring it would reject all 807 shipped mainnet checkpoints.

So it is optional, matching Swift's `decodeIfPresent`. Tightening to a
requirement is a follow-up once the mainnet checkpoints are regenerated with
the field, and at that point the `orchardTree` block is the template.

## Verification

- `:sdk-lib:compileDebugKotlin` and `:sdk-lib:compileDebugAndroidTestKotlin`
  both compile.
- `ktlint` and `detektAll` pass (the three signatures now hit the
  6-parameter `LongParameterList` threshold and carry a scoped `@Suppress`,
  matching how the codebase handles it elsewhere).
- `CheckpointFixture.toJson` writes the field when present, so the
  parse-and-serialise round trip cannot silently drop it again.

### DIFF
--- a/lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/model/TreeStateUnsafe.kt
+++ b/lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/model/TreeStateUnsafe.kt
@@ -8,12 +8,14 @@ class TreeStateUnsafe(
     companion object {
         fun new(treeState: TreeState): TreeStateUnsafe = TreeStateUnsafe(treeState.toByteArray())
 
+        @Suppress("LongParameterList")
         fun fromParts(
             height: Long,
             hash: String,
             time: Int,
             saplingTree: String,
-            orchardTree: String
+            orchardTree: String,
+            ironwoodTree: String?
         ): TreeStateUnsafe {
             val treeState =
                 TreeState
@@ -23,6 +25,7 @@ class TreeStateUnsafe(
                     .setTime(time)
                     .setSaplingTree(saplingTree)
                     .setOrchardTree(orchardTree)
+                    .apply { ironwoodTree?.let { setIronwoodTree(it) } }
                     .build()
             return new(treeState)
         }
--- a/sdk-lib/src/androidTest/java/cash/z/ecc/fixture/CheckpointFixture.kt
+++ b/sdk-lib/src/androidTest/java/cash/z/ecc/fixture/CheckpointFixture.kt
@@ -4,6 +4,7 @@ import cash.z.ecc.android.sdk.internal.model.Checkpoint
 import cash.z.ecc.android.sdk.internal.model.ext.KEY_EPOCH_SECONDS
 import cash.z.ecc.android.sdk.internal.model.ext.KEY_HASH
 import cash.z.ecc.android.sdk.internal.model.ext.KEY_HEIGHT
+import cash.z.ecc.android.sdk.internal.model.ext.KEY_IRONWOOD_TREE
 import cash.z.ecc.android.sdk.internal.model.ext.KEY_ORCHARD_TREE
 import cash.z.ecc.android.sdk.internal.model.ext.KEY_SAPLING_TREE
 import cash.z.ecc.android.sdk.internal.model.ext.KEY_VERSION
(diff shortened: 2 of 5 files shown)


nesence-m (approved):
lgtm, a real loophole in the code

sdk-lib/src/androidTest/java/cash/z/ecc/fixture/CheckpointFixture.kt
```diff
@@ -26,18 +27,21 @@ object CheckpointFixture {
...
 
+    @Suppress("LongParameterList")
     internal fun new(
         height: BlockHeight = HEIGHT,
         hash: String = HASH,
         time: Long = EPOCH_SECONDS,
         saplingTree: String = SAPLING_TREE,
-        orchardTree: String = ORCHARD_TREE
+        orchardTree: String = ORCHARD_TREE,
+        ironwoodTree: String? = null
```
nesence-m, line 37: The PR description says the parse-and-serialise round trip "cannot silently drop it again", but no test actually exercises a non-null `ironwoodTree`: this default is `null` and `CheckpointTest.deserialize` calls `CheckpointFixture.new()` as-is, so the new `toJson` branch and the parser's new block are only ever hit in the absent case. Reverting either one leaves the whole suite green. Suggest a round-trip test that passes `ironwoodTree = <the value from testnet/4134750.json>` and asserts equality after `Checkpoint.from`, and optionally a `CheckpointToolTest` case loading a real >= 4134750 testnet checkpoint and asserting the field comes back non-null.

lightwallet-client-lib/src/main/java/co/electriccoin/lightwallet/client/model/TreeStateUnsafe.kt
```diff
@@ -8,12 +8,14 @@ class TreeStateUnsafe(
...
 
+        @Suppress("LongParameterList")
         fun fromParts(
             height: Long,
             hash: String,
             time: Int,
             saplingTree: String,
-            orchardTree: String
+            orchardTree: String,
+            ironwoodTree: String?
```
nesence-m, line 18: `lightwallet-client-lib` is a published artifact and `fromParts` is public API, so adding a required parameter is a source-incompatible change for external callers. Consider `ironwoodTree: String? = null` here (and on `TreeState.fromParts`) to keep existing call sites compiling — absence already means "no Ironwood state", so a defaulted null hides nothing, unlike on the `Checkpoint` data class where forcing exhaustive construction is the point. If you'd rather keep it required so call sites must decide, the break should at least get a CHANGELOG entry the way the `TransactionOverview.zip318Kind` trailing-parameter change did.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/ext/CheckpointExt.kt
```diff
@@ -56,7 +59,18 @@ private fun Checkpoint.Companion.from(
                     "000000" // NON-NLS
                 }
 
-            return Checkpoint(height, hash, epochSeconds, saplingTree, orchardTree)
+            // Optional, unlike orchardTree: mainnet checkpoints do not carry an
+            // Ironwood tree state yet, so requiring it post-NU6.3 would reject
+            // every mainnet checkpoint currently shipped.
+            val ironwoodTree =
+                try {
+                    jsonObject.getString(Checkpoint.KEY_IRONWOOD_TREE)
```
nesence-m, line 67: Nit: the `getString` + catch shape is borrowed from the orchard block above, but there the catch is a genuine error path (warn + `require`); here absence is the normal case — true of 100% of bundled mainnet checkpoints — so every ordinary mainnet checkpoint parse constructs and verbose-logs a `JSONException`. `if (jsonObject.has(Checkpoint.KEY_IRONWOOD_TREE)) jsonObject.getString(...) else null` expresses an optional field without exception-driven control flow.

nesence-m (commented):
Reviewed against `maint/v2.7.x` locally (stacked PR #2101 excluded). The plumbing is correct end to end: both bundled copies of `service.proto` already declare `ironwoodTree = 7`, the hex string is passed through the protobuf builder unchanged exactly like the sapling/orchard fields, and the pinned librustzcash rev (`8755958`) parses `TreeState.ironwood_tree`, decoding an absent field as an empty frontier — so no Rust or Cargo changes are needed here. I also verified the optionality decision against the shipped data: every testnet checkpoint at or above the Nu6.3 activation height (4,134,000) carries the field (4134750 through 4194750), everything below it does not, and the newest bundled mainnet checkpoint (3427310) is below mainnet activation (3,428,143). So "absent == pre-activation == empty tree is correct" holds for the entire bundled set today.

Two notes that don't anchor to a diff line:

- The follow-up tightening deserves a tracked issue rather than a PR-description promise, because the failure it guards against is silent. Mainnet activation (3,428,143) is just past the newest bundled checkpoint; the first post-activation mainnet checkpoint generated without the field would seed an empty Ironwood frontier — wrong note positions and witnesses for any Ironwood note received after birthday, with no error surfaced anywhere. Closing that hole needs an `ironwoodActivationHeight` on `ZcashNetwork` plus the orchard-style `require`, and confirmation that the checkpoint tooling (Checkmate, per `docs/Architecture.md`) emits the field before post-activation mainnet checkpoints get bundled.
- No CHANGELOG entry. The `[Unreleased]` section documents both behavior changes and API-shape changes (see the `TransactionOverview.zip318Kind` entry); checkpoint-seeded wallets now carrying Ironwood tree state, and the `fromParts` signature changes, both qualify.

PR #2101, opened 2026-07-30
Show the Ironwood balance in the demo app

dannywillems:
Stacked on #2100. Last of the Ironwood audit findings, and the least
consequential: demo app only, no SDK behaviour change.

`AccountBalance` has carried an `ironwood: WalletBalance` since Ironwood
support landed, and **neither** balance screen displayed it. Both showed
Orchard, Sapling and transparent, so after a pool migration a demo-app user
could not see where their shielded funds actually were.

## Both screens, because both were wrong

- `ui/screen/balance/view/BalanceView.kt` — the Compose screen; a row mirroring
  the Orchard one, plus a `balance_ironwood` string.
- `demos/getbalance/GetBalanceFragment.kt` — the older fragment demo. I nearly
  skipped this one assuming it was single-pool, but its layout displays
  "Orchard balance", "Sapling balance" and "Transparent balance", so Ironwood
  was a genuine omission there too. Needed a layout entry, a flow collector and
  a handler to match the three already wired.

## Verification

- `:demo-app:compileZcashmainnetDebugKotlin` compiles (exit 0), which is what
  proves the generated `binding.ironwoodBalance` resolves from the new
  `@+id/ironwood_balance`.
- `ktlint` and `detektAll` pass.

### DIFF
--- a/demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/demos/getbalance/GetBalanceFragment.kt
+++ b/demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/demos/getbalance/GetBalanceFragment.kt
@@ -153,6 +153,21 @@ class GetBalanceFragment : BaseDemoFragment<FragmentGetBalanceBinding>() {
                             val account = it.getAccounts()[CURRENT_ZIP_32_ACCOUNT_INDEX.toInt()]
                             it.walletBalances.combine(it.exchangeRateUsd) { balances, rate ->
                                 balances?.let {
+                                    val walletBalance = balances[account.accountUuid]!!.ironwood
+                                    walletBalance to
+                                        rate.currencyConversion
+                                            ?.priceOfZec
+                                            ?.toBigDecimal()
+                                }
+                            }
+                        }.collect { onIronwoodBalance(it) }
+
+                    sharedViewModel.synchronizerFlow
+                        .filterNotNull()
+                        .flatMapLatest {
+                            val account = it.getAccounts()[CURRENT_ZIP_32_ACCOUNT_INDEX.toInt()]
+                            it.walletBalances.combine(it.exchangeRateUsd) { balances, rate ->
+                                balances?.let {
                                     val walletBalance = balances[account.accountUuid]!!.unshielded
                                     walletBalance to
                                         rate.currencyConversion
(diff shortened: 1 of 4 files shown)


nesence-m (dismissed):
utACK

nesence-m (commented):
Reviewed the delta over #2100 only (base `dw/checkpoint-ironwood-tree`, 4 files, +61). The Compose screen change is a faithful mirror of the existing Orchard block — string resource, `toZecString`/nullable `toUsdString` formatting and pool ordering all match, and every referenced symbol (`AccountBalance.ironwood`, `WalletBalance.pending`) already exists on the base branch, so the change compiles against the parent stack alone. The `strings.xml` and layout additions follow their neighbours' conventions (the hardcoded "Ironwood balance" label matches the sibling TextViews in `fragment_get_balance.xml`).

One functional issue in `GetBalanceFragment`: the new flow collector is sequenced after the orchard `.collect` inside the same `launch` block, so it never runs and the fragment's Ironwood text only updates via the `onStatus` path. Details and a suggested fix inline.

demo-app/src/main/java/cash/z/ecc/android/sdk/demoapp/demos/getbalance/GetBalanceFragment.kt
```diff
@@ -147,6 +147,21 @@ class GetBalanceFragment : BaseDemoFragment<FragmentGetBalanceBinding>() {
...
+                            it.walletBalances.combine(it.exchangeRateUsd) { balances, rate ->
+                                balances?.let {
+                                    val walletBalance = balances[account.accountUuid]!!.ironwood
+                                    walletBalance to
+                                        rate.currencyConversion
+                                            ?.priceOfZec
+                                            ?.toBigDecimal()
+                                }
+                            }
+                        }.collect { onIronwoodBalance(it) }
```
nesence-m, line 163: This new collector is unreachable. `Synchronizer.walletBalances` and `exchangeRateUsd` are `StateFlow`s, so the orchard chain's `.collect { onOrchardBalance(it) }` just above never returns — everything sequenced after it inside this `launch` block never executes. (The base branch already has this latent issue for the transparent collector; this inserts the Ironwood chain into the same dead spot.)

Concretely: the Ironwood row is only ever filled in by the `onStatus` path, i.e. on synchronizer status transitions. It shows "Calculating balance" until the first status change and can then go stale when balances or the exchange rate change without an accompanying status change — exactly the window a pool-migration user would be watching.

Suggested fix: wrap this chain in its own `launch { }` like the status/progress/sapling collectors above. While here, giving the transparent collector its own `launch` as well would fix the pre-existing instance of the same bug instead of propagating the pattern.

PR #2102, opened 2026-07-31
chore: improve release preparation

nuttycom:
## Summary

- rename the release starter to `prepare-release.sh` and consistently use candidate branches
- validate and rewrite release versions through tested, macOS-compatible helpers
- push both branches and open an issue-linked draft release PR automatically
- update release documentation and add an offline shell test suite

## Validation

- `scripts/tests/run-tests.sh` (18 assertions)
- `./scripts/ci-local.sh fast`
- `RUST_ANDROID_GRADLE_PYTHON_COMMAND=/usr/bin/python3 ./scripts/ci-local.sh quick`

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -29,6 +29,9 @@ jobs:
       - name: Checkout
         timeout-minutes: 1
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      - name: Release tooling tests
+        timeout-minutes: 1
+        run: ./scripts/tests/run-tests.sh
       - name: Gradle Wrapper Validation
         timeout-minutes: 1
         uses: gradle/actions/wrapper-validation@v6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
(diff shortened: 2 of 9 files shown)


nuttycom:
This PR should also update the manually-triggered release workflow to merge the release branch back to the maintenance branch. The only question there is how to handle tagging - probably this should be done by the manual release workflow, so that the exact commit whose sources are being published is tagged for the release.

zcash-harry (changes requested):
Could we address these two points before merging?

1. Like you already mentioned, I think we should update `.github/workflows/deploy-release.yml` so the manually triggered workflow tags the exact commit being published and then merges `release/vX.Y.Z` back into `maint/vX.Y.x`, or opens a back-merge PR if direct pushes are blocked. This would preserve the current `vX.Y.Z` tag format while ensuring it points to the exact release commit being published.

2. Skip `require_gh_auth` when `prepare-release.sh` is run with `--dry-run`. A dry run only previews actions and does not actually open a PR, so GitHub authentication should not be required.

nuttycom:
@Cosmos-Harry I've addressed your #2 in this PR; #1 I think we should defer - right now handling these operations is a relatively light manual lift, and more fully automating it is desirable but requires navigating some workflow state challenges, and the merge back to the maintenance branch can't reliably be automated because other commits may have landed on the maintenance branch after a release has begun (this is a desired feature - that once you begin a release, anything else that lands goes into the next release, not the current one) and since that may result in conflicts, we can't fully automate the merge.

zcash-harry (approved):
utACK

PR #2090, opened 2026-07-30
Merge maint/v2.8.x to `main` after v2.8.0-rc.3 release prep.

nuttycom:

PR #2109, opened 2026-07-31
Expose v_transactions.pool_crossing_value as TransactionOverview.poolCrossingValue

### LINKED ISSUE #2103: Expose v_transactions.pool_crossing_value so wallets can display ZIP 318 pool-crossing amounts correctly
`zcash_client_sqlite 0.22.0-rc.6` (adopted on `main` by #2090) added a `pool_crossing_value` column to the `v_transactions` view. It classifies wallet-internal transfers that move an account's own funds between shielded pools (e.g. ZIP 318 Orchard → Ironwood migration transfers) and reports the amount that crossed.

The upstream changelog gives explicit client guidance:

> For such a transaction `account_balance_delta` is just the negated fee, which is not a useful display quantity; clients should present `pool_crossing_value` as the migrated amount instead of deriving an amount from `total_spent` or `total_received`, both of which overstate the crossing whenever the transaction also returns change to a pool it spent from.

The SDK currently has no reference to this column anywhere — it is not read by `backend-lib`, not carried on `DbTransactionOverview`, and not surfaced on the public `TransactionOverview`. A wallet built on this SDK therefore cannot follow the upstream display guidance, and a ZIP 318 migration transfer will render with a misleading amount (or just the fee).

Follow-up work:
- Read `pool_crossing_value` through the FFI and expose it on `TransactionOverview` (nullable; non-null is the classification predicate per upstream).
- Document the display guidance in the API docs and CHANGELOG.

pacu:
Closes #2103. Resolves MOB-1615.

`zcash_client_sqlite 0.22.0-rc.6` (adopted in #2090) added a `pool_crossing_value` column to `v_transactions`, classifying wallet-internal transfers that move an account's own funds between shielded pools (e.g. ZIP 318 Orchard → Ironwood migration transfers) and reporting the amount that crossed. Its changelog directs clients to present `pool_crossing_value` as the migrated amount, because for such a transaction `account_balance_delta` is just the negated fee and `total_spent`/`total_received` overstate the crossing whenever change returns to a spent-from pool. The SDK previously read none of this.

Changes:
- `AllTransactionView` reads the column (the query already selects `*`, so only the cursor parser and column definition change).
- `DbTransactionOverview` carries it internally.
- **New public API**: `TransactionOverview.poolCrossingValue: Zatoshi?`, non-null exactly when the transaction is a wallet-internal pool-crossing transfer, with the upstream classification predicate and display guidance in the KDoc. The parameter defaults to `null`, so existing constructor call sites remain source compatible.
- `TransactionOverviewFixture` and CHANGELOG updated.

Verified with `:sdk-lib:compileDebugKotlin` (passes; native cargo tasks excluded, as this change touches no Rust). 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `TransactionOverview.poolCrossingValue`: the amount a wallet-internal transfer moved
+  between the account's own shielded pools (for example, a ZIP 318 Orchard -> Ironwood
+  migration transfer), read from the `pool_crossing_value` column that
+  `zcash_client_sqlite 0.22.0-rc.6` added to the `v_transactions` view; null for a
+  transaction that is not such a transfer. For a pool-crossing transfer, `netValue` is
+  just the fee, and `totalSpent`/`totalReceived` overstate the crossing whenever the
+  transaction also returns change to a pool it spent from, so wallets should display
+  `poolCrossingValue` as the migrated amount.
+
 ### Changed
 - Updated the librustzcash crates to `zcash_client_backend 0.24.0-rc.6` and
   `zcash_client_sqlite 0.22.0-rc.6`, adopting the revised ZIP 318 migration timing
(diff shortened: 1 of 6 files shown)


dannywillems:
This is a duplicate of https://github.com/zcash/zcash-android-wallet-sdk/pull/2098. Also, it should target `maint/v2.7.x`

pacu:
Closing as a duplicate of #2098, which covers the same `pool_crossing_value` exposure plus the other dropped `v_transactions` columns and `TxOutputsView`, and correctly targets `maint/v2.7.x`. Left a note there with the two small pieces from this PR that may be worth lifting.

PR #2110, opened 2026-07-31
Surface ProposalError::AnchorNotFound as TransactionEncoderException.AnchorNotFoundException

### LINKED ISSUE #2104: AnchorIntervalMismatch / AnchorNotFound have no typed Kotlin surface; they arrive as opaque RustLayerException messages
The Unreleased CHANGELOG (after #2090) tells wallets on a test network that a pool migration planned under a custom anchor grid will be reported as `AnchorIntervalMismatch` "and must be re-planned", and describes proposal-time behavior around `ProposalError::AnchorNotFound`.

Neither identifier exists anywhere in the SDK: not in `backend-lib`'s Rust error mapping and not in the Kotlin exception hierarchy. Both conditions will surface to consumers as a generic `RustLayerException` (or `CreateTransactionException`-style failure) whose only signal is the formatted message string from `anyhow`.

Consequences:
- A caller cannot programmatically distinguish "this migration must be re-planned" from any other proposal failure; string-matching the message is the only option, and message text is not a stable contract.
- There is no API affordance for the prescribed recovery action (re-planning the migration).

Follow-up work:
- Map `AnchorIntervalMismatch` (and ideally `AnchorNotFound`, for diagnostics) to typed errors at the JNI boundary and expose corresponding typed exceptions/sealed results in the Kotlin API.
- Document the recovery path for the mismatch case.

Mainly affects test networks (the mainnet backfill is exact), but the error contract gap is structural.

pacu:
Closes #2104.

A proposal failing with `zcash_client_backend`'s `ProposalError::AnchorNotFound` previously reached callers as a generic `RuntimeException` whose only signal was the formatted Rust message — no way to distinguish "sync further and retry" from any other proposal failure without string matching.

**Rust (`backend-lib`)**: a new `map_proposal_error` helper inspects the concrete proposal error at the JNI boundary. For `Error::Proposal(ProposalError::AnchorNotFound(height))` it constructs and throws a dedicated `ProposalAnchorNotFoundException(message, anchorHeight)` on the JVM, relying on `unwrap_exc_or`'s existing pending-exception check so the typed throw wins and the generic `RuntimeException` remains the fallback. Applied to `proposeTransfer`, `proposeTransferFromUri`, and the Orchard → Ironwood migration proposal (which now takes the `JNIEnv` for error mapping). `proposeShielding` is deliberately untouched: it spends only transparent inputs, where the anchor lookup does not apply.

**Kotlin (`sdk-lib`)**: `TransactionEncoderImpl` translates the internal exception into the new public `TransactionEncoderException.AnchorNotFoundException(anchorHeight: BlockHeight)`, whose KDoc documents the recovery path (scanning creates a checkpoint at every height a proposal can anchor to, so syncing further before retrying is expected to resolve it). `@Throws` contracts updated on the three encoder methods.

**Scope note on `AnchorIntervalMismatch`** (the other error named in #2104): it originates in the `zcash_pool_migration` engine's prove/rebuild steps (`ProveError`/`RebuildError`), and `backend-lib` does not currently invoke that engine anywhere — the SDK's migration entry point is a pinned `propose_send_max_transfer` call. There is nothing for a typed surface to attach to until the engine is wired in, so this PR clarifies the inherited CHANGELOG bullet accordingly instead of adding dead error plumbing. Details on the issue.

Verification: `cargo check` on `backend-lib` passes cleanly; `:sdk-lib:compileDebugKotlin` and `:backend-lib:compileDebugKotlin` pass (native cargo linking excluded — no NDK toolchain in this environment).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `TransactionEncoderException.AnchorNotFoundException`, thrown by the proposal APIs
+  (`Synchronizer.proposeTransfer`, `proposeTransferFromUri` and
+  `proposeOrchardToIronwoodMigration`) when a proposal fails because no anchor is
+  computable at the height the proposal would anchor to (`zcash_client_backend`'s
+  `ProposalError::AnchorNotFound`). This failure previously surfaced as the generic
+  proposal exception, identifiable only by matching the formatted Rust error message.
+  Scanning creates a checkpoint at every height a proposal can anchor to, so syncing
+  further before retrying is expected to resolve it.
+
 ### Changed
 - Updated the librustzcash crates to `zcash_client_backend 0.24.0-rc.6` and
   `zcash_client_sqlite 0.22.0-rc.6`, adopting the revised ZIP 318 migration timing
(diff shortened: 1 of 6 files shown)


pacu:
Superseded by #2114, which carries the identical commits cherry-picked onto `maint/v2.7.x` per the maintenance-branch workflow (the touched files are the same on both branches, and all static/unit checks passed here before the move).

PR #2113, opened 2026-07-31
Surface the v_transactions ZIP 318 classification

dannywillems:
Stacked on #dw/views-columns-up-to-date, which surfaces the other
`v_transactions` columns the SDK was dropping. Review that one first.

## What this adds

Upstream librustzcash adds a `zip318_kind` column to `v_transactions`
(zcash/librustzcash#2874). It records how a transaction classifies against
ZIP 318, the Orchard to Ironwood pool migration, so a wallet can label a
migration transaction in its history without a migration plan, which does not
survive a seed restore.

`Zip318Kind` carries the five values, and `TransactionOverview.zip318Kind`
exposes it.

| code | case | meaning |
| --- | --- | --- |
| 0 | `NOT_CLASSIFIED` | the column default; nothing has looked at this transaction yet |
| 1 | `NONCONFORMING` | classified, and it is not a ZIP 318 transaction |
| 2 | `PREPARATION` | a note-preparation self-send a migration run makes before crossing |
| 3 | `TRANSFER` | a pool crossing paying the account's own internal address |
| 4 | `CROSSING_PAYMENT` | the same shape, paying an external third party |

## Three things a consumer must get right

**`NOT_CLASSIFIED` is not a decision.** It means "we never looked", not "this
is not a migration". Render it as no label at all. Presenting it as "not a
migration" would claim the wallet examined a transaction it never looked at.
That is why 0 and 1 are distinct codes rather than one falsy value.

**Only `PREPARATION` and `TRANSFER` are the wallet's own migration.**
`CROSSING_PAYMENT` has the identical on-chain shape by design, because the
ordinary send path builds such payments deliberately so that they join the
migration anonymity set. It is a payment to a third party and must not be
shown as a migration.

**This names a conformance class, never a provenance.** ZIP 318's privacy
argument is that every migration transaction looks like every other one, so
recognizing the shape places a transaction in the anonymity set and can never
establish which wallet's run produced it.

## Decoding

`Zip318Kind.new(code: Int?)` is total: null, an unrecognized code, or an absent
column all yield `NOT_CLASSIFIED`, never a throw and never `NONCONFORMING`. A
future librustzcash may add codes, and an SDK that meets one has learned
nothing about the transaction, so saying so is the honest answer. Defaulting to
a decision would assert something the SDK never established.

The absent-column case matters today: the pinned `zcash_client_sqlite
0.22.0-rc.6` does not have this column yet, the view is read with `SELECT *`,
and an unguarded `cursor.getInt(-1)` would throw
`CursorIndexOutOfBoundsException` on every transaction-history read. The parse
checks the column index for that reason.

## Dependency, and a commit that must be reverted

The last commit pins the librustzcash crates to the revision of
zcash/librustzcash#2874, so the column actually exists and this code is
exercised rather than silently reporting `NOT_CLASSIFIED` for everything.

**That pin must be reverted before this merges.** It must not outlive the
upstream release carrying the column. It is pinned to a specific revision
rather than a branch so the build stays reproducible while the upstream branch
moves.

## Testing

`./scripts/ci-local.sh quick` (detekt, ktlint, unit tests) passes. The Rust
stage was not run; this change touches no Rust and does not cross the JNI
boundary. Note the unit tests do not exercise a real `v_transactions` row, so
they give no signal on the decode path against a live database.

### DIFF
--- a/backend-lib/Cargo.toml
+++ b/backend-lib/Cargo.toml
@@ -77,16 +77,19 @@ xz2 = { version = "0.1", features = ["static"] }
 component = "0.1.1"
 rust-analyzer = "0.0.1"
 
-## Uncomment this to test someone else's librustzcash changes in a branch
-#[patch.crates-io]
-#pczt = { git = "https://github.com/zcash/librustzcash.git", branch = "main" }
-#transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash", branch = "main" }
-#zcash_address = { git = "https://github.com/zcash/librustzcash", branch = "main" }
-#zcash_client_backend = { git = "https://github.com/zcash/librustzcash", branch = "main" }
-#zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash", branch = "main" }
-#zcash_primitives = { git = "https://github.com/zcash/librustzcash", branch = "main" }
-#zcash_proofs = { git = "https://github.com/zcash/librustzcash", branch = "main" }
-#zcash_protocol = { git = "https://github.com/zcash/librustzcash", branch = "main" }
+## Pinned to the librustzcash branch of zcash/librustzcash#2874, which adds the
+## `zip318_kind` column this SDK reads. Revert to the crates.io versions once that
+## has merged and been released.
+[patch.crates-io]
+pczt = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
 
 [lib]
 name = "zcashwalletsdk"
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Zip318Kind.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Zip318Kind.kt
@@ -1,0 +1,49 @@
+package cash.z.ecc.android.sdk.model
+
+/**
+ * How a transaction classifies against ZIP 318, the Orchard to Ironwood pool migration.
+ *
+ * This is a conformance class, not a provenance: it describes the shape a transaction has on
+ * chain, and can never establish that a transaction came from this wallet's own migration run.
+ * Only [PREPARATION] and [TRANSFER] are the wallet's own migration; [CROSSING_PAYMENT] has the
+ * same shape by design but pays a third party.
+ */
+@Suppress("MagicNumber")
+enum class Zip318Kind(
+    val code: Int
+) {
+    /**
+     * The transaction has not been classified, either because it predates the classification or
+     * because the wallet has not decrypted it yet. It says nothing about whether the transaction
+     * is a ZIP 318 transaction; deciding that requires rescanning it. Present it with no label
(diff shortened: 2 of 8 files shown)


nesence-m (approved):
utACK

backend-lib/Cargo.toml
```diff
@@ -77,16 +77,19 @@ xz2 = { version = "0.1", features = ["static"] }
...
-#zcash_address = { git = "https://github.com/zcash/librustzcash", branch = "main" }
-#zcash_client_backend = { git = "https://github.com/zcash/librustzcash", branch = "main" }
-#zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash", branch = "main" }
-#zcash_primitives = { git = "https://github.com/zcash/librustzcash", branch = "main" }
-#zcash_proofs = { git = "https://github.com/zcash/librustzcash", branch = "main" }
-#zcash_protocol = { git = "https://github.com/zcash/librustzcash", branch = "main" }
+## Pinned to the librustzcash branch of zcash/librustzcash#2874, which adds the
+## `zip318_kind` column this SDK reads. Revert to the crates.io versions once that
+## has merged and been released.
+[patch.crates-io]
```
nesence-m, line 83: The PR description says this pin "must be reverted before this merges", but it merged with the stanza intact — it is now live on `maint/v2.7.x` (still present after #2094 landed on top). That branch is the 2.7.x release line, and no release can be cut while a `[patch.crates-io]` stanza exists in this manifest: it would ship an unreleased state of the librustzcash crates. Until the revert lands, every 2.7.x release is blocked on someone remembering this comment.

Suggest either reverting the pin commit now (accepting that `zip318Kind` reads `NOT_CLASSIFIED` for everything until the upstream release carries the column) or opening a release-blocking tracking issue for its removal, so a 2.7.x release can't ship rev `8755958` by accident.

backend-lib/Cargo.toml
```diff
@@ -77,16 +77,19 @@ xz2 = { version = "0.1", features = ["static"] }
...
+pczt = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
+zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
```
nesence-m, line 88: A second exposure while this pin is live: rev `8755958` is an unmerged PR branch (zcash/librustzcash#2874) that appends two new wallet-DB migrations (`zip318_classification`, `v_transactions_zip318_kind`), and every dev/tester wallet DB opened by a build of this branch gets stamped with them.

If #2874 is revised before it merges, those DBs are stranded either way: a changed migration UUID makes the eventually-released chain treat the migration as unapplied and re-run `ALTER TABLE transactions ADD COLUMN zip318_kind` against a table that already has the column — a duplicate-column failure at DB open, wedging the wallet until it's wiped — while an in-place amendment under the same UUID leaves silent schema drift instead. This branch has no schema-heal mechanism (`backend-lib/src/main/rust/migration.rs` has none), so it's worth being deliberate about which builds cut from `maint/v2.7.x` reach testers before the upstream PR is final.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/Zip318Kind.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    companion object {
+        /**
+         * Decodes the `zip318_kind` column value.
+         *
+         * An unrecognized code decodes to [NOT_CLASSIFIED]: a newer librustzcash may classify a
+         * transaction in a way this SDK does not know, and having learned nothing about it is the
+         * honest reading, rather than claiming it is [NONCONFORMING].
+         */
+        internal fun new(code: Int?): Zip318Kind = entries.firstOrNull { it.code == code } ?: NOT_CLASSIFIED
```
nesence-m, line 47: This decode table is load-bearing and currently untested: nothing pins the entries' codes to upstream `Zip318Classification::to_code` (Unknown=0, Nonconforming=1, Preparation=2, Transfer=3, CanonicalCrossingPayment=4 — I checked these against `components/zcash_protocol/src/zip318.rs` at the pinned rev and they match), nor the null / unrecognized-code → `NOT_CLASSIFIED` fallback the PR description leans on. A reordering or renumbering here would silently mislabel history rows and no test would fail.

A small unit test over the five codes plus `null` and an out-of-range value (e.g. `5`, `-1`) would lock the contract down cheaply.

nesence-m (commented):
Post-merge review — the substance here checks out. I verified the Kotlin decode table against `Zip318Classification::to_code` at the pinned librustzcash rev (codes 0–4 match, and the unrecognized-code → `NOT_CLASSIFIED` fallback mirrors upstream `from_code`); the absent-column guard in `AllTransactionView` is genuinely required, since the view is read via `SELECT *` and `getIntOrNull(-1)` throws when the column is missing; and the `Cargo.lock` diff is source/checksum-only — no version bumps, and no duplicated `zcash_*`/`orchard`/`rusqlite` entries, so the patch stanzas apply cleanly across the whole graph. The remaining concerns are about the pin's lifecycle on this release branch rather than the code — see the inline comments.

One note that doesn't anchor to a diff line: the trailing constructor parameter on `TransactionOverview`/`DbTransactionOverview` is binary-incompatible for consumers constructing or `copy`ing positionally. The CHANGELOG covers the source-level story well; it's just worth accounting for when choosing the next version number cut from this branch.

Issue #680, opened 2022-08-19
WalletTransactionEncoder - createShieldingSpend() - improve implementation


HonzaR:
## Is your feature request related to a problem? Please describe.
Filling an additional issue to provide a link to the code base TODO occurrence.

We'd like to enhance try/catch in `WalletTransactionEncoder#createShieldingSpend` to: If this error matches: Insufficient balance (have 0, need 1000 including fee) then consider custom error that says no UTXOs existed to shield

str4d:
We are now doing this in the new `TransactionEncoder.proposeShielding` method by returning `Proposal?`, with `null` meaning either there were no UTXOs to shield, or the balance to shield was below the shielding threshold. `createShieldingSpend` will be removed in 2.1.0 as its corresponding `Synchronizer` method is being deprecated.

daira:
> If this error matches: Insufficient balance (have 0, need 1000 including fee) then consider custom error that says no UTXOs existed to shield.

That is dependent on the pre-ZIP-317 fee amount. In general it is fragile to match on error strings.

HonzaR:
Closing, as the related functions have been removed.

PR #2114, opened 2026-07-31
Surface ProposalError::AnchorNotFound as TransactionEncoderException.AnchorNotFoundException

### LINKED ISSUE #2104: AnchorIntervalMismatch / AnchorNotFound have no typed Kotlin surface; they arrive as opaque RustLayerException messages

pacu:
Closes #2104. Resolves MOB-1616. Replaces #2110, which targeted `main`; per the maintenance-branch workflow this lands on `maint/v2.7.x` and merges forward.

A failure with `zcash_client_backend`'s `ProposalError::AnchorNotFound` previously reached callers only as a generic exception whose sole signal was the formatted Rust message — no way to distinguish "sync further and re-propose" from any other failure without string matching. This PR surfaces it as a typed exception on the paths that can actually produce it.

**Where the error can occur.** `AnchorNotFound` is constructed only inside `build_proposed_transaction` (`data_api/wallet.rs`), reachable from `create_proposed_transactions` and `create_pczt_from_proposal` — never from the propose entry points: proposal planning selects the anchor *height* but does not compute the root at it. The first revision of this PR instrumented the three propose entry points; per review, that mapping could never fire and has been removed. The PCZT path is included because it shares `build_proposed_transaction` and would otherwise keep the same opaque failure.

**Rust (`backend-lib`)**: `map_proposal_error` inspects the concrete error at the JNI boundary of `createProposedTransactions` and `createPcztFromProposal` and throws a dedicated `ProposalAnchorNotFoundException(message, anchorHeight)` on the JVM. The construct-and-throw plumbing is a new `utils::exception::throw_object` helper next to `unwrap_exc_or`, documenting the shared pending-exception contract; when a typed throw fails it clears the pending JVM exception, so the generic `RuntimeException` remains the fallback in every failure mode rather than leaking an unrelated `NoClassDefFoundError`-style error.

**Kotlin (`sdk-lib`)**:
- `Synchronizer.createProposedTransactions` now throws the public `TransactionEncoderException.AnchorNotFoundException(anchorHeight, rootCause)`, declared along the whole chain (`TransactionEncoder`, the impl, `SdkSynchronizer`, and the `Synchronizer` KDoc). It is a sibling of `TransactionNotCreatedException`, so callers handling this case need a dedicated catch; the CHANGELOG states the call-site edit.
- `Synchronizer.createPcztFromProposal` keeps throwing `PcztException.CreatePcztFromProposalException`, now carrying the same public typed exception as its `cause` (the internal JNI class is not API surface).
- The JNI-constructed exception class is `@Keep`-annotated, matching the `Jni*` precedent, so R8 cannot strip the natively-called constructor in minified consumer builds.
- The `runCatching` chains around the proposal and creation calls rethrow `CancellationException` instead of wrapping it as a failure.
- KDoc documents the recovery: scanning creates a checkpoint at every height a proposal can anchor to, so sync further and then create a **new** proposal — the failed proposal anchors to the same height, so retrying it unchanged is not expected to succeed on its own.

**Scope note on `AnchorIntervalMismatch`** (the other error named in #2104): unchanged from the first revision — it originates in the `zcash_pool_migration` engine's prove/rebuild steps, which `backend-lib` does not invoke, so there is nothing to surface until that engine is wired in.

Also fixed in passing, per review: `proposeTransferFromUri`'s tracing span was copy-paste named `RustBackend.proposeTransfer`.

Rebased onto current `maint/v2.7.x` (`zcash_client_backend 0.24.0-rc.7`; the `AnchorNotFound` construction sites are unchanged between rc.6 and rc.7). #2167 tracks the follow-up of routing all proposal/creation failure mapping through a single choke point.

Verification: `cargo check` and `cargo fmt --check` pass; `:sdk-lib:compileDebugKotlin` and `:backend-lib:compileDebugKotlin` pass (native NDK link excluded in this environment); `./gradlew ktlint` and `./gradlew detektAll` pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `TransactionEncoderException.AnchorNotFoundException`, thrown by the proposal APIs
+  (`Synchronizer.proposeTransfer`, `proposeTransferFromUri` and
+  `proposeOrchardToIronwoodMigration`) when a proposal fails because no anchor is
+  computable at the height the proposal would anchor to (`zcash_client_backend`'s
+  `ProposalError::AnchorNotFound`). This failure previously surfaced as the generic
+  proposal exception, identifiable only by matching the formatted Rust error message.
+  Scanning creates a checkpoint at every height a proposal can anchor to, so syncing
+  further before retrying is expected to resolve it.
+
 ### Changed
 - Updated the librustzcash crates to `zcash_client_backend 0.24.0-rc.6` and
   `zcash_client_sqlite 0.22.0-rc.6`, adopting the revised ZIP 318 migration timing
@@ -29,7 +39,9 @@ All of the following were picked up from the librustzcash update:
   a pool migration was in progress. A new database migration adds the missing
   column. The backfilled value is exact on the production network; on a test
   network, a pool migration planned under a custom anchor grid is reported as
-  `AnchorIntervalMismatch` and must be re-planned.
+  `AnchorIntervalMismatch` and must be re-planned. (That report originates in the
+  `zcash_pool_migration` engine's prove and rebuild steps, which this SDK does not
+  currently invoke, so it is not surfaced through this SDK's API.)
 - A ZIP 318 crossing anchored to a bucket boundary whose block contains no note
   commitments in any pool no longer fails with `ProposalError::AnchorNotFound`:
   scanning now creates a checkpoint at every anchor-retention grid height, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `TransactionEncoderException.AnchorNotFoundException`, thrown by
+  `Synchronizer.createProposedTransactions` and `Broadcaster.createProposedTransactions`
+  when transactions cannot be created from a proposal because no anchor is computable at
+  the height the proposal anchors to
+  (`zcash_client_backend`'s `ProposalError::AnchorNotFound`). When
+  `Synchronizer.createPcztFromProposal` fails for the same reason, the
+  `CreatePcztFromProposalException` it throws now carries this exception as its `cause`.
+  The failure previously surfaced only as the generic
+  `TransactionEncoderException.TransactionNotCreatedException` (or the generic PCZT
+  exception), identifiable only by matching the formatted Rust error message. The new
+  exception is a sibling of `TransactionNotCreatedException`, not a subtype: code that
+  catches `TransactionNotCreatedException` around `createProposedTransactions` and
+  needs to handle this case must add a catch for the new exception. Scanning
+  creates a checkpoint at every height a proposal can anchor to, so the expected
+  recovery is to sync further and then create a new proposal; the failed proposal
+  anchors to the same height, so retrying it unchanged is not expected to succeed on
+  its own.
+
 ## [2.7.0] - 2026-08-20
 
 ### Added
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/ProposalAnchorNotFoundException.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/ProposalAnchorNotFoundException.kt
@@ -1,0 +1,15 @@
+package cash.z.ecc.android.sdk.internal.jni
+
+/**
+ * Thrown across the JNI boundary when a transaction proposal fails because no anchor is
+ * computable at the height the proposal would anchor to (`zcash_client_backend`'s
+ * `ProposalError::AnchorNotFound`).
+ *
+ * This is a distinct type so that `sdk-lib` can surface a typed error instead of matching
+ * message text. It is constructed from native code: the `(String, Long)` constructor
+ * signature is part of the JNI contract with `map_proposal_error` in `lib.rs`.
+ */
+class ProposalAnchorNotFoundException(
+    message: String,
+    val anchorHeight: Long
+) : RuntimeException(message)
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/ProposalAnchorNotFoundException.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/ProposalAnchorNotFoundException.kt
@@ -1,0 +1,19 @@
+package cash.z.ecc.android.sdk.internal.jni
+
+import androidx.annotation.Keep
+
+/**
+ * Thrown across the JNI boundary when creating transactions from a proposal fails because no
+ * anchor is computable at the height the proposal anchors to (`zcash_client_backend`'s
+ * `ProposalError::AnchorNotFound`).
+ *
+ * This is a distinct type so that `sdk-lib` can surface a typed error instead of matching
+ * message text. It is constructed from native code, which is why it must be kept: the class
+ * name and the `(String, Long)` constructor signature are part of the JNI contract with
+ * `map_proposal_error` in `lib.rs`, and are never referenced from bytecode.
+ */
+@Keep
+class ProposalAnchorNotFoundException(
+    message: String,
+    val anchorHeight: Long
+) : RuntimeException(message)
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -47,6 +47,7 @@ use zcash_client_backend::{
         TransactionStatusFilter, TransparentKeyOrigin, WalletCommitmentTrees, WalletRead,
         WalletSummary, WalletWrite, Zip32Derivation,
         chain::{CommitmentTreeRoot, ScanSummary, scan_cached_blocks},
+        error::Error as DataApiError,
         scanning::{ScanPriority, ScanRange},
         wallet::{
             self, SignerView, create_pczt_from_proposal, create_proposed_transactions,
@@ -61,6 +62,7 @@ use zcash_client_backend::{
         DecodingError, Era, ReceiverRequirement, ReceiverRequirementError, UnifiedAddressRequest,
         UnifiedFullViewingKey, UnifiedSpendingKey,
     },
+    proposal::ProposalError,
     proto::{proposal::Proposal, service::TreeState},
     tor::{
         DormantMode,
@@ -2107,6 +2109,47 @@ fn zip317_helper<DbT>(
     )
 }
 
+/// The JVM class thrown when creating transactions from a proposal fails with
+/// [`ProposalError::AnchorNotFound`]. The `(String, long)` constructor signature is part of
+/// the JNI contract with the Kotlin class.
+const ANCHOR_NOT_FOUND_EXCEPTION_CLASS: &str =
+    "cash/z/ecc/android/sdk/internal/jni/ProposalAnchorNotFoundException";
+
+/// Converts a failure from building transactions (or a PCZT) out of a proposal into the
+/// `anyhow` error reported through the generic exception path, first raising a typed
+/// `ProposalAnchorNotFoundException` on the JVM when the failure is
+/// [`ProposalError::AnchorNotFound`], so that callers can react to it without matching
+/// message text. `utils::exception::throw_object` owns the pending-exception contract: the
+/// typed throw wins when it succeeds and the generic `RuntimeException` remains the fallback
+/// when it does not.
+fn map_proposal_error<DE, TE, SE, FE, CE, N>(
+    env: &mut JNIEnv,
+    context: &str,
+    e: DataApiError<DE, TE, SE, FE, CE, N>,
+) -> anyhow::Error
+where
+    DataApiError<DE, TE, SE, FE, CE, N>: std::fmt::Display,
+{
+    if let DataApiError::Proposal(ProposalError::AnchorNotFound(anchor_height)) = &e {
+        let description = format!(
+            "{}: no anchor is computable at height {}",
+            context, anchor_height
+        );
+        utils::exception::throw_object(env, ANCHOR_NOT_FOUND_EXCEPTION_CLASS, |env| {
+            let message = env.new_string(&description)?;
+            env.new_object(
+                ANCHOR_NOT_FOUND_EXCEPTION_CLASS,
+                "(Ljava/lang/String;J)V",
+                &[
+                    JValue::Object(&message),
+                    JValue::Long(u32::from(*anchor_height) as jlong),
+                ],
+            )
+        });
+    }
+    anyhow!("{}: {}", context, e)
+}
+
 #[unsafe(no_mangle)]
 pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeTransferFromUri<
     'local,
@@ -2119,7 +2162,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeTr
     network_id: jint,
 ) -> jbyteArray {
     let res = catch_unwind(&mut env, |env| {
-        let _span = tracing::info_span!("RustBackend.proposeTransfer").entered();
+        let _span = tracing::info_span!("RustBackend.proposeTransferFromUri").entered();
         let network = parse_network(network_id as u32)?;
         let mut db_data = wallet_db(env, network, db_data)?;
         let account_uuid = account_id_from_jni(env, account_uuid)?;
@@ -2439,7 +2482,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_createPro
             &proposal,
             expiry_height,
         )
-        .map_err(|e| anyhow!("Error while creating transactions: {}", e))?;
+        .map_err(|e| map_proposal_error(env, "Error while creating transactions", e))?;
 
         Ok(
             utils::rust_vec_to_java(env, txids.into(), "[B", |env, txid| {
@@ -2494,7 +2537,9 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_createPcz
                 expiry_height,
                 orchard_pool_padding,
             )
-            .map_err(|e| anyhow!("Error creating PCZT from single-step proposal: {}", e))?;
+            .map_err(|e| {
+                map_proposal_error(env, "Error creating PCZT from single-step proposal", e)
+            })?;
 
             Ok(utils::rust_bytes_to_java(
                 env,
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -2290,8 +2290,12 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeOr
         let mut db_data = wallet_db(env, network, db_data)?;
         let account_uuid = account_id_from_jni(env, account_uuid)?;
 
-        let proposal =
-            crate::migration::propose_orchard_to_ironwood(env, &mut db_data, &network, account_uuid)?;
+        let proposal = crate::migration::propose_orchard_to_ironwood(
+            env,
+            &mut db_data,
+            &network,
+            account_uuid,
+        )?;
 
         Ok(utils::rust_bytes_to_java(
             env,
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -13,7 +13,7 @@ use bytes::Bytes;
 use http_body_util::BodyExt;
 use jni::{
     JNIEnv,
-    objects::{JByteArray, JClass, JObject, JObjectArray, JString, JValue},
+    objects::{JByteArray, JClass, JObject, JObjectArray, JString, JThrowable, JValue},
     sys::{JNI_FALSE, JNI_TRUE, jboolean, jbyteArray, jint, jlong, jobject, jobjectArray, jstring},
 };
 use nonempty::NonEmpty;
@@ -47,6 +47,7 @@ use zcash_client_backend::{
         TransactionStatusFilter, TransparentKeyOrigin, WalletCommitmentTrees, WalletRead,
         WalletSummary, WalletWrite, Zip32Derivation,
         chain::{CommitmentTreeRoot, ScanSummary, scan_cached_blocks},
+        error::Error as DataApiError,
         scanning::{ScanPriority, ScanRange},
         wallet::{
             self, SignerView, create_pczt_from_proposal, create_proposed_transactions,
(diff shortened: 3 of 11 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Added
+- `TransactionEncoderException.AnchorNotFoundException`, thrown by the proposal APIs
+  (`Synchronizer.proposeTransfer`, `proposeTransferFromUri` and
+  `proposeOrchardToIronwoodMigration`) when a proposal fails because no anchor is
+  computable at the height the proposal would anchor to (`zcash_client_backend`'s
+  `ProposalError::AnchorNotFound`). This failure previously surfaced as the generic
+  proposal exception, identifiable only by matching the formatted Rust error message.
+  Scanning creates a checkpoint at every height a proposal can anchor to, so syncing
+  further before retrying is expected to resolve it.
```
nesence-m: This claim doesn't hold for the crate version the lockfile actually resolves. In `zcash_client_backend 0.24.0-rc.6`, `ProposalError::AnchorNotFound` is constructed at exactly three sites (`data_api/wallet.rs:1760/1805/1858`), all inside the private `build_proposed_transaction`, which is reachable only from `create_proposed_transactions` — never from `propose_transfer`, the ZIP-321 propose path, or `propose_send_max_transfer`.

So a wallet that hits the missing-anchor state fails inside `RustBackend.createProposedTransactions`, whose `map_err` (`lib.rs:2490`) still produces the generic `anyhow` error → `TransactionNotCreatedException` on the Kotlin side. The three `map_proposal_error` call sites added here sit on entry points that can't produce the error, and the exception this bullet advertises never fires.

Could you either wire the mapping into `createProposedTransactions` (adjusting the Kotlin surface on that path accordingly) or point me at a propose-time path in rc.6 that constructs `AnchorNotFound`? If the propose-time instrumentation is meant for a future crate generation, the bullet and the `@Throws` additions should say so — as written they describe behavior the SDK doesn't have.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/ProposalAnchorNotFoundException.kt
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Thrown across the JNI boundary when a transaction proposal fails because no anchor is
+ * computable at the height the proposal would anchor to (`zcash_client_backend`'s
+ * `ProposalError::AnchorNotFound`).
+ *
+ * This is a distinct type so that `sdk-lib` can surface a typed error instead of matching
+ * message text. It is constructed from native code: the `(String, Long)` constructor
+ * signature is part of the JNI contract with `map_proposal_error` in `lib.rs`.
+ */
+class ProposalAnchorNotFoundException(
```
nesence-m, line 16: This class is constructed from native code by hard-coded class-name string, but has no `@Keep` annotation and no ProGuard rule covering it — `proguard-consumer.txt` keeps only classes with native methods, and `proguard-project.txt` keeps `internal.model.*` plus `RustBackend`. Every other JNI-constructed class here (`JniAccount`, `JniWalletSummary`, `JniBlockMeta`, `JniEip681TransactionRequest`) carries `@Keep`.

In a `minifyEnabled` consumer build, R8 renames the class and strips the `(String, long)` constructor (never called from bytecode), so `env.new_object(...)` fails and leaves a pending `NoClassDefFoundError` that `unwrap_exc_or` passes through — release builds get an `Error` that escapes `catch (e: Exception)` instead of either the typed or the generic exception. Debug builds work, which makes this easy to miss.

Please add `@Keep` on the class (matching the `Jni*` precedent) or a `-keep` rule with `<init>(java.lang.String, long)` in `proguard-consumer.txt`.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/TransactionEncoderImpl.kt
```diff
@@ -39,9 +40,13 @@ internal class TransactionEncoderImpl(
...
      * @return the proposal or an exception
      *
+     * @throws TransactionEncoderException.AnchorNotFoundException
      * @throws TransactionEncoderException.ProposalFromUriException
      */
-    @Throws(TransactionEncoderException.ProposalFromUriException::class)
+    @Throws(
+        TransactionEncoderException.AnchorNotFoundException::class,
+        TransactionEncoderException.ProposalFromUriException::class
+    )
```
nesence-m: The contract update stops at this internal impl: the `TransactionEncoder` interface (lines 25/43/54), `SdkSynchronizer` (`@Throws` at 995/1006/1018), and the `Synchronizer` interface KDoc (`Synchronizer.kt:318`) all still declare only `ProposalFromUriException`/`ProposalFromParametersException`.

That matters beyond documentation: `AnchorNotFoundException` is a sibling of `ProposalFromParametersException` under the sealed class, not a subtype. Existing app code written to the documented contract — `catch (e: ProposalFromParametersException)` around `proposeTransfer`/`proposeOrchardToIronwoodMigration` — will no longer catch the under-synced case; the one failure this PR makes recoverable becomes the one that escapes uncaught. The interface and impl `@Throws` lists also diverge now. Please extend the declarations up the whole public chain.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2107,6 +2109,48 @@ fn zip317_helper<DbT>(
...
+                "(Ljava/lang/String;J)V",
+                &[
+                    JValue::Object(&message),
+                    JValue::Long(u32::from(*anchor_height) as jlong),
+                ],
+            )?;
+            env.throw(JThrowable::from(exception))
+        })();
+        if let Err(err) = thrown {
+            error!("Unable to throw ProposalAnchorNotFoundException: {}", err);
```
nesence-m: When the throw itself fails, `jni-rs` returns `Err` but leaves the *causing* exception (e.g. `NoClassDefFoundError`, `NoSuchMethodError`, OOME) pending on the env — it documents that it does not clear it. This branch only logs, so `unwrap_exc_or` then sees `exception_check() == true` and skips the generic fallback: the caller gets the raw JVM error and the proposal-error message is lost entirely. That contradicts the doc comment above ("the generic `RuntimeException` remains the fallback when it does not [succeed]") — it currently holds only when the failure is pure-Rust.

```rust
if let Err(err) = thrown {
    error!("Unable to throw ProposalAnchorNotFoundException: {}", err);
    let _ = env.exception_clear();
}
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/TransactionEncoderImpl.kt
```diff
@@ -293,3 +304,21 @@ internal class TransactionEncoderImpl(
...
+private inline fun Throwable.asProposalException(
+    fallback: () -> TransactionEncoderException
+): TransactionEncoderException =
+    when (this) {
+        is ProposalAnchorNotFoundException -> {
+            TransactionEncoderException.AnchorNotFoundException(BlockHeight.new(anchorHeight))
+        }
+
+        else -> {
+            fallback()
```
nesence-m: `runCatching` catches `Throwable`, and the `backend.propose*` calls dispatch via `withContext(SdkDispatchers.DATABASE_IO)` — so if the caller's scope is cancelled mid-call (user leaves the send screen), the `CancellationException` lands here and gets wrapped as a proposal failure: structured concurrency sees an error instead of cancellation and the app shows a spurious failure path. Pre-existing pattern, but this PR rewrote exactly these throw sites and introduced the helper, which makes it the natural place to fix:

```kotlin
is CancellationException -> throw this
```

(or check before wrapping in the `getOrElse` blocks).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/exception/Exceptions.kt
```diff
@@ -558,6 +558,18 @@ sealed class TransactionEncoderException(
...
+    /**
+     * The proposal could not be created because no anchor was computable at [anchorHeight], the
+     * height the proposal would anchor to. Scanning creates a checkpoint at every height a
+     * proposal can anchor to, so syncing further before retrying is expected to resolve this.
+     */
+    data class AnchorNotFoundException(
+        val anchorHeight: BlockHeight
+    ) : TransactionEncoderException(
+            "The proposal could not be created because no anchor was computable at height $anchorHeight. " +
+                "Syncing further before retrying is expected to resolve this."
```
nesence-m: Two small things on the new exception:

1. Every sibling proposal exception (`ProposalFromUriException`, `ProposalFromParametersException`, `ProposalShieldingException`) declares `val rootCause: Throwable` and forwards it to the parent's `cause`. This one drops the original `ProposalAnchorNotFoundException` — its Rust context string (which distinguishes the transfer path from the migration path) and stack survive only in the Twig log, invisible to crash reporters. One-line fix: carry it as `rootCause`/`cause`.
2. `$anchorHeight` interpolates the data-class `toString`, so the message reads "…at height BlockHeight(value=2434000)". Use `${anchorHeight.value}` (precedent at line 373: `${birthday?.value}`).

backend-lib/src/main/rust/lib.rs
```diff
@@ -2148,7 +2192,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_proposeTr
             lock_inputs,
             proposed_version,
         )
-        .map_err(|e| anyhow!("Error creating transaction proposal: {}", e))?;
+        .map_err(|e| map_proposal_error(env, "Error creating transaction proposal", e))?;
```
nesence-m: Not introduced here, but since this function is being touched: the `info_span!` at line 2166 in this function is copy-paste named `"RustBackend.proposeTransfer"`, identical to the real `proposeTransfer` span at line 2220 — traces of the two entry points are indistinguishable, including when diagnosing exactly the anchor failures this PR targets. Cheap to fix in passing: `"RustBackend.proposeTransferFromUri"`.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/TransactionEncoderImpl.kt
```diff
@@ -60,11 +65,14 @@ internal class TransactionEncoderImpl(
         }.onFailure {
             Twig.error(it) { "Caught exception while creating proposal from URI String." }
         }.getOrElse {
-            throw TransactionEncoderException.ProposalFromUriException(it)
+            throw it.asProposalException { TransactionEncoderException.ProposalFromUriException(it) }
```
nesence-m: Structural observation: this PR had to edit three identical `runCatching`/`getOrElse` chains in lockstep (here, line 86, line 117), `proposeShielding` (lines 128–139) uses the same chain without the mapping, and nothing enforces the pattern at compile time — the next proposal entry point silently regresses to the generic exception. All four propose calls already flow through a single choke point, `TypesafeBackendImpl`, which does no exception handling; mapping there (or factoring the debug/call/onFailure/getOrElse chain into one helper that owns `asProposalException`) would cover every current and future entry point instead of relying on per-method memory. Fine as a follow-up, especially given the entry-point question in the main comment.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2107,6 +2109,48 @@ fn zip317_helper<DbT>(
...
+/// `(String, long)` constructor signature is part of the JNI contract with the Kotlin class.
+const ANCHOR_NOT_FOUND_EXCEPTION_CLASS: &str =
+    "cash/z/ecc/android/sdk/internal/jni/ProposalAnchorNotFoundException";
+
+/// Converts a proposal failure into the `anyhow` error reported through the generic exception
+/// path, first raising a typed `ProposalAnchorNotFoundException` on the JVM when the failure is
+/// [`ProposalError::AnchorNotFound`], so that callers can react to it without matching message
+/// text. `unwrap_exc_or` leaves a pending exception in place, so the typed throw wins when it
+/// succeeds and the generic `RuntimeException` remains the fallback when it does not.
+fn map_proposal_error<DE, TE, SE, FE, CE, N>(
```
nesence-m, line 2125: Suggestion: this re-implements throw-with-log-fallback plumbing inline, while `utils/exception.rs` already centralizes JVM exception throwing and documents the pending-exception contract this function depends on — but gets no cross-reference, so someone simplifying its `exception_check` branch could silently break the typed throw. The next typed error (`InsufficientFunds`, per the TODO referencing #680) would copy-paste this 11-line closure with only the class constant and ctor signature changed. A `throw_object(env, class, ctor_sig, args)` helper next to `throw()` (since `env.throw_new` can't be used — the ctor is `(String, long)`) would keep the plumbing and its contract in one module. Non-blocking.

nesence-m (changes requested):
Thanks for the thorough writeup — the maintenance-branch targeting, the `proposeShielding` scope-out, and the `AnchorIntervalMismatch` CHANGELOG clarification are all right. But I don't think the typed exception can actually fire as wired, so requesting changes.

**Main issue (details on the CHANGELOG comment):** in `zcash_client_backend 0.24.0-rc.6` — the version `Cargo.lock` resolves, with the `[patch.crates-io]` block commented out — `ProposalError::AnchorNotFound` is constructed only inside the private `build_proposed_transaction` (`data_api/wallet.rs:1760/1805/1858`), which is reachable only from `create_proposed_transactions`. None of the three propose entry points this PR instruments can produce it, while `createProposedTransactions`' `map_err` (`lib.rs:2490`) still goes through the generic `anyhow` path and surfaces as `TransactionNotCreatedException`. So the typed mapping is dead code for the real failure path and the string-matching problem persists.

**Also outside the diff, so noting here:** the public contracts weren't extended — `SdkSynchronizer` (`@Throws` at lines 995/1006/1018), the `Synchronizer` interface KDoc (`Synchronizer.kt:318`), and the `TransactionEncoder` interface (lines 25/43/54) still declare only `ProposalFromUriException`/`ProposalFromParametersException`. Since `AnchorNotFoundException` is a *sibling* of those (not a subtype), app code that follows the documented contract with `catch (e: ProposalFromParametersException)` will let the new — deliberately recoverable — exception escape uncaught. The interface/impl `@Throws` lists also diverge now.

Remaining inline comments cover: a missing ProGuard keep for the JNI-constructed class (release-build breaker), a missing `exception_clear()` in the failed-throw fallback, `CancellationException` handling, cause-chain/message polish on the new exception, and a couple of structural suggestions.

The branch also has conflicts against `maint/v2.7.x` and needs a rebase.

nesence-m (changes requested):
^^

CHANGELOG.md
pacu: Confirmed — in both rc.6 and rc.7 the only construction sites are the three you cite, all inside `build_proposed_transaction`. One extension of the same analysis: `build_proposed_transaction` has a second caller, `create_pczt_from_proposal`, so the PCZT path can produce the error too.

Reworked accordingly: the mapping now lives on `createProposedTransactions` and `createPcztFromProposal`, the propose-path instrumentation is removed (no `@Throws` declares an exception that cannot occur), and this bullet now describes create-time behavior, including the call-site note for the sibling catch. The recovery wording also changed to "sync further, then create a new proposal" — the failed proposal anchors to the same height it failed at, so retrying create with it unchanged is not the fix.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/ProposalAnchorNotFoundException.kt
pacu: Added `@Keep` matching the `Jni*` precedent, with a comment noting the class is constructed only from native code and never referenced from bytecode.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/TransactionEncoderImpl.kt
pacu: With the mapping moved to the creation path, the propose-path `@Throws` additions are gone entirely — those contracts revert to what they were. The new exception is declared along the create chain instead: `TransactionEncoder.createProposedTransactions`, the impl, `SdkSynchronizer`, and the `Synchronizer` KDoc, with interface and impl lists kept identical.

The sibling concern transfers to `TransactionNotCreatedException` on that path and cannot be avoided by subtyping (it is a data class), so the CHANGELOG now states the required call-site edit explicitly.

backend-lib/src/main/rust/lib.rs
pacu: Fixed, and moved into a new `utils::exception::throw_object` helper that owns the construct-and-throw plumbing: on failure it logs and clears the pending exception, so the generic fallback engages in every failure mode instead of leaking the causing JVM error. Both `throw_object` and `unwrap_exc_or` now document the shared contract.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/TransactionEncoderImpl.kt
pacu: Done — the `getOrElse` blocks in the three propose methods, `proposeShielding`, `createProposedTransactions`, and `createPcztFromProposal` now rethrow `CancellationException` before wrapping. The remaining PCZT chains in this file are covered by the consolidation tracked in #2167.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/exception/Exceptions.kt
pacu: Both done: the exception now carries `rootCause` (the JNI exception, preserving the Rust context string and stack for crash reporters) and the message interpolates `${anchorHeight.value}`.

backend-lib/src/main/rust/lib.rs
pacu: Fixed in passing as suggested.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/transaction/TransactionEncoderImpl.kt
pacu: Agreed — filed as #2167. After the rework the mapping sites are the two creation paths rather than the propose methods, but the underlying point stands: the chains are still per-method and nothing enforces the pattern at compile time.

backend-lib/src/main/rust/lib.rs
pacu: Done in this PR after all: `utils::exception::throw_object` now owns the plumbing (a closure-building variant, since the constructor arguments need `env` to create the message string first), documents the shared pending-exception contract, and `unwrap_exc_or`'s `exception_check` branch gained a comment pointing back at it.

pacu:
Reworked and rebased onto current `maint/v2.7.x` (rc.7) — ready for another look. The updated description has the full summary; the short version: the typed mapping moved to `createProposedTransactions` and `createPcztFromProposal` (the second `build_proposed_transaction` caller), the propose-path instrumentation is removed, and the secondary items are addressed (`@Keep`, pending-exception clearing via a shared `throw_object` helper, contract declarations on the create chain, `CancellationException` rethrow, cause chain and message interpolation, span name). #2167 tracks the choke-point consolidation.

nesence-m (approved):
Re-verified every item from the 2026-08-11 review against the live code at 5e36515a (not the replies) — all blocking findings are addressed.

- Dead mapping on the propose path: `map_proposal_error` (`lib.rs:2125-2150`) is now called only from `createProposedTransactions` (`lib.rs:2485`) and `createPcztFromProposal` (`lib.rs:2541`), which are exactly the two callers of `build_proposed_transaction` in `zcash_client_backend 0.24.0` as `Cargo.lock` resolves it (no active `[patch.crates-io]`). The propose paths carry no instrumentation any more; CHANGELOG.md:9-26 describes create-time behaviour.
- `@Keep`: `ProposalAnchorNotFoundException.kt:15`, imported from `androidx.annotation` like the other JNI classes; `(String, Long)` matches `"(Ljava/lang/String;J)V"` at `lib.rs:2142`.
- Contracts: `TransactionEncoder.kt:96-100`, `TransactionEncoderImpl.kt:136-140` and `SdkSynchronizer.kt:1079-1083` carry identical `@Throws` lists; `Synchronizer.kt:397-404/422-425` and `Broadcaster.kt:26-33` document both the sibling exception and the PCZT `cause` semantics.
- `throw_object` (`utils/exception.rs:83-92`) clears only on failure and leaves a successful typed throw pending for `unwrap_exc_or:35`; no JNI call happens between the throw and the return.
- `CancellationException` is rethrown in all ten chains of `TransactionEncoderImpl.kt` (65, 81, 110, 132, 159, 187, 199, 211, 230, 249); the four added in 5e36515a change nothing else in those chains.
- `Exceptions.kt:568-575` carries `rootCause` as cause and interpolates `${anchorHeight.value}`; span name fixed at `lib.rs:2165`; #2167 tracks the structural consolidation.
- The five tests in `TransactionEncoderImplAnchorNotFoundTest` ran in CI (5/5, 0 failures) and do pin the mapping: removing `asAnchorNotFoundException()` at `TransactionEncoderImpl.kt:160` or `:188` fails them.

Non-blocking notes:
- `TransactionEncoderImplAnchorNotFoundTest.kt:84`: `check(cause is …)` reads better as `assertIs`; and no test asserts the height in `AnchorNotFoundException.message` or the cancellation rethrow on the other three PCZT wrappers.
- `TransactionEncoderImplAnchorNotFoundTest.kt:92-103` pins the pre-existing behaviour at `TransactionEncoderImpl.kt:188` where a non-anchor failure keeps `it.cause` and drops `it` itself from the chain — fine to leave, just noting it is not a contract worth preserving.
- The test KDoc / commit message say "faked" `TypesafeBackend`; it is a Mockito mock.
- `TransactionEncoderImpl.kt:19` imports `kotlin.coroutines.cancellation.CancellationException` while the rest of sdk-lib uses `kotlinx.coroutines.CancellationException` — same class, just inconsistent.
- App-consumer heads-up (not for this PR): the Zodl app currently detects the anchor case via `TransactionNotCreatedException` + message text; with the sibling `AnchorNotFoundException` that check stops matching, so the app needs the call-site edit the CHANGELOG describes when it picks this SDK up.

✅ No blocking issues — ready to approve

PR #2116, opened 2026-07-31
CI: check branch topology invariants

dannywillems:
Two advisory checks over the branch topology, both non-blocking and both
runnable locally.

## 1. Merge forward

The maintenance branches only work as a chain if every hop still merges: a fix
lands on the oldest line that needs it and is merged forward, oldest to newest,
with conflicts resolved explicitly at each hop. A conflict costs far less to see
while the change and its rationale are still in front of the author than weeks
later, when whoever runs the merge has to reconstruct the intent from a diff.

| Job | Checks |
|---|---|
| `maint/v2.7.x into maint/v2.8.x` | `check-merge-forward.sh maint/v2.7.x maint/v2.8.x "$BASE_REF"` |
| `maint/v2.8.x into main` | `check-merge-forward.sh maint/v2.8.x main "$BASE_REF"` |

One job per hop, so the failing hop is visible from the checks list without
opening the run. Both share one script with the source and target swapped.

**Pre-existing vs introduced.** When a hop conflicts, the script re-runs it on
the untouched branches. If the conflict is there without the change, it warns
and exits 0; only a conflict the change introduces exits non-zero. This matters
immediately: `maint/v2.8.x -> main` currently conflicts in `.gitignore` and
`AGENTS.md`. Without the split, every PR to `maint/v2.7.x` or `maint/v2.8.x`
would report red for a reason unrelated to it and the check would be tuned out.

## 2. Release merged back

A release is cut on a `release/` branch and tagged there. If that branch is only
cherry-picked back instead of merged, the tag stops being reachable from any
live branch: the commit that actually shipped is no longer part of the history
it shipped from, and every later "is this fix in the release" question has to be
answered by reading diffs instead of asking git.

`release branches merged back` walks every `release/` branch on origin and
checks it is an ancestor of the maintenance line it belongs to, and of
everything downstream to main.

- A branch counts as released only when a **tag points at its tip**, so
  in-flight release branches are skipped rather than reported.
- The maintenance line comes from the **tag**, not the branch name, because the
  two disagree: `release/v3.0.0` is tagged `v2.8.0-rc.1` and belongs to
  `maint/v2.8.x`. Deriving from the name would look for `maint/v3.0.x`.
- Only a release missing from its **own** line fails. Not having reached main
  yet is a warning, since the merge-forward jobs already track that, and failing
  on it would make the merge-back PR itself report red for a branch it was never
  going to update.

Current state on this PR:

```
:grey_question: release/v2.7.0-rc.4 has no tag at its tip; not a released branch, skipped.
:white_check_mark: release/v2.8.0-rc.2 (v2.8.0-rc.2) is in maint/v2.8.x and downstream.
:grey_question: release/v2.8.0-rc.3 has no tag at its tip; not a released branch, skipped.
:white_check_mark: release/v3.0.0 (v2.8.0-rc.1) is in maint/v2.8.x and downstream.
```

## Notes

- Neither trigger carries `paths-ignore`, unlike `pull-request.yml`. A merge
  conflict in `README.md` or `docs/` counts just as much as one in Kotlin, and
  the conflict we have right now is in `.gitignore` and `AGENTS.md`.
- The merge-forward hops are checked in isolation. A PR to `maint/v2.7.x` that
  merges fine into 2.8.x but would conflict with `main` after that merge is not
  caught, since the second job compares plain `origin/maint/v2.8.x` against
  `main`. Chaining the hops would catch it at the cost of the one-job-per-flow
  shape.
- The logic is in scripts rather than inline YAML so it can be run locally:
  ```
  ./.github/scripts/check-merge-forward.sh maint/v2.7.x maint/v2.8.x
  ./.github/scripts/check-release-merged-back.sh
  ```
  Both defer to a locally configured git identity, so no setup is needed.
- A new release line needs a new job in `merge-forward.yml`.
- No `CHANGELOG.md` entry: per `AGENTS.md` that file is for changes visible to
  consumers of the published library.

## Testing

`shellcheck` clean, `bash -n` clean, YAML parses, ASCII only, both scripts
committed executable.

- Merge forward: six scenarios against the real branches, including a synthetic
  conflicting commit to confirm the introduced case exits 1.
- Release merged back: exercised against a purpose-built synthetic repo covering
  a merged-back release, a released branch never merged back (exits 1), an
  untagged in-flight branch (skipped), the merge-back PR itself (warns on main,
  exits 0), and the state after main catches up.
- All three jobs then verified green in CI on this PR, with output matching the
  local runs.

### DIFF
--- a/.github/scripts/check-merge-forward.sh
+++ b/.github/scripts/check-merge-forward.sh
@@ -1,0 +1,96 @@
+#!/usr/bin/env bash
+#
+# Check that one hop of the maintenance chain still merges cleanly, optionally
+# with a pending change applied to either side.
+#
+# Usage: check-merge-forward.sh <source> <target> [pr-base] [pr-ref]
+#
+#   source    branch merged from, e.g. maint/v2.7.x
+#   target    branch merged into, e.g. maint/v2.8.x
+#   pr-base   branch a pending change targets; when it is <source> or <target>
+#             that side is replaced by <pr-ref>. Omit to check the branches as
+#             they stand.
+#   pr-ref    the pending change; defaults to HEAD (refs/pull/N/merge in CI).
+#
+# Branches resolve under refs/remotes/origin, so run `git fetch origin` first.
+# Exits 1 only when the change introduces the conflict, 0 otherwise.
+
+set -euo pipefail
+
+SOURCE="${1:?usage: check-merge-forward.sh <source> <target> [pr-base] [pr-ref]}"
+TARGET="${2:?usage: check-merge-forward.sh <source> <target> [pr-base] [pr-ref]}"
+PR_BASE="${3:-}"
+PR_REF="${4:-HEAD}"
+
+# commit-tree needs an identity. Fall back only when git has none of its own, so
+# a local run keeps the user's; git var fails exactly when commit-tree would.
+if ! git var GIT_COMMITTER_IDENT >/dev/null 2>&1; then
+  export GIT_AUTHOR_NAME='merge-forward check' GIT_AUTHOR_EMAIL='ci@localhost'
+  export GIT_COMMITTER_NAME='merge-forward check' GIT_COMMITTER_EMAIL='ci@localhost'
+fi
+
+CONFLICTS=''
+
+# Merge $2 into $1 in memory. Sets CONFLICTS and returns 1 on conflict.
+merge_hop() {
+  local into="$1" from="$2" out rc=0
+  out="$(git merge-tree --write-tree --name-only "$into" "$from")" || rc=$?
+  if [ "$rc" -ne 0 ]; then
+    # Line 1 is the tree, then the conflicting paths up to a blank line.
+    CONFLICTS="$(printf '%s\n' "$out" | tail -n +2 | sed -n '/^$/q;p')"
+    return 1
+  fi
+}
+
+say() {
+  printf '%s\n' "$*"
+  if [ -n "${GITHUB_STEP_SUMMARY:-}" ]; then printf '%s\n' "$*" >>"$GITHUB_STEP_SUMMARY"; fi
+}
+
+src="origin/${SOURCE}"
+dst="origin/${TARGET}"
+applied='no'
+if [ -n "$PR_BASE" ]; then
+  if [ "$PR_BASE" = "$SOURCE" ]; then src="$(git rev-parse "$PR_REF")"; applied='yes'; fi
+  if [ "$PR_BASE" = "$TARGET" ]; then dst="$(git rev-parse "$PR_REF")"; applied='yes'; fi
+fi
+
+say "## Merge-forward: \`${SOURCE}\` -> \`${TARGET}\`"
+say ''
+
+if merge_hop "$dst" "$src"; then
+  if [ "$applied" = 'yes' ]; then
+    say ':white_check_mark: Still merges cleanly with this change applied.'
+  elif [ -n "$PR_BASE" ]; then
+    say ':white_check_mark: Merges cleanly. This change touches neither side.'
+  else
+    say ':white_check_mark: Merges cleanly.'
+  fi
+  exit 0
+fi
+
+conflicts="$CONFLICTS"
+
+say 'Conflicting paths:'
+say ''
+say '```'
+say "$conflicts"
+say '```'
+say ''
+
+if [ "$applied" = 'no' ]; then
+  say ':warning: **Pre-existing.** This change touches neither side of the hop.'
+  exit 0
+fi
+
+# Re-run untouched, so a conflict the branches already carry is not blamed on
+# this change.
+if ! merge_hop "origin/${TARGET}" "origin/${SOURCE}"; then
+  say ':warning: **Pre-existing.** This hop already conflicts without the change.'
+  exit 0
+fi
+
+say ':x: **Introduced by this change.** The hop merges cleanly without it.'
+say ''
+say 'Advisory only. Resolve during the merge-forward, or reshape the change.'
+exit 1
--- a/.github/scripts/check-release-merged-back.sh
+++ b/.github/scripts/check-release-merged-back.sh
@@ -1,0 +1,130 @@
+#!/usr/bin/env bash
+#
+# Check that every tagged release/ branch has been merged back into its
+# maintenance line, and forward down to main.
+#
+# Usage: check-release-merged-back.sh [pr-base] [pr-ref]
+#
+#   pr-base   branch a pending change targets; that branch is then evaluated
+#             as it would be with the change applied. Omit to check the
+#             branches as they stand.
+#   pr-ref    the pending change; defaults to HEAD (refs/pull/N/merge in CI).
+#
+# A release/ branch counts as released only when a tag points at its tip, so
+# in-flight release branches are skipped. The maintenance line comes from the
+# TAG, not the branch name: release/v3.0.0 is tagged v2.8.0-rc.1 and belongs to
+# maint/v2.8.x.
+#
+# Branches resolve under refs/remotes/origin, so run `git fetch origin` first.
+# Exits 1 when a tagged release is missing from a branch it should be in.
+
+set -euo pipefail
+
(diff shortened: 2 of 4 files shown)


dannywillems:
See the example in the CI job for maint/v2.8.x into main.

dannywillems:
> See the example in the CI job for maint/v2.8.x into main.

```
Run git fetch --quiet --no-tags origin '+refs/heads/*:refs/remotes/origin/*'
## Merge-forward: `maint/v2.8.x` -> `main`

:white_check_mark: Merges cleanly. This change touches neither side.
```

No conflict anymore.

.github/workflows/merge-forward.yml
```diff
@@ -0,0 +1,71 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  merge_forward_v2_7_into_v2_8:
+    name: maint/v2.7.x into maint/v2.8.x
+    runs-on: ubuntu-latest
+    continue-on-error: true
```
nesence-m, line 33: Job-level `continue-on-error: true` reports the check run as **successful** in the PR checks list even when the script exits 1 — the failure is only visible inside the run view (long-standing runner behavior, actions/runner#2347). That defeats the stated goal of the one-job-per-hop layout: an "Introduced by this change" exit 1 surfaces as a green `maint/v2.7.x into maint/v2.8.x` check, and nobody opens green checks. The same applies to the job below and to `release-merged-back.yml`, where the `:x: not merged back` exit 1 would be equally invisible.

The usual way to get advisory-but-visible is to drop `continue-on-error` and simply not add these jobs to the branch-protection required checks: they then show a red X without blocking anything (merging and auto-merge only wait on required checks).

.github/scripts/check-merge-forward.sh
```diff
@@ -0,0 +1,96 @@
...
+  export GIT_AUTHOR_NAME='merge-forward check' GIT_AUTHOR_EMAIL='ci@localhost'
+  export GIT_COMMITTER_NAME='merge-forward check' GIT_COMMITTER_EMAIL='ci@localhost'
+fi
+
+CONFLICTS=''
+
+# Merge $2 into $1 in memory. Sets CONFLICTS and returns 1 on conflict.
+merge_hop() {
+  local into="$1" from="$2" out rc=0
+  out="$(git merge-tree --write-tree --name-only "$into" "$from")" || rc=$?
```
nesence-m, line 37: `git merge-tree` exits 1 both for conflicts and for an unresolvable ref ("not something we can merge"), so an error takes the conflict path with an empty path list. Verified against this script as-is (git 2.50):

```
$ ./check-merge-forward.sh maint/v2.9.x main
merge-tree: origin/maint/v2.9.x - not something we can merge
Conflicting paths:

```
```

:warning: **Pre-existing.** This change touches neither side of the hop.
$ echo $?
0
```

So a typo'd branch name in a future job — or a retired maint line whose job is left behind — turns that job into a permanent silent pass. A `git rev-parse --verify --quiet "$into^{commit}" "$from^{commit}"` guard before merging (or treating an empty conflict list as a hard error rather than a conflict) would make the broken-ref case fail loudly instead.

.github/scripts/check-merge-forward.sh
```diff
@@ -0,0 +1,96 @@
...
+say ''
+
+if [ "$applied" = 'no' ]; then
+  say ':warning: **Pre-existing.** This change touches neither side of the hop.'
+  exit 0
+fi
+
+# Re-run untouched, so a conflict the branches already carry is not blamed on
+# this change.
+if ! merge_hop "origin/${TARGET}" "origin/${SOURCE}"; then
```
nesence-m, line 88: This re-run makes the pre-existing detection all-or-nothing: if the untouched hop conflicts on *any* path, the script exits 0 — even when the applied merge additionally conflicts on paths that only conflict because of this change. The PR description itself documents the window where this bites: while `maint/v2.8.x -> main` sat conflicted on `.gitignore`/`AGENTS.md`, every PR to that line would have passed this check no matter how many new conflicts it introduced — precisely the period when the check is most needed.

Since `merge_hop` already collects the conflicting paths, comparing the two sets and exiting 1 when the applied run conflicts on paths the baseline run does not would keep the pre-existing split without the blind spot.

.github/scripts/check-merge-forward.sh
```diff
@@ -0,0 +1,96 @@
...
+set -euo pipefail
+
+SOURCE="${1:?usage: check-merge-forward.sh <source> <target> [pr-base] [pr-ref]}"
+TARGET="${2:?usage: check-merge-forward.sh <source> <target> [pr-base] [pr-ref]}"
+PR_BASE="${3:-}"
+PR_REF="${4:-HEAD}"
+
+# commit-tree needs an identity. Fall back only when git has none of its own, so
+# a local run keeps the user's; git var fails exactly when commit-tree would.
+if ! git var GIT_COMMITTER_IDENT >/dev/null 2>&1; then
```
nesence-m, line 27: Minor: nothing in this script creates a commit — `git merge-tree --write-tree` only writes trees, and it runs fine with no identity configured (checked with `env -i HOME=/nonexistent git merge-tree --write-tree ...`). The comment references `commit-tree`, which is never invoked, so this looks like a leftover from an earlier implementation. The block is harmless but can go, along with the "defer to a locally configured git identity" note in the PR description.

nesence-m (commented):
Solid PR overall: the invariants encoded here match CONTRIBUTING.md's model exactly (fixes merge forward along the maint chain, oldest to newest; a tagged `release/` branch must be merged back into its own line), the maint line is correctly derived from the tag rather than the branch name, the pre-existing/introduced split is the right idea, and the workflows follow this repo's conventions (SHA-pinned actions, step timeouts, scoped permissions). I ran `check-merge-forward.sh` locally against the real branches while reviewing.

Three findings are variations of "the check can pass without checking", which is the failure mode that matters most for an advisory CI check: job-level `continue-on-error` hides the one red signal these jobs can produce, a `merge-tree` error is indistinguishable from a conflict and ends in exit 0, and a pre-existing conflict on a hop masks any new conflicts a PR introduces on the same hop. Details inline; the rest is minor.

.github/scripts/check-release-merged-back.sh
```diff
@@ -0,0 +1,130 @@
...
+say ''
+
+failed=0
+checked=0
+
+while IFS= read -r rel; do
+  [ -n "$rel" ] || continue
+  ref="origin/${rel}"
+
+  tags="$(git tag --points-at "$ref" | tr '\n' ' ')"
```
nesence-m, line 61: Minor: any tag at the tip counts as "released", and the version parse then runs on whatever sorts first. A non-release tag (say `foo`) derives `maint/vfoo.foo.x`, falls into the "no longer exists" branch, and gets checked against main only — a spurious result either way. Restricting the lookup to release-shaped tags, e.g. `git tag --points-at "$ref" 'v[0-9]*'`, keeps stray tags from changing the outcome.

.github/workflows/release-merged-back.yml
```diff
@@ -0,0 +1,47 @@
...
+# Run locally with:
+#   ./.github/scripts/check-release-merged-back.sh
+
+name: Release Merged Back
+
+on:
+  pull_request:
+    branches:
+      - 'maint/v*'
+      - main
```
nesence-m, line 21: Minor: the state this job checks changes on pushes, not PRs — the tag lands after the review->release PR merges, and the post-release merge-back/forward steps in CONTRIBUTING.md can be direct merges. With a `pull_request`-only trigger, a release that gets cherry-picked instead of merged back only surfaces when the next unrelated PR to `maint/*` or `main` happens to open, and it flags that PR's author for someone else's omission. A `push: branches: [maint/v*, main]` trigger or a weekly `schedule:` would catch the drift when it actually happens (the script already tolerates an empty first argument, so the same invocation works); keeping the `pull_request` trigger is still worthwhile for the `resolve()` green-path on the merge-back PR itself.

nesence-m (approved):
utACK

PR #2123, opened 2026-08-03
Merge v2.7.x into v2.8.x

dannywillems:


nullcopy (approved):
utACK d6ed4d5c7b09bcdfce66e2f6dfafcdeb67a820f2

.github/workflows/merge-forward.yml
nuttycom, line 1: I'm not sure that this is actually a desirable check. My reasoning here is that if this reports red in CI, there isn't really a good mitigation for it, and we explicitly do NOT want people to (for example) merge `maint/v2.8.x` into `maint/v2.7.x` to try to resolve such complaints. Conflicts are *expected* merging forward, and I think it does not make sense to explicitly try to avoid them, so we shouldn't flag them as problematic.

.github/workflows/release-merged-back.yml
nuttycom, line 1: This by contrast is a good and important check to run.

PR #2130, opened 2026-08-04
Adopt released librustzcash rc.6/rc.7 crates, un-pin from temporary git rev

noop-sk:
## Summary

Un-pins `backend-lib` from the temporary git-pinned `librustzcash` revision (added to get early access to `zip318_kind` classification before it was released) and adopts the officially released crates: `zcash_pool_migration 0.1.0-rc.6`, `zcash_client_backend 0.24.0-rc.7`, `zcash_client_sqlite 0.22.0-rc.7`, `pczt 0.9.2`, `zcash_protocol 0.10.4`.

The released `zcash_pool_migration` crate diverged from the pinned rev by three release candidates' worth of API changes, so this is not a mechanical version bump — it required adapting `migration.rs`/`migration_engine.rs` to:

- New required `PoolMigrationRead`/`PoolMigrationWrite` trait methods (`check_step_satisfiability`, `mined_height`, `store_proved_transaction`) — delegated to `zcash_client_sqlite`'s own store implementation.
- A new `ReplanThreshold` policy parameter threaded through `commit_preparation`/`build_preparation_unsigned`/`MigrationState::from_parts` (9 call sites).
- A `DuenessTargets` two-tip type replacing bare `BlockHeight` in dueness queries.
- A changed `prove_transfer`/`prove_preparation` signature returning `ProveOutcome` — required restructuring `try_prove` so proving and persistence happen in strict sequence (a `Backend` cannot be constructed while the prover still holds the wallet mutably).
- `next_broadcastable`/`next_step` becoming crate-private, forcing adoption of the crate's own verified `advance_migration` driver in place of this codebase's hand-rolled polling (`guarded_next_step`). This also surfaces two new signals (`Replan`, `Reevaluate`) not previously representable in the JNI contract.
- Removing a local "late-dependency anchor" guard that predates this PR: `zcash_pool_migration`'s `prove_transfer` now redraws a transfer's anchor boundary inline at prove time when its funding dependency mined after the originally-drawn boundary (no re-signing needed, per ZIP 374) — this is exactly the fix requested from the core team in `spec/2026-07-30-engine-change-request-unprovable-boundary.md` months ago. The local workaround that excluded such transfers from proving (until ~30-day expiry) is now redundant and was actively conflicting with the new `advance_migration`-driven next-step path (which has no such guard), so it's removed and the driver surfaces reconciled.

Implemented and reviewed task-by-task (9 tasks + a final whole-branch review that caught the late-dependency guard conflict above, fixed and re-reviewed).

[...]

PR #2131, opened 2026-08-04
Merge librustzcash rc.7 into feature/ironwood-slipstream

nuttycom:
Subsumes #2130, which has been closed in favor of this PR.

Merges the rc.7 librustzcash line into the slipstream feature branch, plus a
rustfmt pass the merge's new CI job requires.

## The interesting part

`zcash_client_sqlite 0.22.0-rc.7` / `zcash_pool_migration 0.1.0-rc.6` have
**absorbed this branch's own two-tip design**. `DuenessTargets::new(scanned,
estimated)` is exactly broadcast@estimated / prove@scanned, same `height + 1`
target convention, same clamp — and `MigrationState::next_broadcastable` is now
private.

This PR **adapts rather than redesigns**. A merge forward should not rewrite the
branch it is merging into, so `guarded_next_step` survives, rebuilt over the
public `transaction_statuses(DuenessTargets)`. Adopting `advance_migration` —
which additionally does the in-flight sweep (mined promotion, foreign-spend and
anchor-invalidation marking) that the hand-rolled selector does not do at all —
is deliberately left as its own change. See "Follow-ups" below.

Worth noting the rebuilt selector is **strictly more conservative** than the
method it replaces: it also withholds rows carrying a broadcast-failure report,
rows marked unsatisfiable or stranded behind a dead dependency, and rows whose
expiry the tip estimate has probably passed. Each of those was previously a
submission the wallet could not see was doomed.

## Conflict resolutions worth a second look

**`Zip318Kind.CROSSING_PAYMENT(4)` dropped.** Nothing emitted it. The value comes
from `v_transactions.zip318_kind`, which librustzcash writes via
`Zip318Classification::to_code()`; `zcash_protocol 0.10.4` encodes only `0..=3`,
and `Zip318TxKind` has exactly `Preparation` and `Transfer`. Upstream refuses the
distinction on purpose — a canonical crossing paying a third party is built in
that shape *so that* it is indistinguishable. The enum KDoc referenced the
variant on **both** sides of the merge (the maint side already carried the
dangling link); it now states the property instead of citing a variant.

**Slipstream credentials on `static_analysis_rustfmt`.** The incoming job's Setup
step passes none, because the maint line has no private dependency. Here
`cargo fmt --all` resolves the workspace before formatting and this branch
patches `slipstream-core` to a private git rev, so the job needs them.
`check_proposal_proto` does not — its script is `curl` plus a lockfile parse.

**`ReplanThreshold`.** `MigrationState::from_parts` reconstructions preserve
`replan_threshold()`; fresh commits stamp `DEFAULT`, the value the rc.7 schema
migration backfills existing rows to.

**Three new `Blocker` variants** cross the JNI boundary, so codes 7/8/9 land with
`MigrationBlocker.EXPIRY_IMMINENT` / `AWAITING_REEVALUATION` / `UNSATISFIABLE` and
the decoder, in lockstep. `EXPIRY_IMMINENT` is protective and reverses itself once
the scan reaches the expiry — a UI must not render it as failure.

## Behaviour change to be aware of

`store_proved_transaction` means the wallet records a migration transaction at
**prove** time rather than at broadcast, so its inputs read as spent for the whole
prove-to-broadcast window. That is upstream's intent — it stops a foreground send
consuming a migration input — but it touches this branch's `PendingSubmitPlanStore`
work and the Keystone flow, and deserves eyes.

## Two pre-existing defects the merge surfaced

Both predate this PR; both are fixed here because they were red.

- `SlipstreamHostReadModelsTest` pinned a **16**-argument
  `SlipstreamTransactionRow` constructor. The row gained `zip318Kind` and the
  guard was never updated, so the JNI signature guard had been failing. Now
  matches `TX_ROW_CTOR`.
- `TransactionOverviewCursor` does not project `spent_note_count`,
  `pool_crossing_value` or `trust_status` — `host_read.rs`'s SQL and the JNI row
  have no slot for them. The values passed reproduce the shape that path had
  before those columns existed, so nothing regresses; it simply does not gain
  them.

## Verification

| Check | Result |
|---|---|
| `make test-rust` | pass — 56 passed, 0 failed, 13 ignored |
| `make lint-rust` (clippy `-D warnings`) | pass |
| `make check-format-rust` | pass (after the rustfmt commit) |
| `ci-local.sh unit` | pass — 21 test tasks |
| detekt | **fails, pre-existing — see below** |

## Update: `advance_migration` adopted (third commit)

`guarded_next_step` is gone; `nextStepNative` now drives the engine's own
`advance_migration` with `SETTLE_DEPTH = 10` blocks (~12 minutes at the
75-second target). Two steps that had no hand-rolled equivalent become reachable
and cross the JNI boundary as codes 5/6 with `MigrationAdvanceStep.Replan` and
`.Reevaluate`. The two-tip tests were ported onto `advance_migration` over an
in-memory store and still pass.

Note the drive loop is **not** a pure read: it sweeps in-flight transactions and
persists what it finds — promoting a `Proved` row whose broadcast record was lost
to a crash (which the old selector could never do), and marking transactions whose
inputs were spent elsewhere or whose anchor has left the chain.

## Follow-ups, not in this PR

1. **`state_machine_trace_tests` is still gated off.** Its `drain` harness drives
   the now-private `next_step` and rebuilds state through `from_parts` between
   steps; re-pointing it at `advance_migration` means threading the in-memory
   store through the whole harness. **That engine-trace coverage is not running.**
   The two-tip tests it shares helpers with were ported and do run.
2. **detekt**: 24 findings. Confirmed pre-existing — a worktree at the pre-merge
   tip (`2d1192a5`) produces the identical set, so the merge neither caused nor
   worsened it. Belongs to `feature/ironwood-slipstream`, not to this merge.
3. **Widen the slipstream projection** to carry the three `v_transactions`
   columns, which means extending the SQL, the row type, and the
   `SlipstreamTransactionRow` JNI constructor in lockstep.


nuttycom:
I recommend commit-by-commit review beginning with https://github.com/zcash/zcash-android-wallet-sdk/pull/2131/changes/a4e84157023590be48bb248cf01615575fcd3d1e (this is where changes relevant to `feature/ironwood-slipstream` start; the rest are changes from the maintenance branches.)

cc/ @noop-sk

backend-lib/Cargo.toml
```diff
@@ -52,15 +52,15 @@ pczt = { version = "0.9.0", features = ["prover", "orchard", "sapling", "signer"
 sapling = { package = "sapling-crypto", version = "0.7", default-features = false }
 transparent = { package = "zcash_transparent", version = "0.10", default-features = false }
 zcash_address = "0.13"
-zcash_client_backend = { version = "0.24.0-rc.6", features = [
+zcash_client_backend = { version = "0.24.0-rc.7", features = [
```
noop-sk, line 55: **Stale version requirements nearby**

`zcash_client_backend`/`zcash_client_sqlite` got bumped to rc.7 here, but `zcash_pool_migration` (still `0.1.0-rc.6`'s predecessor `0.1.0-rc.3`), `pczt` (`0.9.0`), and `zcash_protocol` (`0.10`) a few lines above/below weren't touched. `Cargo.lock` does resolve them correctly (`0.1.0-rc.6`/`0.9.2`/`0.10.4` — pre-release semver matching pulls the newer rc through), so the build is correct, but the declared requirement strings no longer document the actual floor this branch depends on. Worth bumping them to match for clarity.

noop-sk (commented):
**Stale "Follow-ups" note**

The PR description's "Follow-ups, not in this PR" section still lists `state_machine_trace_tests` as "still gated off" / "not running". That looks superseded by the later `test(migration): restore the state-machine trace tests over a mock store` commit (`f4bc1c5bf`) — I checked the head commit directly and all four golden-trace tests (`live_plan_happy_trace_completes_with_late_but_in_margin_prep`, `unprovable_boundary_heals_via_prove_time_redraw`, `expiry_surfaces_rebuild_for_the_stuck_transfer`, `awaiting_signature_blocks_until_apply_signature`) are present with no `#[ignore]`. Worth updating the description so reviewers don't think that engine-trace coverage is still missing.

noop-sk (approved):

PR #2133, opened 2026-08-04
feat(slipstream): deferred preparation with DbReady read phase

nesence-m:
## Why

Today `SlipstreamSynchronizer.new()` performs the entire heavy construction before returning — `restoreAnchor` (a blocking network fetch; on a cold-Tor restore that's bootstrap + up to ~45 s of bounded circuit attempts) → `initDataDb` → `createAccount` → `engine.open` → `engine.start`. `WalletCoordinator` calls it inside a `callbackFlow`, so for that whole window the app's `synchronizer` StateFlow is null (unexplained shimmer, the MOB-1634 failure class), and any construction throw kills the flow and leaves it null **forever** with zero UI feedback. It is also why a Tor-enabled restore shows a long dead window while a Tor-off restore doesn't.

## Design

Mirrors the iOS SDK's `prepare()` / `Initializer.slipstreamAnchorSource` split **behind the unchanged Android API**: `new()` returns in milliseconds with `status = INITIALIZING`; the tail runs in an internal prepare job guarded by a four-state machine (`Preparing → DbReady → Ready`, terminal `Failed`/`Closed`, CAS-only transitions so `close()` always wins races).

> **Design-review flag (@nuttycom):** this deliberately deviates from the upstream `Synchronizer.new` contract of returning a fully-ready instance. Engine/DB operations now await readiness internally; failures surface through `onSetupErrorHandler` instead of a throw from `new()`. The iOS `prepare()` precedent is the model; flagging explicitly since the "new() returns ready" posture was previously load-bearing.

## The changes

1. **Deferred prepare + readiness gates** (`SlipstreamSynchronizer.kt`): every engine/DB-touching operation awaits `Ready` (`awaitReady()` throwing / `awaitPrepared()` boolean); lifecycle blocks (`onForeground`/`onBackground`/`pause`/`resume`) queue behind the gate inside `launchGuarded`; `syncBurst` probes synchronously and returns `UNAVAILABLE` while preparing; `close()` cancels-and-joins the prepare job before any engine teardown step.
2. **DbReady read phase**: the tail is reordered iOS-style — `initDataDb` **before** the network anchor — and read-only surfaces (`getAccounts`, address getters, transactions/memos/outputs/recipients, transparent balance, `debugQuery`, plus the `accountsFlow`/`allTransactions` cold flows) relax to `awaitDbReady()`. An existing wallet serves accounts and transactions as soon as migrations finish, before any network round-trip; the rewind targets are coerced to `>= startBirthday` to keep `WalletCoordinator.rescanBlockchain()` safe against the provisional birthday.
3. **Failure surfacing**: a failed prepare forces engine status `DISCONNECTED` and latches the error into `onSetupErrorHandler` (with late-attach replay, since the host attaches handlers only after the first emission). Cold flows complete instead of killing the host's eager collectors.
4. **Typed provisioning**: the treestate plumbing carries `TreeState` end-to-end and unwraps `.encoded` exactly once, at the `createAccount` JNI boundary.
5. **Truthful-from-DbReady balances**: `engine.walletBalances` is seeded from `TypesafeBackend.getWalletSummary()` at DbReady (best-effort; the first engine tick overwrites with its stale-tip masking) — closes the zero-balance flash on relaunch.
6. **No transient "no accounts" state**: `accountsFlow` suppresses empty emissions while the tail's `createAccount` is still pending, so a restore's anchor window renders as *loading*, not *absent*; `accountsVersion` (StateFlow counter, replacing the replaying SharedFlow) drives re-reads from `createAccount`/`importAccountByUfvk`/`deleteAccount`.

## App-side counterpart

Landed on `zashi-android` `feature/ironwood-slipstream` as [Handle deferred slipstream preparation on home messages and migration worker (`95e93f4d3`)](https://github.com/Electric-Coin-Company/zashi-android/commit/95e93f4d3): setup-error banner precedence, Restoring-banner narration during INITIALIZING, and MigrationWorker INITIALIZING-awareness.

## Testing

- `SlipstreamSynchronizerLifecycleTest`: **44/44 green** — covers cheap construction, gate ordering (initDataDb → createAccount → open → start → startPolling), the close/prepare CAS race, prepare-failure latch + DISCONNECTED, flow survival on Failed-after-DbReady, syncBurst probing, foreground queuing, migration-pause interaction, rewind coercion, the balance seed, and the empty-accounts suppression. The suppression and seed tests are negative-controlled (removing the fix line fails exactly those tests).
- Known pre-existing failure left untouched: `SlipstreamHostReadModelsTest.transaction_row_constructor_matches_rust_jni_signature` (zip318 reflection pin vs `2d1192a5`).
- Whole-stack compile: `zashi-android` `assembleZcashmainnetStoreDebug` and `assembleZcashmainnetInternalRelease` (R8) both green against this branch via the included build.

## Manual verification status

- RIM build (`zcashmainnetInternalRelease`, v3.8.1) passed the merged-lib integrity checks (RustBackend/SlipstreamNative/MigrationRustBackend families, 16 KB alignment, single native lib per ABI) and is deployed to a physical Pixel 9.
- Pending device matrix: existing-wallet relaunch (instant accounts/balances, no 0.000 flash), restore with Tor on/off (loading → account+banner together), airplane-mode restore (setup-error banner), close-during-prepare (wallet switch mid-restore).

## Review follow-ups

- [`496c1d9c`](https://github.com/zcash/zcash-android-wallet-sdk/commit/496c1d9c): `onSetupErrorHandler` gets a `@Volatile` backing field (store→load pairing with the atomic latch); `newBlocking`/`erase` restored — `erase` strengthened to the upstream wait-for-shutdown posture via `InstanceGuard.withKeyInactive`; and `WalletCoordinator.resetSdk()`/`deleteSdkDataFlow()` now also wipe Slipstream's `data.sqlite3` when `isSlipstreamEnabled` (pre-existing reset gap).

### DIFF
--- a/sdk-lib/src/main/java/com/zodl/slipstream/SlipstreamSynchronizer.kt
+++ b/sdk-lib/src/main/java/com/zodl/slipstream/SlipstreamSynchronizer.kt
@@ -625,9 +625,13 @@ class SlipstreamSynchronizer internal constructor(
             Twig.error(t) { "Slipstream synchronizer preparation failed" }
             val landedFailed =
                 prepareState.updateAndGet {
-                    if (it is PrepareState.Preparing || it is PrepareState.DbReady) PrepareState.Failed(
-                        t
-                    ) else it
+                    if (it is PrepareState.Preparing || it is PrepareState.DbReady) {
+                        PrepareState.Failed(
+                            t
+                        )
+                    } else {
+                        it
+                    }
                 } is PrepareState.Failed
             if (landedFailed) {
                 /*
@@ -727,7 +731,9 @@ class SlipstreamSynchronizer internal constructor(
                     recoverUntil = provisioning.recoverUntil
                 )
             }.getOrElse { throw InitializeException.CreateAccountException(it) }
-            /* Releases [accountsFlow], which suppressed the empty reads of the whole DbReady window. */
+            /*
+             * Releases [accountsFlow], which suppressed the empty reads of the whole DbReady window.
+             */
             accountsVersion.update { it + 1 }
         }
 
@@ -873,7 +879,7 @@ class SlipstreamSynchronizer internal constructor(
             throw InitializeException.ImportAccountCheckpointsNotReadyException(
                 IllegalStateException(
                     "Slipstream engine did not quiesce before account import; refusing to import " +
-                            "to avoid racing the still-unwinding sync pass."
+                        "to avoid racing the still-unwinding sync pass."
                 )
             )
         }
@@ -1126,21 +1132,13 @@ class SlipstreamSynchronizer internal constructor(
                     }
                 }
             val sdkBranchId =
-                runCatchingCancellable {
-                    "%x".format(
-                        Locale.ROOT,
-                        backend.getBranchIdForHeight(currentChainTip.value)
-                    )
-                }
+                runCatchingCancellable { "%x".format(Locale.ROOT, backend.getBranchIdForHeight(currentChainTip.value)) }
                     .getOrElse { return ServerValidation.InValid(it) }
             return if (remoteInfo.consensusBranchId.equals(sdkBranchId, ignoreCase = true)) {
                 ServerValidation.Valid
             } else {
                 ServerValidation.InValid(
-                    CompactBlockProcessorException.MismatchedConsensusBranch(
-                        sdkBranchId,
-                        remoteInfo.consensusBranchId
-                    )
+                    CompactBlockProcessorException.MismatchedConsensusBranch(sdkBranchId, remoteInfo.consensusBranchId)
                 )
             }
         } finally {
@@ -1257,8 +1255,10 @@ class SlipstreamSynchronizer internal constructor(
     override fun getTransactionsByMemoSubstring(query: String): Flow<List<TransactionId>> =
         flow {
             emit(
-                transactionReader.getTransactionsByMemoSubstring(query)
-                    .map { TransactionId.new(it) })
+                transactionReader
+                    .getTransactionsByMemoSubstring(query)
+                    .map { TransactionId.new(it) }
+            )
         }.onStart { awaitDbReady() }
 
     override fun getRecipients(transactionOverview: TransactionOverview): Flow<TransactionRecipient> {
@@ -1290,7 +1290,8 @@ class SlipstreamSynchronizer internal constructor(
 
     override suspend fun getTransactionOutputs(): Map<TransactionId, List<TransactionOutput>> {
         awaitDbReady()
-        return transactionReader.getAllTransactionOutputs()
+        return transactionReader
+            .getAllTransactionOutputs()
             .mapKeys { (txId, _) -> TransactionId.new(txId) }
     }
 
@@ -1310,10 +1311,13 @@ class SlipstreamSynchronizer internal constructor(
     }
 
     override suspend fun checkSingleUseTransparentAddress(accountUuid: AccountUuid): Boolean =
-        when (val response = walletClient.checkSingleUseTransparentAddress(
-            accountUuid.value,
-            ServiceMode.UniqueTor
-        )) {
+        when (
+            val response =
+                walletClient.checkSingleUseTransparentAddress(
+                    accountUuid.value,
+                    ServiceMode.UniqueTor
+                )
+        ) {
             is Response.Success -> response.result != null
             is Response.Failure -> false
         }
@@ -1322,8 +1326,10 @@ class SlipstreamSynchronizer internal constructor(
         accountUuid: AccountUuid,
         address: String
     ): Boolean =
-        when (val response =
-            walletClient.fetchUtxosByAddress(accountUuid.value, address, ServiceMode.UniqueTor)) {
+        when (
+            val response =
+                walletClient.fetchUtxosByAddress(accountUuid.value, address, ServiceMode.UniqueTor)
+        ) {
             is Response.Success -> response.result != null
             is Response.Failure -> false
         }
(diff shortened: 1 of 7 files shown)


sdk-lib/src/main/java/com/zodl/slipstream/SlipstreamSynchronizer.kt
```diff
@@ -1360,58 +1934,15 @@ class SlipstreamSynchronizer internal constructor(
...
-            return withContext(Dispatchers.IO) {
-                val dbFile =
-                    DataDbPath.dataDbFile(appContext.applicationContext.getNoBackupFilesDirSuspend(), alias, network)
-                val walFile = File("${dbFile.path}-wal")
-                val shmFile = File("${dbFile.path}-shm")
-                listOf(dbFile, walFile, shmFile).map { !it.exists() || it.delete() }.all { it }
-            }
-        }
-
         private const val ENGINE_TOR_SUBDIR = "slipstream_tor"
```
noop-sk, line 2032: **Silently dropped `newBlocking`/`erase`**

This PR deletes `Companion.newBlocking()` (the Java-interop blocking constructor) and `Companion.erase()` (deletes Slipstream's `data.sqlite3`/-wal/-shm) entirely, but the unmodified "Companion factory surface" KDoc a few lines above `companion object {` still claims `new`/`newBlocking`/`erase` exist here as "the parallel surface" because `Synchronizer`'s own companion members can't be ridden by an external implementation. With `erase` gone, there is no public API left to wipe a Slipstream-backed wallet's database — `Synchronizer.erase()` unconditionally calls `SdkSynchronizer.erase(...)`, which targets a different DB file. Was this deletion intentional? If so, the parity KDoc and any "erase wallet"/Java call sites need to follow.

sdk-lib/src/main/java/com/zodl/slipstream/SlipstreamSynchronizer.kt
```diff
@@ -300,11 +455,23 @@ class SlipstreamSynchronizer internal constructor(
...
+     * attached after it already happened is invoked immediately on assignment - the host attaches
+     * its handlers only once the synchronizer has been emitted, which is routinely later than the
+     * failure. Assigning `null` (the host's teardown) never replays anything. A handler that is
+     * detached and re-attached is therefore invoked once per attachment; that repeat is benign,
+     * since the host maps it onto the same latched error state.
+     *
+     * Server-mismatch and runtime sync problems still surface through [onProcessorErrorHandler].
      */
     override var onSetupErrorHandler: ((Throwable?) -> Boolean)? = null
+        set(value) {
```
noop-sk, line 477: **Unsynchronized handler crosses threads**

`onSetupErrorHandler` is a plain `var` (no `@Volatile`/atomic backing) that is now actually invoked cross-thread: `runPrepare`'s catch block calls `onSetupErrorHandler?.invoke(t)` from `Dispatchers.IO`, while the host assigns the property from its own thread once the synchronizer is emitted. If a preparation failure lands at nearly the same moment as attachment (e.g. the airplane-mode restore case in this PR's own manual test matrix, where `restoreAnchor` can fail almost immediately), the IO-side read could observe a stale `null` while the setter's `latchedSetupError.get()` replay also finds nothing latched yet - silently dropping the setup-error banner. `latchedSetupError` is already an `AtomicReference` for exactly this kind of cross-thread hazard; the handler field itself isn't.

sdk-lib/src/main/java/com/zodl/slipstream/SlipstreamSynchronizer.kt
nesence-m: Unintentional — restored in 496c1d9c, and the comment also surfaced that the erase story needed more than a revert:

- `newBlocking` is back verbatim (and is now strictly cheaper than before, since `new()` returns in milliseconds), so the companion "factory surface" KDoc is accurate again.
- `erase` is back **strengthened**: the old bare `check(!InstanceGuard.isActive(key))` raced `close()`, which marks the key shutting down synchronously while the engine teardown that releases the database files runs asynchronously — deleting the `-wal`/`-shm` under a live engine is exactly the one-SQLite-instance corruption class. It now runs through `InstanceGuard.withKeyInactive`, mirroring `SdkSynchronizer.erase`'s `mutex.withLock { waitForShutdown(); check(); delete }` posture (mutex held across the deletion, in-flight shutdown of the same key awaited first). Covered by a new `InstanceGuardTest` case.
- The gap you flagged ("no public API left to wipe a Slipstream-backed wallet's database") was actually pre-existing in spirit: nothing ever *called* `SlipstreamSynchronizer.erase`, so `WalletCoordinator.resetSdk()`/`deleteSdkDataFlow()` left `data.sqlite3` behind even before this PR. Both paths now also wipe the Slipstream database when `isSlipstreamEnabled`, with an active-instance refusal logged and reported as a failed delete rather than killing the `walletScope` coroutine.

sdk-lib/src/main/java/com/zodl/slipstream/SlipstreamSynchronizer.kt
nesence-m: Fixed in 496c1d9c: the backing field is now `@Volatile`. Each side stores into its own cell before loading the other's — the IO side does `latchedSetupError.set(t)` (atomic store) then reads the handler field (volatile load); the attach side writes the field (volatile store) then reads the latch (atomic load) — so with both cells having volatile semantics, at least one of the two loads must observe the other side's store. The scenario you describe (stale `null` handler AND empty latch replay) becomes impossible; the worst case is a double invocation, which the KDoc already declared benign since the host maps it onto the same latched error state. Extended the KDoc with a paragraph naming this pairing.

noop-sk (approved):

PR #2136, opened 2026-08-06
fix(migration): scope the migration DB mutex — pure-read bypass and broadcast phase split

nesence-m:
## Motivation (MOB-1623)

Every `OrchardMigrationSdkImpl` operation ran inside `logged`, i.e. under the process-wide `MIGRATION_DB_ACCESS_MUTEX` for its entire duration. That put two unrelated things behind one lock: multi-second network sends, and reads that need no mutual exclusion at all.

The consequence the app sees is the migration Progress screen's first paint. `getMigrationTransferStates` queues behind whatever currently holds the mutex — a proving pass, a 60 s Tor broadcast, another instance's retry sleeps — for up to ~15 s of blank screen in practice.

This PR narrows what the mutex covers. It does not change the mutex's purpose, and no mutating operation loses protection.

### (a) A pure-read lane

`loggedRetryLoop` is factored out of `logged`, and a new `loggedRead` gives the same logging and bounded retry **without** the mutex.

Read consistency does not need it: calls into `migrationBackend` are already serialized onto the single `zc-io` thread, and Rust opens its own connection per call with a `busy_timeout` — the same arrangement under which migration reads and the sync engine's writes already coexist. What the bypass buys is liveness; the retry sleeps that remain on this path now block nobody.

Only verified-pure single reads move over:

- `getMigrationTransferStates`
- `getMigrationSummary`
- `syncWakeupSchedule`
- `estimateMigrationRunCount`
- `isNoteSplitNeeded`
- `migrationDustThresholdZatoshi`
- the four Keystone QR codec calls (which touch no database at all)

Deliberately **not** moved — and now documented on `logged` as the rule for future edits — is everything reaching Rust's `read_reconciled` (`get_migration` → `mark_mined` → `replace_migration`): `getMigrationState`, `getMigrationProgress`, `hasOverdueTransfers`, `hasInvalidTransfers`, `nextStep`, and `isSyncBlockedNow`'s overdue read. Those read like queries but are read-modify-write; running two concurrently can lose an update, including a just-recorded broadcast result.

### (b) A flat three-phase broadcast split

All three broadcasting methods (`executeNextPendingTransfer`, `submitNoteSplit`, `storeSignedNoteSplitPczt`) now run three **flat** phases sharing a `PreparedBroadcast` carrier:

1. **prepare** — under the DB mutex;
2. **send** — under a new `BROADCAST_MUTEX` only;
3. **record** — back under the DB mutex, inside `NonCancellable`.

The phases are flat rather than nested, so the two mutexes are never held at once and there is no lock ordering to get wrong. The lock acquisition sits *inside* the `NonCancellable` boundary in phase 3, so a cancellation arriving while queueing for the mutex cannot drop an already-broadcast result.

This also fixes a latent bug. A `"database is locked"` from `recordTransferResult` used to make `logged`'s retry re-run the whole block — **broadcasting the same transaction a second time**. That was survivable only because `classifyNonGrpcFailure` maps a duplicate rejection back to success, and terminally fatal to the pre-signed plan on any node whose rejection text lacks those markers. A failed record now retries the record alone.

Two guards move with the split, because the whole-method mutex used to provide them implicitly:

- The `MIGRATION_BROADCAST_IN_FLIGHT_UNTIL` mark now also gates re-entry in phase 1. With the send outside the DB mutex, a second caller (`MigrationProgressVM`'s foreground pass, `MigrationSendingVM`, `MigrationWorker` — all real concurrent entrants) could otherwise re-extract and re-broadcast a transfer already on the wire. The recovery guard runs first, so an active mark means a send is genuinely in progress.
- That mark is refreshed to a full window on **acquiring** `BROADCAST_MUTEX`, not only when it is first set. A queued caller can have waited most of a `BROADCAST_TIMEOUT` behind another broadcast, and a mark expiring mid-send would un-gate the sync engine with bytes still on the wire.

The mined-height probe (F2) runs on the pure-read lane rather than the mutex — it follows a send that may well have succeeded, and losing it to transient contention would strand that outcome unrecorded — and stays outside the `NonCancellable` boundary with the send, so an outer timeout can still kill a hung attempt.

### (c) Mutation retries stay under the lock — deliberately

Releasing the lock between attempts would let another queued operation change the very state the failed attempt read (which transfer is next due, the plan cache), reintroducing exactly the check-then-act interleaving the mutex exists to prevent. Sleeping under the lock is the price; now that sends and pure reads are out from under it, no UI-visible read waits behind those sleeps.

### Test seam

An optional `broadcaster` constructor parameter (the `MigrationTransactionBroadcaster` fun interface) allows the network phase to be substituted in unit tests — the real path builds a `WalletClientFactory` and, with Tor, a Tor runtime, neither of which runs in a plain JVM test. It is null on every production caller; `MigrationSdk.new()` is unchanged.

## Verification

- `OrchardMigrationSdkImplTest` — 22/22 green, including four new concurrency tests, one per property the split is meant to guarantee (each fails against the pre-split code):
  - `getMigrationTransferStates_completes_while_a_broadcast_is_parked` — the MOB-1623 symptom itself;
  - `executeNextPendingTransfer_retries_the_record_without_re_broadcasting` — the latent double-broadcast bug;
  - `a_second_caller_does_not_re_broadcast_the_transfer_already_in_flight` — the in-flight re-entry guard;
  - `a_transfer_send_and_a_note_split_send_never_overlap` — `BROADCAST_MUTEX` serializes the network phase across the different broadcasting methods.
- Full `:sdk-lib` unit-test suite green.
- Mutation-checked: the serialization test was confirmed to fail when the split is reverted, so it is testing the property and not passing vacuously.

## Reviewer note — design sign-off wanted

This changes a documented invariant of `logged()`: it was previously true that *every* migration DB operation runs under `MIGRATION_DB_ACCESS_MUTEX`. After this PR, that holds for mutations and for read-modify-write reads, but no longer for the pure-read list above. The correctness argument rests on (1) `zc-io` single-thread serialization plus Rust's per-call connection with `busy_timeout`, and (2) the claim that each listed call is genuinely a single pure read. Both are worth a second pair of eyes before this lands.

### DIFF
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/OrchardMigrationSdkImpl.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/OrchardMigrationSdkImpl.kt
@@ -119,6 +119,14 @@ internal class OrchardMigrationSdkImpl(
     // without a substitutable PreferenceHolder. Production callers are unaffected: MigrationSdk.new()
     // still passes a real EncryptedPreferenceProvider, which is a PreferenceHolder.
     private val preferenceProviderHolder: PreferenceHolder,
+    /**
+     * Test seam for the network phase of a broadcast, and nothing else: the real path builds a
+     * [WalletClientFactory] (plus, with Tor, a real Tor runtime), neither of which can run in a plain
+     * JVM unit test, so the broadcast-serialization and phase-ordering tests substitute a gateable
+     * implementation. Left `null` by every production caller — `MigrationSdk.new()` included — in
+     * which case [broadcast] takes its ordinary [WalletClientFactory] path.
+     */
+    private val broadcaster: MigrationTransactionBroadcaster? = null,
 ) : OrchardMigrationSdk {
     /**
      * [NetworkPrivacyOptions.useTor] is a per-migration setting, independent of the app's global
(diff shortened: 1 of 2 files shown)


noop-sk:
FYI — opened #2137, which merges this PR's mutex scoping together with the tag=4/observedTip broadcast-rejection rewiring that landed in parallel on `feature/ironwood-slipstream-rc7-adoption`. Kept as a separate PR rather than pushing onto this branch so it doesn't disrupt the design-sign-off review here. Full test suite (this PR's 4 new concurrency tests + the tag=4 regression tests) green together in #2137.

PR #2137, opened 2026-08-06
merge(migration): reconcile #2136 mutex-scoping with tag=4/observedTip rewiring

noop-sk:
## Summary

Merges #2136 (mutex scoping — pure-read bypass + three-phase broadcast
split) with the migration engine work done in parallel on
`feature/ironwood-slipstream-rc7-adoption` (the tag=4/AwaitingReevaluation
rejection-handling rewrite, the observedTip-on-rejection fetch, and the
`wasOverdue` simplification that drops a redundant `hasOverdueTransfers()`
read).

Both refactors touched `executeNextPendingTransfer`'s broadcast path in
overlapping but structurally different ways, so this required weaving them
together rather than a mechanical merge — see the merge commit message for
the full breakdown of what was kept from each side and why.

## Verification

- `:sdk-lib:testDebugUnitTest` — 237/237 green, including all 4 of #2136's
  new concurrency tests (`getMigrationTransferStates_completes_while_a_broadcast_is_parked`,
  `executeNextPendingTransfer_retries_the_record_without_re_broadcasting`,
  `a_second_caller_does_not_re_broadcast_the_transfer_already_in_flight`,
  `a_transfer_send_and_a_note_split_send_never_overlap`) running alongside
  the existing tag=4/wasOverdue regression tests.
- Full composite build (app `feature/ironwood-slipstream` + this branch)
  compiles clean with `-Werror` and the app's own test suite (615 tests) is
  green against this SDK branch.

## Relationship to #2136

This is a separate PR rather than an addendum to #2136, so #2136's own
review isn't disrupted. #2136's design-sign-off question (the `logged()`
mutex invariant change) still applies unchanged here — this PR only adds
where the tag=4 rewiring sits relative to the new phase split, it doesn't
touch the mutex-scoping reasoning itself.


nesence-m (dismissed):
utACK

PR #2138, opened 2026-08-07
Merge `feature/ironwood-slipstream` changes back to `main`

nuttycom:
This PR brings `main` up to date with the `feature/ironwood-slipstream` work. Work for the Ironwood/Slipstream release continues to be based on and merge to `feature/ironwood-slipstream`; the purpose of this is to integrate all outstanding work so that we can begin working on coinholder polling on `main` without duplicating or conflicting with ongoing work on the feature branch.

This is the Android counterpart of zcash/zcash-swift-wallet-sdk#1954, and takes the same resolution.

## Conflicts

254 hunks across 26 files, 211 of them in the voting module.

**Voting takes `main`'s side wholesale.** The branches had diverged on it completely. `main` re-enabled shielded voting on the published `zcash_voting` crate (#2075); this feature branch parked the module behind `cfg(zcash_voting)` with the dependency commented out, so its voting edits were made against sources that branch never compiled. `main`'s is the only version that builds. The feature branch said as much itself:

> Voting is re-enabled separately, against `main`, by #2075 -- do not re-enable it here.

**One Ironwood semantic is carried across:** voting weight now reads `get_unspent_ironwood_notes_at_historical_height` rather than the Orchard equivalent, because votes are cast with Ironwood-pool notes and an Orchard read measures the wrong balance. This is the same change the Swift merge made.

**`zcash_voting` moves to the published 2.0.0-rc.3** and its `[patch.crates-io]` git entry is dropped. That patch existed to avoid the duplicate Zcash crate graph of #2056, and the published release pins the Ironwood family directly.

**Re-enabling voting does put the share-reveal circuit back into the shipped library**, via `voting-circuits`, which this feature branch had deliberately removed. Worth a reviewer's explicit agreement. Note that naming orchard's `unstable-voting-circuits` feature is not what controls this: `zcash_voting` enables it transitively either way, so the only way to keep the circuit out is to not ship voting. `cargo tree -i voting-circuits` shows the single path.

Everything else takes the newer or additive side: librustzcash rc.7 and the feature branch's dependency set (`slipstream-core`, `zcash_pool_migration`, Keystone `ur`/`ur-registry`, `shardtree`, `tokio`); `actions/checkout` v7.0.1 from `main` **combined with** the feature branch's `persist-credentials: false` hardening; and the union of both `CHANGELOG.md` Unreleased sections, less one entry the feature branch had duplicated into Unreleased that `main` already records under the published 2.6.0 heading.

## Testing

`./scripts/ci-local.sh quick` passes: detekt, ktlint, 68 Rust unit tests (0 failed), Android unit tests. `cargo check --all-targets` is clean at all four points of the feature matrix (default, `slipstream`, `android-test-fixtures`, both).

Two defects surfaced during verification and are fixed here. `serde`/`serde_json` were dropped from `backend-lib/Cargo.toml` and only the `slipstream`-enabled build noticed. `JniVotingModelsTest.kt` auto-merged *without* conflicting and silently mixed both sides' test code, giving duplicate overloads and the feature branch's `rawAddress` naming; it is reset to `main`'s version with the rest of the module.

**Not run:** `./scripts/ci-local.sh full` (androidTest, 15-30m). A reviewer may want that before merge.

---
*Opened with [Claude Code](https://claude.com/claude-code) assistance.*


backend-lib/src/main/rust/lib.rs
```diff
@@ -106,10 +108,22 @@ use crate::utils::{
 
 mod eip681;
 mod migration;
+mod migration_engine;
+mod migration_keystone;
+mod migration_plan_cache;
+mod migration_send_max;
 mod tor;
 mod utils;
 mod voting;
```
nullcopy, line 117: This has a feature gate on the slipstream branch. Do we want to pull the feature gate in, so that voting can be optionally built?

nullcopy (approved):
I diffed against `feature/ironwood-slipstream` with the expectation that `main` should become effectively the slipstream branch + voting updates

utACK 4818067828658521e0dcfbbe4ccafb49da6e4f33

PR #2156, opened 2026-08-10
MOB-1669: snapshot spendable-note read on first use in migration Backend

noop-sk:
## Summary
`Backend::resolve_wallet_note` (via `spendable_orchard`) re-ran `select_unspent_notes` — a real SQLite query enumerating and sorting every spendable Orchard note for the account — from scratch on every single call, with no caching. It's called once per wallet-sourced spend input while a preparation tree is built, so a note split needing N wallet inputs paid that full query cost N times over.

Confirmed live on a physical Pixel: `createUnsignedNoteSplitPczt` dropped from ~7.1-7.3s to ~845ms (an ~8.4x improvement) — this was the dominant cause of a QA report ("Android took 3.5 minutes to generate 3 batched QRs" during Keystone migration signing, vs. iOS "instant" on the same flow).

Fix mirrors `zcash_pool_migration`'s own upstream `WalletMigration` adapter, which already snapshots spendable notes on first read within one adapter instance's lifetime — the exact registry `0.1.0-rc.7` version this crate already depends on, no dependency bump needed. See `spec/2026-08-10-migration-engine-walletmigration-adapter-plan.md` for why wholesale-adopting `WalletMigration` itself (to literally mirror iOS's architecture) is a separate, deferred, higher-risk change — iOS pins an unmerged upstream `librustzcash` revision with a breaking constructor signature change that doesn't fit ~39 of our 51 `Backend::new` call sites (Keystone/hardware-wallet and read-only flows, which hold no local spending key).

## Test plan
- [x] `cargo test --features slipstream` — 62/62 passing
- [x] `cargo clippy --tests --all-features -- -W clippy::all` — clean (no new findings)
- [x] `cargo fmt --all -- --check` — clean
- [x] Live-verified on a physical Pixel (mainnet internal debug): `createUnsignedNoteSplitPczt` timing confirmed dropping from ~7.1-7.3s to ~845ms
- [ ] Dedicated cache-behavior regression test (proving the snapshot holds across a mid-lifetime wallet mutation) is a tracked fast-follow — no existing lightweight `InputSource`/wallet test double exists in this codebase to build one on cheaply; every existing test touching `Backend` requires a real `MIGRATION_TEST_WALLET_DB`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Migration Keystone batch signing no longer stalls for seconds when building the first note-split
+  PCZT of a run: spendable-note selection is now cached for the lifetime of one migration call
+  instead of being re-queried from the wallet database on every note the plan spends (MOB-1669).
+
 ## [3.0.1] - 2026-08-08
 
 ### Changed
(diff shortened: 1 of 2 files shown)


nesence-m (approved):
utACK

PR #2157, opened 2026-08-10
MOB-1678: Fix multi-bundle CHP delegation, adopt tx1_effects

noop-sk:
## Summary
- `delegate-vote` submission now sends the versioned `tx1_effects` blob the vote-chain server requires (zcash_voting 2.0.0-rc.4) instead of the legacy `sighash` field the server no longer accepts.
- `build_governance_pczt_for_bundle` no longer advances the round's phase to `HotkeyGenerated`/`DelegationConstructed` during construction. That write was a zodl-SDK addition the crate itself never does, and it made constructing bundle 1+ of a multi-bundle round fail with "refusing to regress round phase" once bundle 0 had already been proved — leaving bundle 1's `alpha` NULL and crashing `build_and_prove_delegation`.
- Exposes the crate's own per-bundle `DelegationPhase` (`delegationPhasesNative`) and the sanctioned setup-reset recovery path (`resetVotingSessionStateNative`) so callers can make per-bundle progress decisions instead of relying on the coarse round-level phase.

Companion app PR: zodl-inc/zodl-android#2406

## Test plan
- [x] Verified end-to-end on-device via the companion zodl-android branch: multi-bundle CHP voting round completes delegation construction, proving, and submission for bundle 1+ after bundle 0 is already proved.
- [x] `VotingRustBackendTest` / `TypesafeVotingBackendImplTest` androidTest suites compile and pass.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -361,8 +361,16 @@ jobs:
           rust-cache: true
           slipstream-app-id: ${{ secrets.SLIPSTREAM_APP_ID }}
           slipstream-app-private-key: ${{ secrets.SLIPSTREAM_APP_PRIVATE_KEY }}
+      # 60 rather than the usual 30: this branch re-enables shielded voting, which
+      # turns on orchard's `unstable-voting-circuits` and pulls the halo2 voting
+      # circuits into the native build. `make test-unit` builds the crate for all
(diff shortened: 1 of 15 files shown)


dannywillems:
This will need to be merged back in `main`.

dannywillems (approved):
utACK 0d27626a087561f50b059b835040ade879f5f0fd.

PR #2158, opened 2026-08-10
Merge maint/v3.0.x into main

nesence-m:
Forward merge of the `maint/v3.0.x` maintenance line into `main`, per the maintenance-branch model in [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md): fixes land on the oldest supported `maint/` line and are merged forward, never backward.

## What this forwards

`maint/v3.0.x` already contains everything from `release/v3.0.1` and `release/v3.0.2`, and those in turn contain the `maint/v2.8.x` (`release/v2.8.0-rc.3`) and `maint/v2.7.x` (`release/v2.7.0-rc.4`) lines. All of those release→maint and maint→maint merges were verified already contained (`git merge-base --is-ancestor`) before starting, so `maint/v3.0.x` → `main` was the only merge outstanding. Before this PR, `maint/v3.0.x` was 87 commits ahead of `main` and `main` 27 ahead of `maint`.

## :rotating_light: Merge with a merge commit — do not squash

The template checklist below says to squash before review. **That does not apply to this PR.** Squashing would destroy the shared commit identity the maintenance-branch model depends on: `main` would no longer contain `maint/v3.0.x`'s commits by sha, so every future forward merge would re-conflict on the same content. Use "Create a merge commit".

## Conflicts and how each was resolved

**`backend-lib/Cargo.toml` — slipstream engine dependency and `[patch.crates-io]`** (real + silent semantic conflict)
Took **maint's side**: the published crates.io crate, `slipstream-core = { package = "zodl-slipstream", version = "0.1" }`, and dropped `main`'s `[patch.crates-io] slipstream-core` git pin entirely (the section retains only main's comment on why `zcash_voting 2.0.0-rc.3` needs no patch — no active entries remain).

Why not keep main's git pin: the pinned rev (`ab89ada4`) no longer exists on `zodl-inc/slipstream` — the repo's history was rewritten to the published `zodl-slipstream 0.1.x` line, so main's pin is unfetchable on any fresh machine (`upload-pack: not our ref`; this is exactly what failed all six CI jobs on the first version of this PR — and would fail the same way on `main` itself). `zodl-slipstream 0.1.1` is published against the same librustzcash generation as this graph, so no duplicate family crates enter (verified below). Note for `main` follow-up: `main` still carries the dead pin and cannot build from scratch until this merge lands.

**`backend-lib/Cargo.lock`** (real conflict)
Reset to `main`'s, then reconciled by cargo after `Cargo.toml` was settled. The resulting diff is only:
- `zcash_client_sqlite` 0.22.0-rc.7 → rc.8 and `zcash_pool_migration` 0.1.0-rc.6 → rc.7 (the bumps maint brings),
- `pczt` 0.9.2 → 0.9.3, required by the above,
- the new transitive tree of maint's test-only `zcash_client_sqlite` dev-dependency (`ambassador`, `proptest`, `wagyu-zcash-parameters*`, …),
- the engine swap: `slipstream-core 0.6.4` (git) → `zodl-slipstream 0.1.1` (crates.io registry).

No unrelated crate-version changes.

**`CHANGELOG.md`** — auto-merged cleanly into the intended union: `[Unreleased]` on top, then `[3.0.1]`, `[3.0.0]`, `[2.8.0-rc.3]`, … Verified line-by-line that no entry from either side was dropped.

## Auto-merged, deliberately left as maint's

`LIBRARY_VERSION` 2.8.0-rc.2 → 3.0.1 (`gradle.properties`) and the crate `version` → 3.0.0. These were not conflicts — `main` had not touched either since the merge base, so its 2.8.0-rc.2 was simply stale. Forcing main's value back would regress a released version number below the content the branch now carries.

## Verification

- `cargo tree -d` — no duplicated `zcash_*`, `orchard`, `sapling-crypto`, `rusqlite` or `libsqlite3-sys`, both with default features and with `--features slipstream`.
- `cargo tree -i` — exactly one node each: `zcash_protocol v0.10.4`, `orchard v0.15.4`, `sapling-crypto v0.7.0`, `rusqlite v0.37.0`, `libsqlite3-sys v0.35.0`, `zcash_client_sqlite v0.22.0-rc.8`, `zcash_client_backend v0.24.0-rc.7`, `zcash_pool_migration v0.1.0-rc.7`, `zodl-slipstream v0.1.1` (crates.io).
- `cargo check --features slipstream` — passes. The one `dead_code` warning (`clear_invalidation`) is inherited from `maint/v3.0.x`, which removed the production caller along with the raw-SQL self-heal path; it is not introduced here.
- `./gradlew :sdk-lib:compileDebugKotlin :backend-lib:compileDebugKotlin` — BUILD SUCCESSFUL.

> **Note**

# Author

# Reviewer

---


nuttycom (commented):
I resolved this same merge independently (same two parents, `70179133` + `41e9e833`) before seeing this PR, so I diffed the two resolutions. They differ in exactly three files. One looks like a real problem, one is a nit, and on one your resolution is better than mine.

## `CHANGELOG.md` — `## [3.0.0]` absorbed `main`'s entire `Unreleased` block

`main` had no `3.0.0` section, so the incoming `## [3.0.0] - 2026-08-08` heading landed on top of `main`'s `Unreleased` content rather than alongside it. Git can't see this as a conflict — one side inserted a heading, the other edited the text below it — so it merges silently.

The effect:

- `## [3.0.0]` is **244 lines** here; the `v3.0.0` tag's own section is **129**.
- All **35** of `main`'s unreleased entries now sit under the `3.0.0` heading. Only **18** of them shipped in a 3.0 release.
- The other **17** are the shielded-voting work and the #2056 double-crate-graph fix. No 3.0 release contains them — `maint/v3.0.x` has `zcash_voting` commented out entirely.
- `Unreleased` is left with 1 entry.

Reproduce with:

```bash
git show v3.0.0:CHANGELOG.md | awk '/^## \[3\.0\.0\]/{f=1} /^## \[2\.8\.0-rc\.3\]/{f=0} f' | wc -l   # 129
git show HEAD:CHANGELOG.md   | awk '/^## \[3\.0\.0\]/{f=1} /^## \[2\.8\.0-rc\.3\]/{f=0} f' | wc -l   # 244
```

What makes this worth fixing rather than tidying later: several of the 17 are breaking API changes announced as such — "**Shielded voting works again, on a source-incompatible API**", "hotkeys are no longer derived from the wallet seed", `deriveHotkeyRawAddress` keeping its signature while changing its meaning. Publishing those under a GA heading tells integrators they can depend on API that has not shipped, and the changelog is the only interface most of them read.

Suggested resolution: restore `3.0.0` and `3.0.1` to their tagged text verbatim, and keep `Unreleased` as the 17 entries that shipped in neither, plus the one `maint/v3.0.x` had pending (MOB-1669). Worth checking both directions afterwards — that no shipped entry is still listed as unreleased, and that no entry was dropped. While you're in there, `main`'s `Unreleased` had also accumulated duplicate `### Added` and `### Fixed` headings from earlier merges, which are easy to consolidate in the same pass.

This is the fifth time in this merge chain that a release branch cut, then merged back after the trunk edited the same entries in place, has silently reattributed work — `2.7.0-rc.4`, `2.8.0-rc.2`, `2.8.0-rc.3`, `3.0.0`'s crate version, and now this. A cheap mechanical guard is to diff every `## [x.y.z]` section against `git show <tag>:CHANGELOG.md` and flag any difference containing deletions; additions alone are usually legitimate backfill.

## `backend-lib/Cargo.lock` — reintroduces a deduplicated `multimap` (nit)

This keeps both `multimap 0.8.3` and `multimap 0.10.1` from `main`, with references disambiguated. `maint/v3.0.x` had unified on `0.8.3`. `prost-build 0.14.4` requires `multimap >=0.8, <=0.10`, so `0.8.3` satisfies it, and a plain `cargo metadata` does not re-add `0.10.1` to either lock — both are valid and stable, so this is only a redundant duplicate rather than a defect. Taking the 3.0 line's deduplication seems marginally preferable.

## `backend-lib/Cargo.toml` — your version is better than mine

The comment above `slipstream-core` on `maint/v3.0.x` is stale: it still says the crate "is git-pinned" and that `[patch.crates-io]` is "the single source of truth", which stopped being true when the line moved to the published `zodl-slipstream` crate. You rewrote it; I merged the stale text through. I'll take yours.

One small addition worth considering: leaving a line in the now-empty-looking `[patch.crates-io]` saying *why* `slipstream-core` is absent, next to the existing note about `zcash_voting`, so the next person doesn't wonder whether the patch was dropped by accident.

---
Reviewed with Claude Code assistance.

backend-lib/Cargo.toml
```diff
@@ -134,11 +134,10 @@ zcash_voting = { version = "2.0.0-rc.3" }
...
 # from under it (the dual-instance SIGBUS/disk-I/O hazard; see
-# docs/slipstream/android/worklog/ARCHIVE-phases-0-9.md for the full writeup). `slipstream-core`
-# is git-pinned (no machine-local path); backend-lib's `[patch.crates-io]` below is the single
-# source of truth governing the librustzcash rev for `core` and the whole graph. Keep this rev
-# in step with that patch block.
-slipstream-core = { version = "0.6", optional = true }
+# docs/slipstream/android/worklog/ARCHIVE-phases-0-9.md for the full writeup). The engine is
+# the published `zodl-slipstream` crate from crates.io, built against the same librustzcash
+# generation as the rest of this graph — verify with `cargo tree -d` after any bump.
+slipstream-core = { package = "zodl-slipstream", version = "0.1", optional = true }
```
noop-sk, line 140: **Unpinned crates.io dependency**

`slipstream-core` moves from a git-pinned rev (immutable, single source of truth for the whole graph) to `zodl-slipstream` on crates.io with an unpinned `"0.1"` range. Trust now rests on whoever holds crates.io publish rights for that name, and a future `cargo update` can silently pull a new `0.1.x` without an explicit rev-level diff to review — previously any change to this dependency showed up as a rev bump in this file.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/OrchardMigrationSdkImpl.kt
```diff
@@ -725,32 +1032,79 @@ internal class OrchardMigrationSdkImpl(
...
+                            MigrationAdvanceStep.Rebuild(id)
+                        }
+
+                        STEP_COMPLETE -> {
+                            MigrationAdvanceStep.Complete
+                        }
+
+                        STEP_REPLAN -> {
+                            // Every consumer of nextStep() gets a superseded migration automatically,
+                            // not just the app worker's specific Replan handler.
```
noop-sk, line 1055: **`nextStep()` read now writes**

Hitting `STEP_REPLAN` here unconditionally calls `markMigrationSuperseded`, a DB write, on every call to `nextStep()` — not just the worker loop's dedicated Replan handler. Any read-only/status-polling caller in the app that happens to invoke `nextStep()` now also mutates migration state as a side effect. Is this intended for all current app call sites, or should the write be scoped to the driver loop specifically?

nuttycom:
I've pushed `7d5db35c` to this branch with the CHANGELOG fix from my review above, rather than leaving you to reconstruct it — shout if you'd rather I hadn't and I'll drop it.

It restores `## [3.0.0]` and `## [3.0.1]` to byte-identical with their tags, and rebuilds `Unreleased` as the 17 entries that shipped in neither release plus the one `maint/v3.0.x` had pending (MOB-1669). Checked both directions afterwards: no shipped entry is still listed as unreleased, and no entry from either parent was dropped. It also consolidates the duplicate `### Added` and `### Fixed` headings `main` had accumulated from earlier merges.

Nothing else on the branch is touched — the `multimap` and `slipstream-core` points from the review are untouched, and on the latter yours is the better version, so it stands.

---
Pushed with Claude Code assistance.

backend-lib/Cargo.toml
nesence-m: True in general, but this shape is the model working as intended rather than a loosening: the git pin was scaffolding for the *unreleased* engine, and its referent no longer exists — `ab89ada4` was erased when zodl-inc/slipstream was rewritten to the published `zodl-slipstream 0.1.x` line, so the "immutable" rev is unfetchable on any fresh machine (it is what failed all six CI jobs on the first version of this branch). Post-publication, the engine is consumed exactly like the rest of the released librustzcash family: a caret range in the manifest, the exact version (`0.1.1`) pinned by `Cargo.lock`, so `cargo update` cannot move it silently — any bump shows up as a reviewable lockfile diff, same as a rev bump did. crates.io versions are immutable once published (unlike git branches, and evidently unlike this git history), and supply-chain trust in the `zodl-slipstream` publisher is the same trust the released `maint/v3.0.x` line (v3.0.0/v3.0.1) already shipped with this exact dependency shape.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/OrchardMigrationSdkImpl.kt
nesence-m: Good catch on the behavior, but it is not a merge artifact: this file is carried byte-identical from `maint/v3.0.x` — the write was introduced there by `df5fdc95` ("feat(migration): wire mark_superseded into nextStep()'s Replan handling") and already shipped in v3.0.0 and v3.0.1; `main`'s side had no `markMigrationSuperseded` call for the merge to interact with. So the question of whether every `nextStep()` caller should trigger the supersede write is about the released behavior itself and worth raising against that change (or as a follow-up issue), rather than something this forward-merge can alter without diverging from what already shipped.

nesence-m:
Thanks for the CHANGELOG fix — please keep `7d5db35c`, that's exactly the right resolution and better than reconstructing it here. Verified it in both directions on my side as well: the `3.0.0`/`3.0.1` sections are byte-identical to their tags again and the awk section-count check now gives 129/129.

The two remaining points are addressed in `3b89d20d`:
- **multimap**: lockfile unified on `0.8.3` (`cargo update -p multimap@0.10.1 --precise 0.8.3`), matching maint's deduplication; `cargo tree -d` shows a single node and `cargo check --features slipstream` still passes.
- **\[patch.crates-io\] note**: added a line next to the `zcash_voting` note stating `slipstream-core`'s absence is deliberate since the engine moved to the published `zodl-slipstream 0.1.x` line.

Replied to the two inline threads: the unpinned-range concern (lockfile pins 0.1.1 exactly; the old git rev no longer exists upstream), and the `nextStep()` write, which is carried byte-identical from maint (`df5fdc95`, shipped in v3.0.0/v3.0.1) rather than introduced by this merge.

noop-sk (approved):

PR #2161, opened 2026-08-10
Use zcash_voting 2.0.0-rc.5 from crates.io

dannywillems:
Bumps the shielded-voting dependency from `2.0.0-rc.4` to `2.0.0-rc.5` and
removes the `[patch.crates-io]` git pin.

## Changes

- `zcash_voting` `2.0.0-rc.4` -> `2.0.0-rc.5`.
- Dropped the `[patch.crates-io]` entry pointing at the valargroup git
  remote. rc.5 is a published crates.io release, so the git revision the
  patch existed to supply is no longer needed. `zcash_voting`,
  `vote-commitment-tree` and `vote-commitment-tree-client` now all resolve
  from the registry in `Cargo.lock`.
- Updated the stale re-enable comment in `Cargo.toml`, which told the
  reader to uncomment a patch entry that no longer exists.
- Second commit: added the missing `tx1_effects` field to the test-only
  `DelegationSubmissionData` fixture. That constructor was not updated when
  the production JNI path adopted `tx1_effects`, so the test target failed
  to build with E0063 before this change.

## Notes

rc.5 changes no API this tree depends on. `DelegationSubmissionData`
already carried `tx1_effects` in rc.4, so the only build break was the
pre-existing test fixture above.

## Testing

- `RUSTFLAGS="--cfg zcash_voting" cargo check --all-targets` is clean.
- `RUSTFLAGS="--cfg zcash_voting" cargo test voting::` passes, 8/8
  (including the three delegation tests).

### DIFF
--- a/backend-lib/Cargo.lock
+++ b/backend-lib/Cargo.lock
@@ -1206,7 +1206,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d308ebe4b10924331bd079044b418da7b227d724d3e2408567a47ad7c3da2a0"
 dependencies = [
  "derive-deftly-macros",
- "heck 0.4.1",
+ "heck 0.5.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


backend-lib/Cargo.lock
pacu, line 1: are these downgrades intentional?

pacu (dismissed):
utACK modulo determining that the version downgrades are intentional

pacu (approved):

PR #2165, opened 2026-08-11
MOB-1678: CHP voting fixes and VotingSdk API extraction

noop-sk:
## Summary
- Multi-bundle CHP voting delegation fix: stop faking a round-level phase during PCZT construction (which hard-rejected later bundles as a "phase regression"); every construct/prove/submit decision now uses the SDK's own per-bundle `DelegationPhase`. Adopts `tx1_effects` in delegate-vote submissions (the vote-chain server rejects the legacy `sighash` field with a 400). Originally shipped as #2157 into `maint/v3.0.x`; this PR carries the same fix forward now that this branch is reparented onto `main`.
- Adds a public `VotingSdk`/`VotingDbSession` API (mirrors the existing `MigrationSdk` pattern) so the app never needs to see `VotingRustBackend`/JNI types directly — closes a layering violation the app-side `feature-voting` extraction (companion PR below) depends on.
- Reparented from `maint` onto `main` via merge (not rebase): `main` had picked up its own, independent CHP rewrite in July (#2056) that turned out to be stale/protocol-incompatible (still sends `sighash`, not `tx1_effects`) — this branch's voting implementation was kept, `main`'s discarded.
- Fixes an unconditional nullifier-size validation in `TypesafeVotingBackendImpl.recordShareDelegation` that crashed on a wallet's first share delegation (`nullifier must be 32 bytes, got 0`) — the JNI layer's own documented contract treats an empty nullifier as the normal case for callers that don't have one yet; only the Kotlin typesafe wrapper contradicted it.

## Test plan
- [x] `:sdk-lib` / `:backend-lib` compile and unit tests green
- [x] New instrumented regression tests for the nullifier fix pass on-device (21/21, `TypesafeVotingBackendImplTest`)
- [x] Full multi-bundle CHP voting round tested live end-to-end via the companion app PR, on-device

Companion app PR: zodl-inc/zodl-android#2406


backend-lib/Cargo.toml
```diff
@@ -121,11 +121,19 @@ xz2 = { version = "0.1", features = ["static"] }
...
-# the shipped library, via the `voting-circuits` crate it depends on. That is the cost of
-# shielded voting on this branch, accepted deliberately; the orchard note above records why
-# turning off orchard's `unstable-voting-circuits` feature does not avoid it.
-zcash_voting = { version = "2.0.0-rc.3" }
+# zcash_voting 2.0.0-rc.3 (main's version) still sends the legacy `sighash` field on
+# delegate-vote submissions; the vote-chain server now rejects that with "invalid message
+# field: tx1 effects must be 821 bytes, got 0" (confirmed 2026-08-10). rc.4 replaces it with
+# the required `tx1_effects` field — that server-protocol fix is why this branch is pinned
+# ahead of what's published to crates.io, via the git rev in `[patch.crates-io]` below.
+zcash_voting = { version = "2.0.0-rc.4" }
```
dannywillems: Nit: `maint/v3.0.x` uses `2.0.0-rc.5`

backend-lib/Cargo.lock
```diff
@@ -6204,7 +6306,7 @@ dependencies = [
  "static_assertions",
  "subtle",
  "sync_wrapper",
- "thiserror 2.0.19",
+ "thiserror 2.0.20",
```
dannywillems, line 6309: Are these updates expected?
EDIT: it seems coming from the bump up of `0.6.0-rc.2`.

backend-lib/src/main/rust/voting/share_tracking.rs
```diff
@@ -16,13 +16,59 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_com
     blind: JByteArray<'local>,
 ) -> jbyteArray {
     let res = catch_unwind(&mut env, |env| {
+        // zcash_voting 1.0.0 (merged-library patch) moved this from
```
dannywillems: Nit: this can be removed in a follow-up

backend-lib/src/main/rust/voting/recovery.rs
```diff
@@ -157,32 +179,55 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_rec
...
     bundle_index: jint,
     proposal_id: jint,
-    vc_tree_position: jlong,
+) -> jobject {
+    let res = catch_unwind(&mut env, |env| {
+        let db = db_from_handle(db_handle)?;
+        let _access_lock = db.access_lock()?;
+        let bundle_index = jint_to_u32(bundle_index, "bundle_index")?;
+        let record = optional_recovery_lookup(
+            db.get_commitment_bundle(
```
nesence-m, line 217: **Major — throws instead of returning null for a committed-but-unconfirmed vote.**

The base code phase-gated this state (`VotePhase::Confirmed`) to null, with a comment documenting the crate's deliberate failure; that gate was removed and not replaced. `optional_recovery_lookup` maps only "query returned no rows" to null — but between `vote::commit` (which persists `commitment_bundle_json` with `vc_tree_position` NULL) and `recordVcPosition` (only possible after the cast-vote tx confirms), the crate fails with "commitment bundle is stored without vc_tree_position; refusing to assume position 0", which surfaces as a `RuntimeException`.

That window is the *normal* state during confirmation — so the natural poll-`getCommitmentBundle`-until-non-null loop on the new nullable public API crashes on every vote, and any app restart before confirmation hits the same. Fix: map this specific crate error to null too (or restore the phase gate before the lookup).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/voting/VotingModels.kt
```diff
@@ -0,0 +1,452 @@
...
+    val voteAuthorityNoteNew: ByteArray,
+    val voteCommitment: ByteArray,
+    val proof: ByteArray,
+    val encShares: List<VotingEncryptedShare>,
+    val anchorHeight: Long,
+    val sharesHash: ByteArray,
+    val shareComms: List<ByteArray>,
+    val rVpk: ByteArray,
+    val voteAuthSig: ByteArray,
+    val sharePayloads: List<VotingSharePayload>
```
nesence-m, line 213: **Major — these payloads are built against a placeholder `vc_tree_position` of 0, but the class KDoc says "ready to broadcast and share out".**

`vote.rs` builds them before the real position exists (hardcoded 0 at ~line 81); they must be rebuilt after `recordVcPosition`. An app that follows the KDoc and hands these to helper servers binds every share to the wrong tree position — rejected (or invalid) once the real position is recorded, and the vote silently loses its delegated shares. There's not even a direct rebuild path from this type: `VotingSdk.buildSharePayloads` takes a `VotingCommitmentResult`, not a `VotingCommitResult`.

Fix: KDoc must state these are provisional and point at the rebuild path — or better, don't expose pre-position payloads on the public result at all.

backend-lib/src/main/rust/voting/util.rs
```diff
@@ -71,28 +76,54 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_ext
...
-/// are fixed by `zcash_voting` and are not caller-selectable, which is why the
-/// address is derived by the crate rather than here.
+/// The ZIP-32 account index is intentionally fixed at 0 and is not exposed as a parameter.
+/// zcash_voting's own hotkey derivation (`VotingHotkey::from_stored_secret`) hardcodes the
+/// same account 0 / address 0 pair for its stored-secret hotkeys. Letting callers pass an
+/// arbitrary account here would allow delegation to be built against a hotkey the vote
+/// construction path cannot subsequently sign for, so the API surface hides the constraint
+/// instead of leaving it as a runtime trap.
 #[unsafe(no_mangle)]
 pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_deriveHotkeyRawAddressNative<
```
nesence-m, line 99: **Major — accepts 32–63-byte seeds that no hotkey-reconstruction path can ever use.**

`java_secret_bytes_at_least(PROTOCOL_FIELD_BYTES=32)` here, but `VotingHotkey::from_stored_secret` — used by `generateHotkey`, `buildVoteCommitment`, `buildAndProveDelegation` — rejects anything that isn't exactly `HOTKEY_STORED_SECRET_BYTES=64`. The base signature required exactly 64, making the mismatch unrepresentable.

Failure mode is nasty: a caller passes 32 bytes, gets back a perfectly valid-looking Orchard address, and delegates round voting weight to it — then every later operation rejects the same bytes, so no hotkey can ever be reconstructed for that address and the delegated weight is irrecoverably forfeited. Restore the exact-64 requirement.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/voting/VotingModels.kt
```diff
@@ -0,0 +1,452 @@
+package cash.z.ecc.android.sdk.model.voting
+
+/** Public mirror of the internal `VotingNoteScope` — which shielded-address scope a note came from. */
+enum class VotingNoteScope {
+    EXTERNAL,
+    INTERNAL
+}
+
+/** A shielded note eligible for voting weight, as read from wallet state. */
+data class VotingNoteInfo(
```
nesence-m, line 13: **Major — secrets-in-toString regression.** The base `JniNoteInfo`/`JniSharePayload`/`JniShareDelegationRecord` overrode `toString` to "(redacted)", with covering tests; both the overrides and the tests were deleted, and neither the JNI classes nor these new public `VotingNoteInfo`/`VotingSharePayload`/`VotingShareDelegationRecord` redact — while siblings in this same file (`VotingCommitmentResult`, `VotingCommitResult`, `VotingHotkey`) do, which reads as an oversight rather than a policy change.

Any log line / exception message / crash-reporter dump interpolating a note list or share payload prints the account's full `ufvk`, note `rseed`/`rho`, the vote-opening blind (`primaryBlind` — i.e. the plaintext vote), and the share nullifier next to its round/proposal. On a *public* API type, third-party apps will log freely. Restore the redaction (and the tests).

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/VotingSdkImpl.kt
```diff
@@ -0,0 +1,379 @@
...
+import cash.z.ecc.android.sdk.model.voting.VotingVoteRecord
+import cash.z.ecc.android.sdk.model.voting.VotingWitness
+
+@Suppress("TooManyFunctions", "LongParameterList")
+internal class VotingSdkImpl(
+    private val backend: TypesafeVotingBackend = TypesafeVotingBackendImpl()
+) : VotingSdk {
+    override suspend fun isAvailable(): Boolean =
+        runCatching { backend.warmProvingCaches() }
+            .fold(onSuccess = { true }, onFailure = { it !is UnsatisfiedLinkError })
```
nesence-m: **Major — `isAvailable()` returns true when the voting .so is missing.** `NativeLibraryLoader` wraps load failures in `AssertionError`, not `UnsatisfiedLinkError`, so `it !is UnsatisfiedLinkError` evaluates true for exactly the case this gate exists to catch — the app proceeds and crashes on the next real voting call. Every other runtime failure (I/O, OOM, `NoClassDefFoundError` from a poisoned static init) also yields true.

Separately: the probe re-runs the expensive Halo2 `warmProvingCaches` on every call with no memoization — seconds of native work for a boolean fixed at library-load time; the KDoc pushes "call once and cache" onto every caller instead of memoizing here.

Fix: treat any load/probe failure as unavailable (or allowlist only genuine success), and memoize the result.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
```diff
@@ -663,21 +698,31 @@ class VotingRustBackend private constructor() {
...
+                "storeVoteTxHash failed for roundId=$roundId bundleIndex=$bundleIndex proposalId=$proposalId"
+            }
+        }
+
+        /**
+         * Idempotently re-marks this vote as submitted using the tx hash [storeVoteTxHash] already
+         * recorded. Fails if no tx hash has been recorded yet for this vote.
+         */
+        @Throws(RuntimeException::class)
+        suspend fun markVoteSubmitted(
```
nesence-m, line 756: **Major — same name, inverted contract.** The old 4-arg `markVoteSubmitted` recorded the tx hash *and* marked submitted; that job moved to `storeVoteTxHash` / atomic `record_vote_submission`. The surviving 3-arg version throws "no vote tx_hash recorded … call store_vote_tx_hash first" unless the hash already exists — and when it does exist, the call is a pure no-op re-record. Every reachable sequence makes it either an error or a no-op.

A caller migrating mechanically (the compiler only flags the removed `txHash` arg) deletes the argument, keeps the call, and crashes at runtime on a flow that previously worked — vote never marked submitted, wallet stalls or re-submits. Remove or rename it rather than shipping a same-name trap across seven plumbing layers.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
```diff
@@ -602,58 +727,44 @@ private class RustVotingDbBackend(
     ): JniVanWitness =
         votingDb.generateVanWitness(roundId, bundleIndex, anchorHeight)
 
-    override suspend fun commitVote(
+    override suspend fun buildVoteCommitment(
```
nesence-m, line 744: **Major — dropped boundary validations on `buildVoteCommitment`.** The base `commitVote` checked (a) the returned commitment's `bundleIndex` echoes the request, and (b) the payload count matches the requested mode; the surviving `JniVoteCommitResult.requireValid` accepts 1 *or* `JNI_VOTE_SHARE_COUNT` payloads regardless of `singleShare`, and the covering test (`commit_vote_rejects_a_result_for_a_different_bundle`) was deleted.

Because `vote::commit` is idempotent, a retry that flips `singleShare` (e.g. app switches to single-share after a crash mid-flow) silently receives the originally persisted 5-payload commitment and proceeds to schedule helper shares under the wrong mode — the base check failed loudly here. Restore both checks (and the test).

backend-lib/src/main/rust/voting/delegation.rs
```diff
@@ -300,37 +339,37 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_pre
...
-/// Builds and proves the delegation ZKP for one bundle. Long-running.
-///
-/// The former `hotkeyRawAddress` parameter is replaced by the full delegation
-/// key material (`fvkBytes`, `hotkeyStoredSecret`, `seedFingerprint`,
-/// `accountIndex`, `roundName`): `zcash_voting` now takes the same typed
-/// `DelegationKeys` the PCZT builder takes, and validates it against the stored
-/// round rather than re-reading pieces of it from the database.
-#[allow(clippy::too_many_arguments)]
 #[unsafe(no_mangle)]
 pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_buildAndProveDelegationNative<
```
nesence-m, line 374: **Moderate — lost the base's `require_round_phase_not_after(RoundPhase::DelegationProved)` fast-fail.** At the pinned crate rev, `build_and_prove_delegation` consults no phase before entering DB loads, PIR fetches, and Halo2 proof generation — so a stale app-level retry (a re-dispatched work item after the vote already advanced) silently re-enters the minutes-long proving pipeline for a round already past submission, burning CPU/battery and rewriting proof rows, instead of failing fast with the base's clear error. Re-proving reproduces the same committed values, so it's wasted work rather than corruption — but nothing replaced the guard. Restore the fast-fail (here or crate-side).

backend-lib/build.gradle.kts
```diff
@@ -145,6 +143,11 @@ cargo {
     // https://developer.android.com/about/versions/15/behavior-changes-all#16-kb
     exec = { spec, _ ->
         spec.environment["RUST_ANDROID_GRADLE_CC_LINK_ARG"] = "-Wl,-z,max-page-size=16384"
+        // chp worktree: shielded voting re-enabled for real-device trial. `mod voting` in lib.rs
+        // is gated behind this cfg (not a Cargo feature, see backend-lib/Cargo.toml), so it must
+        // be threaded through here too or the shipped .so still omits the VotingRustBackend_*
+        // JNI exports even with the zcash_voting dependency uncommented.
+        spec.environment["RUSTFLAGS"] = "--cfg zcash_voting"
```
nesence-m: **Moderate —** two issues on this line: (1) assignment rather than append silently clobbers any ambient developer/CI `RUSTFLAGS` (sccache, `remap-path-prefix`, target-feature flags) for all four ABI builds — different codegen than intended, no warning; append to the inherited value instead. (2) The accompanying comment ("`mod voting` in lib.rs is gated behind this cfg") is false — `lib.rs` declares `mod voting` unconditionally and no `#[cfg(zcash_voting)]` exists in backend-lib, so either the flag is vestigial or it's consumed by a dependency; the comment should say which. (Same stale-rationale problem as the `TypesafeVotingBackendImpl.kt` header still claiming "native symbols are not built in this release" while this PR re-enables them.)

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/voting/VotingModels.kt
```diff
@@ -0,0 +1,452 @@
...
+
+    override fun hashCode(): Int = rk.contentHashCode()
+}
+
+data class VotingDelegationSubmissionResult(
+    val proof: ByteArray,
+    val rk: ByteArray,
+    val spendAuthSig: ByteArray,
+    val sighash: ByteArray,
+    val tx1Effects: ByteArray,
```
nesence-m, line 410: **Moderate — `sighash` vs `tx1Effects`: two adjacent undocumented `ByteArray`s, one of which reproduces the exact bug this PR fixes.** The vote-chain server now rejects `sighash`-based registration with the 400 ("tx1 effects must be 821 bytes, got 0"); the only text distinguishing these fields is a Rust-side comment no sdk-lib consumer will ever see. `sighash` has a legitimate public use (Keystone signature verification), so it can't just be removed — add KDoc marking `tx1Effects` as the sole server-bound field and `sighash` as verification-only, so the compiler-invisible wrong choice at least fails code review.

nesence-m (changes requested):
Deep review of the new `VotingSdk` public API layer — the layering fix itself is the right move (mirrors `MigrationSdk`, kills the app's raw-JNI reach-through). Every finding was adversarially verified against the PR head and the pinned `zcash_voting` rev `60f468a` before being reported; 26 findings survived verification, the 10 most severe are inline. Requesting changes mainly for the confirmation-poll crash (comment on `recovery.rs`), the seed-length regression, and the secrets-redaction regression.

**Verified-and-cleared, for the record:** the Keystone action-shuffle concern and the unclamped progress-bridge candidate were both refuted on verification (index derived post-shuffle and checked against the serialized PCZT; all progress feeder values are literal constants in [0,1] at the pinned rev).

**Confirmed but below the inline cut (16 items, non-blocking):** dead `nullifier` param threaded through all five layers of `recordShareDelegation`; `VotingDelegationPhase.phase` as raw wire `String` where a typed enum is precedented in the same file; stale crate-pin reference (`v1.0.0`/`7d39d02b`) in `VotingRustBackendTest.kt`'s header vs the manifest's `2.0.0-rc.4`/`60f468a`; public API primitives (`networkId: Int`, `accountUuid: String`, `Long` heights, raw PCZT `ByteArray`s) where `ZcashNetwork`/`AccountUuid`/`BlockHeight`/`Pczt` exist; hand-written equals/hashCode replaceable by `FirstClassByteArray`; deleted `delegation_keys()` helper re-inlined; ~80% duplicated commit-result marshalling across three layers; `getDelegationSubmission` redoing full note selection per bundle only to extract keys; five unused imports and one dead mapper in the new sdk-lib files; per-call `SecureRandom().generateSeed(32)`; triple read/parse in `recoverCommittedVoteNative`.

🤖 Review prepared with [Claude Code](https://claude.com/claude-code); all findings human-triaged.

backend-lib/src/main/rust/voting/rounds.rs
```diff
@@ -120,17 +172,12 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_get
     let res = catch_unwind(&mut env, |env| {
         let db = db_from_handle(db_handle)?;
         let _access_lock = db.access_lock()?;
```
dannywillems, line 174: Reviewer note: using `let _access_lock = [...]` is important to actually acquire the lock. If `let _ = [...]` is used, the lock won't be acquired, as `let _` is a pattern. See https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8f8ec6532a2590ffadd57abbdebb97d8

backend-lib/src/main/rust/voting/recovery.rs
```diff
@@ -78,13 +69,50 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_mar
...
-            jint_to_u32(bundle_index, "bundle_index")?,
-            jint_to_u32(proposal_id, "proposal_id")?,
-            &java_string_to_rust(env, &tx_hash)?,
-        )
-        .map_err(|e| anyhow!("mark_vote_submitted: {e}"))?;
+        let round_id = java_string_to_rust(env, &round_id)?;
+        let bundle_index = jint_to_u32(bundle_index, "bundle_index")?;
+        let proposal_id = jint_to_u32(proposal_id, "proposal_id")?;
+        let tx_hash = java_string_to_rust(env, &tx_hash)?;
+        // zcash_voting 1.0.0 folded the old store-then-mark-submitted split into
```
dannywillems: Nit: this is an old comment. We can remove it.

backend-lib/src/main/rust/voting/progress.rs
```diff
@@ -38,46 +33,16 @@ impl ProgressReporter for JniProgressReporter {
     }
 }
 
-fn reporter_from_callback(
+pub(super) fn progress_reporter_from_callback(
     env: &mut JNIEnv<'_>,
     callback: &JObject<'_>,
-) -> anyhow::Result<Option<JniProgressReporter>> {
+) -> anyhow::Result<Box<dyn ProgressReporter>> {
```
dannywillems, line 37: Nit: we could try to avoid using `dyn` later.

backend-lib/src/main/rust/voting/progress.rs
```diff
@@ -4,23 +4,18 @@ use super::*;
...
 }
 
+// zcash_voting 1.0.0 (merged-library patch) renamed this trait from `ProofProgressReporter`
+// to `ProgressReporter`; the single `on_progress(&self, progress: f64)` method is unchanged.
 impl ProgressReporter for JniProgressReporter {
     fn on_progress(&self, progress: f64) {
-        // zcash_voting calls this at coarse milestones outside the spawned Halo2
-        // proving closure. Attach on each callback so the bridge remains correct
-        // if future progress calls come from another native thread.
+        // zcash_voting 0.5.9 calls this at coarse milestones outside the spawned
```
dannywillems: Nit: I think we can remove the version in the comment.

backend-lib/src/main/rust/voting/progress.rs
```diff
@@ -4,23 +4,18 @@ use super::*;
...
-/// impls of `DelegationProgressReporter` and `VoteCommitStageReporter` for every
-/// `ProgressReporter`, so implementing either by hand would collide with them.
-/// The blanket impls forward the clamped proof-progress fraction and drop the
-/// non-proof lifecycle stages, which is exactly this callback's contract.
 struct JniProgressReporter {
     vm: JavaVM,
     callback: GlobalRef,
 }
 
+// zcash_voting 1.0.0 (merged-library patch) renamed this trait from `ProofProgressReporter`
```
dannywillems: Nit: I think we can remove comments that are related to API changes.

backend-lib/src/main/rust/voting/notes.rs
```diff
@@ -40,11 +40,13 @@ fn validate_tree_state_bytes_for_round(
 
     let tree_state =
         TreeState::decode(tree_state_bytes).map_err(|e| anyhow!("decode TreeState: {}", e))?;
-    let orchard_ct = tree_state
-        .orchard_tree()
-        .map_err(|e| anyhow!("parse orchard_tree: {}", e))?;
-    let orchard_root = orchard_ct.root().to_bytes();
-    validate_cached_tree_state_for_round(&tree_state, &orchard_root[..], params)
+    // Voting notes are Ironwood/V3 notes; the round's nc_root is anchored to
```
dannywillems, line 45: Important change!

dannywillems: We should spawn an agent to ensure that no other orchard references exist.

backend-lib/src/main/rust/voting/delegation.rs
```diff
@@ -33,33 +25,37 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_VotingRustBackend_bui
...
-        let keys = delegation_keys(
+        let pczt = build_governance_pczt_for_bundle(
+            &db,
+            &round_id,
+            bundle_index,
+            &notes,
             fvk_bytes,
-            &hotkey,
-            seed_fingerprint,
+            hotkey_secret.expose_secret(),
```
dannywillems, line 48: We should have a deeper look at this kind of code in a follow-up.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
```diff
@@ -48,17 +52,43 @@ private const val PROOF_PROGRESS_REENTRY_ERROR =
...
+/**
+ * Raw JNI bindings to the native shielded-voting backend.
+ *
+ * The only permitted caller is `sdk-lib`'s `TypesafeVotingBackendImpl` — every other consumer,
+ * including the app, must go through the public `cash.z.ecc.android.sdk.VotingSdk` API instead.
+ * That boundary is enforced two ways: this `@Deprecated(ERROR)` (which forces any legitimate
+ * caller to carry an explicit, grep-able `@Suppress("DEPRECATION_ERROR")`), and — once the app
+ * finishes migrating off direct `VotingRustBackend` usage — dropping this module's JNI artifact
+ * from the app's compile classpath entirely, per the CHP feature-module extraction design
+ * (`docs/superpowers/specs/2026-08-10-chp-feature-module-extraction-design.md`).
```
dannywillems, line 70: This seems to be part of a larger task. I guess this will be cleaned after.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
```diff
@@ -108,45 +111,32 @@ internal interface TypesafeVotingDb {
...
+     * An empty [storedSecret] mints a fresh, app-owned random hotkey; a previously persisted
+     * [JniVotingHotkey.storedSecret] deterministically reconstructs the same hotkey. This call
+     * is not scoped to a round.
      */
-    // nosemgrep: kotlin-typesafe-returns-jni-model -- voting internals consume this JNI carrier.
-    suspend fun generateHotkey(
-        roundId: String,
-        networkId: Int
-    ): JniVotingHotkey
+    suspend fun generateHotkey(storedSecret: ByteArray): JniVotingHotkey
```
dannywillems, line 126: Not sure we actually want to pass the stored secret. As mentioned before, we might want to manage this differently.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
```diff
@@ -246,104 +244,80 @@ internal interface TypesafeVotingDb {
...
 
+    suspend fun markVoteSubmitted(roundId: String, bundleIndex: Int, proposalId: Int)
+
     suspend fun getVoteTxHash(
         roundId: String,
         bundleIndex: Int,
         proposalId: Int
     ): VotingTxHashLookup
 
-    /**
```
dannywillems, line 320: Is it expected to remove these comments?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
```diff
@@ -368,16 +342,20 @@ internal interface TypesafeVotingDb {
...
+    suspend fun delegationPhases(roundId: String): List<JniDelegationPhase>
+
+    /**
+     * Clears a bundle's unsigned delegation setup fields (PCZT/rk/sighash) and any stale proof
+     * row so a subsequent construct call starts clean. Does not delete round-level state.
+     */
+    suspend fun resetVotingSessionState(roundId: String)
 }
 
-/**
```
dannywillems, line 373: Nit: is it expected to remove this documentation?

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
```diff
@@ -109,16 +156,31 @@ internal interface VotingBackendBridge {
...
 
-    suspend fun openVotingDb(dbPath: String, walletId: String): VotingDbBackend
+    suspend fun openVotingDb(dbPath: String, walletId: String, networkId: Int): VotingDbBackend
 
     suspend fun computeBundleSetup(notes: List<JniNoteInfo>): JniBundleSetupResult
 
     suspend fun warmProvingCaches()
 
+    suspend fun scheduledShareSubmitAt(
+        nowSeconds: Long,
```
dannywillems, line 166: Reviewer note: to check that the types are correctly matching the Rust.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
```diff
@@ -234,20 +313,13 @@ class VotingRustBackend private constructor() {
             }
         }
 
-        /**
```
dannywillems, line 237: Is it expected to have these comments removed?

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
```diff
@@ -146,30 +134,31 @@ data class JniWireEncryptedShare(
...
- * `rVpk` and `voteAuthSig` are signing material for the cast-vote transaction and
- * must not be logged; [toString] is redacted for that reason. The blinds, share
- * commitments and per-vote randomness that earlier releases carried here are now
- * owned by `zcash_voting` and are never exposed across JNI — recovery reads them
- * back through `getCommitmentBundle`, not from a caller-held copy.
+ * Typed JNI carrier for vote commitment outputs.
  *
- * [sharePayloads] arrives with the commitment rather than from a separate call:
- * the payloads are derived from the same commitment the native side just built.
+ * `shareBlinds`, `rVpk`, and `alphaV` are sensitive reveal/signing inputs.
```
dannywillems, line 141: I'm wondering if these are logged on the Rust side.

backend-lib/src/main/rust/voting/rounds.rs
noop-sk: Checked every `access_lock()` call site in the module against this exact footgun — all use a named binding (`let _access_lock = ...`), never the `let _ = ...` pattern. No bug here. Thanks for the pointer, that's a good one to keep an eye on in review generally.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
noop-sk: Investigated: `markVoteSubmittedNative`/`markVoteSubmitted` is indeed vestigial now (every reachable call is an error-or-no-op, confirmed). Couldn't remove it outright in this PR without breaking zodl-android's `SubmitVotesUseCase`, which still calls it (harmlessly, in the no-op case) after `storeVoteTxHash` on both submission paths — removing the SDK method would need a coordinated app-side change first. Documented it as deprecated/vestigial at every layer (JNI doc comment, typesafe wrapper, public `VotingSdk.markVoteSubmitted` KDoc) with an explicit "do not add new callers" note, and flagged in the app repo that removing it is a follow-up once the app stops calling it.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
noop-sk: Acknowledged — no action taken here, leaving as a follow-up per your note.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
noop-sk: Traced this end to end. On the Rust side, `generateHotkeyNative` (`backend-lib/src/main/rust/voting/notes.rs`) was reading `storedSecret` via the bare `java_bytes` helper into an unwrapped `Vec<u8>` — the *only* hotkey-secret call site in the module that did this. Every sibling (`buildGovernancePcztNative`, `buildAndProveDelegationNative`, `buildVoteCommitmentNative`, `deriveHotkeyRawAddressNative`) reads its hotkey secret via `java_secret_bytes_at_least`/`java_secret_bytes_exact` (`voting/helpers.rs`), which wraps it in `secrecy::SecretVec<u8>` — zeroized on drop, and its `Debug` impl won't print the contents. That was a real, mechanical inconsistency, so I fixed it in b622ae2f: `generateHotkeyNative` now wraps `storedSecret` in `SecretVec` immediately after reading it, matching every other call site.

On the Kotlin side there's no deviation to fix — every hotkey/wallet-seed-shaped secret in this file (`hotkeySecret`, `walletSeed`, `senderSeed`, and `storedSecret` itself) is a bare `ByteArray` with a "must not be logged" KDoc warning; there's no `SecretKey`/wrapper convention anywhere in `sdk-lib` for these, so `storedSecret: ByteArray` isn't an outlier there — it's consistent with how this file already handles every other secret parameter.

That said, this doesn't answer the actual question you're raising, which I read as being about the API *shape*: whether `generateHotkey` should accept a raw stored-secret `ByteArray` for reconstruction at all, versus some other mechanism (a keystore-backed handle, a callback, etc.). I don't have the earlier conversation you're referencing and that's a design call for you/the author, not something I can resolve from the code alone — leaving this open rather than closing it out.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
noop-sk: Checked each one:

- `commitVote`'s doc → `buildVoteCommitment` ended up with **no** KDoc at all, which was a real gap: its content ("builds, signs and stores in one call"; `hotkeyStoredSecret`/now `hotkeySecret` is the persisted secret from `generateHotkey`) is still accurate for the renamed method, it just didn't get carried over. Restored, adapted for the new parameter name, in 88065245.
- `markVoteSubmitted`'s old doc (idempotent hash recording, fails on a *different* hash) described behavior that moved to `storeVoteTxHash` — the Rust side confirms this: `storeVoteTxHashNative` calls `record_vote_submission`, described in its own comment as "the atomic hash+submitted recorder", while `markVoteSubmittedNative` is explicitly commented as vestigial (kept only because zodl-android's `SubmitVotesUseCase` still calls it as a harmless no-op after `storeVoteTxHash`). Both `storeVoteTxHash` and `markVoteSubmitted` were left undocumented on the Kotlin side even though the Rust JNI layer has all this context — restored KDoc on both, including markVoteSubmitted's now-vestigial status.
- `voteSubmission`/`JniVoteSubmission` — removed entirely, not renamed. Verified this is a legitimate cleanup, not a functionality loss: `zodl-android`'s `SubmitVotesUseCase` (the only real consumer) never called `voteSubmission` — it does its whole submit-and-recover flow through `storeVoteTxHash` → `markVoteSubmitted` → `recordVcPosition` → `getCommitmentBundle`/`recoverCommittedVote`. It was speculative pre-confirmation-resend API surface that the app ended up not needing (it already holds the built commitment from `buildVoteCommitment`'s synchronous return). `recoverCommittedVoteNative` also isn't a substitute — it requires a recorded vc-tree position and errors otherwise, so it doesn't cover the "before confirmation" case `voteSubmission` targeted; that case is just unused now.
- `recordVcPosition`'s doc wasn't deleted, just reworded (still present, still accurate for the current signature) — but the reword dropped the cross-reference "`getCommitmentBundle` reports null until this has been called", and `getCommitmentBundle` itself had never had its own KDoc. Given the Rust side has a whole comment explaining exactly why that null is a normal, expected pre-confirmation state (not an error) — restored that on `getCommitmentBundle` directly, pointed at `storeVoteTxHash` (not the vestigial `markVoteSubmitted`) as the confirmation trigger.

All four restorations are in 88065245.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
noop-sk: Traced this through `git log -p` on this file — the removed doc block is `VotingNoteInfo`'s (the `internal` one in this file, distinct from the public `cash.z.ecc.android.sdk.model.voting.VotingNoteInfo`). It went missing in commit `f3e55aae` ("Fix build after merge: restore our voting/ module wholesale"), a merge-conflict-resolution commit that clobbered more than just the KDoc:

```
-/**
- * The typesafe view of a spendable note the voting backend may draw voting weight from.
- *
- * [toString] is redacted: [rseed] and [rho] reconstruct the note's spending randomness,
- * [nullifier] links the note to its spend, and [ufvk] is a full viewing key that discloses
- * the entire account's transaction history. The generated `data class` rendering would print
- * all four into any log line that interpolates a note.
- */
 internal data class VotingNoteInfo(
     ...
-    override fun toString(): String = "VotingNoteInfo(redacted)"
```

That same commit dropped the `toString()` redaction override, not just the comment describing it — so this type has been printing `rho`/`rseed`/`nullifier`/`ufvk` in plaintext via the default `data class` `toString()` since then. The prior review round's fix (C) ("restore toString redaction on note/share-payload/delegation models") caught and fixed this exact regression on the *public* `VotingNoteInfo`/`VotingSharePayload`/`VotingShareDelegationRecord` and their JNI mirrors, but missed this internal intermediate-layer type, which carries the identical sensitive fields.

Restored both the KDoc and the `toString()` override (byte-for-byte what the merge fixup deleted), plus a covering test asserting the sensitive values don't appear in the rendered string — commit 88065245.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackendImpl.kt
noop-sk: Traced both `openVotingDb`'s new `networkId` and `scheduledShareSubmitAt`'s parameters end to end; both check out.

**`openVotingDb`'s `networkId: Int`:**
`VotingRustBackend.openVotingDb(dbPath, walletId, networkId: Int)` → `openVotingDbNative(dbPath, walletId, networkId)` (`external fun ...: Long`, `VotingRustBackend.kt:1053`) → Rust `Java_..._openVotingDbNative(..., network_id: jint)` (`voting/db.rs:119-146`) → `voting_network_from_id(network_id)` (`voting/helpers.rs:394-399`), which matches `NETWORK_ID_TESTNET = 0` / `NETWORK_ID_MAINNET = 1` (`voting/helpers.rs:131-132`). That's the same 0/1 convention `ZcashNetwork.ID_TESTNET`/`ID_MAINNET` already use elsewhere in `sdk-lib` (`ZcashNetwork.kt`), and the same convention zodl-android's `ZcashNetwork.toVotingNetworkId()` call sites already produce (`if (isMainnet()) 1 else 0`). No mismatch — this is the existing network-id convention this JNI boundary already used for `extractOrchardFvkFromUfvk`/`deriveHotkeyRawAddress`, just now also threaded through `openVotingDb` so the DB handle itself carries the network (`db.network`) instead of every round-scoped call repeating it (see `initRound`'s dropped `networkId` param, a few hunks down — same reasoning).

**`scheduledShareSubmitAt`'s params:**
`nowSeconds`/`ceremonyStartSeconds`/`voteEndTimeSeconds: Long` → `scheduledShareSubmitAtNative(..., entropy: ByteArray): Long` (`jlong` throughout, `VotingRustBackend.kt:994-1000`) → Rust `now_seconds`/`ceremony_start_seconds`/`vote_end_time_seconds: jlong`, `single_share: jboolean`, `entropy: JByteArray` (`voting/share_tracking.rs:37-71`). Each `jlong` goes through `jlong_to_u64` (checked, rejects negative via `u64::try_from`), `single_share` is compared against `JNI_TRUE`, `entropy` (32 bytes from `SecureRandom`) is checked `>= 8` bytes via `require_min_len`. The `Long` return goes through `u64_to_jlong` (checked, rejects overflow via `jlong::try_from`). All conversions are the fallible, checked helpers in `voting/helpers.rs` — no truncating cast anywhere in this path.

No bug found; nothing changed here.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/VotingRustBackend.kt
noop-sk: Found it via the full base...head patch (the net merge-base→HEAD diff doesn't show it, since it was added and then edited within this branch's own history). It's `initRound`'s KDoc:

```
-        /**
-         * Creates a round and binds it to [networkId].
-         *
-         * [roundId] must be 64 lowercase hex characters encoding a canonical Pallas field
-         * element; the native side rejects anything else.
-         */
         @Throws(RuntimeException::class)
         suspend fun initRound(
             roundId: String,
             ...
-            networkId: Int,
             sessionJson: String?
```

The doc's first sentence ("binds it to `[networkId]`") went stale because this PR removes `initRound`'s `networkId` parameter entirely — network is now fixed at `openVotingDb` time (`db.network` in `voting/rounds.rs:34`: `db.init_round(db.network, &params, ...)`), so a round no longer takes its own network id. But the second sentence — the 64-lowercase-hex-Pallas-field-element format requirement on `roundId` — has nothing to do with `networkId` and is still enforced (`roundId` is still the same first parameter, unchanged validation path into `zcash_voting`'s `init_round`). Deleting the whole comment along with the stale sentence was collateral damage, not intentional.

Restored the still-accurate part, reworded the first sentence to reflect that the round is now bound to whatever network the already-open db carries — commit 88065245.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/model/voting/JniVotingModels.kt
noop-sk: Checked specifically for Rust-side logging (separate from the Kotlin `toString` redaction, which is already handled — `JniVoteCommitmentResult.toString()` is `"JniVoteCommitmentResult(redacted)"` and the "must not be logged" sentence did survive the KDoc rewrite, it's just not visible in the truncated hunk).

On the Rust side (`backend-lib/src/main/rust/voting/`): grepped the whole module for `log::`, `tracing::`, `println!`, `eprintln!`, `dbg!`. There are exactly two logging calls in the entire module, both in `voting/progress.rs`, both about JNI plumbing failures in the proof-progress callback (`"proof progress callback failed: {e}"`, `"attach_current_thread for progress callback failed: {e}"`) — neither touches `share_blinds`/`r_vpk`/`alpha_v` or anything derived from them. The only `{:?}` Debug-format calls in the module are on a PCZT parse error and a round-phase enum (`voting/delegation.rs`) — again unrelated. `JniVoteCommitmentResultPayload` (the Rust struct these three fields land in before crossing JNI, `voting/helpers.rs:148-160`) has no `#[derive(Debug)]` and no manual `Debug` impl, so it isn't even `{:?}`-printable from this crate's own code. Field-name strings like `"shareBlinds"`/`"rVpk"`/`"alphaV"` do appear in a few `anyhow!` error messages, but only as the label in bounds-checking errors like `"{field} must be exactly {expected} bytes, got {}"` — that interpolates the byte *count*, never the byte content.

One thing worth flagging even though it's not a live exposure: the upstream `zcash_voting` crate's `VoteRecoveryBundle` (`src/vote.rs:500-501`, the source these three fields get copied from via `from_recovery_bundle`) does derive `#[derive(Clone, Debug)]`. Nothing in this repo currently `{:?}`-formats it or logs it, so today it's clean, but it means the redaction is a discipline this repo's own code has to keep maintaining rather than something the dependency's type system enforces — a future `tracing::debug!("{:?}", bundle)` dropped in for troubleshooting would silently reintroduce the leak. Not something to fix now, just noting it since you asked specifically about the Rust side.

Confirmed clean today — resolving.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/VotingSdk.kt
```diff
@@ -0,0 +1,350 @@
...
+
+    /**
+     * Vestigial: [storeVoteTxHash] is now the sole atomic recorder for "this vote's tx hash is
+     * known and it is submitted" — that single call already does everything this method used
+     * to. Calling this after [storeVoteTxHash] is always a harmless no-op (it re-asserts the
+     * same already-recorded hash); calling it before [storeVoteTxHash] for the same vote throws.
+     * Kept only for existing callers — do not add new call sites, rely on [storeVoteTxHash]
+     * alone instead.
+     */
+    suspend fun markVoteSubmitted(roundId: String, bundleIndex: Int, proposalId: Int)
```
nesence-m, line 309: **Medium (API design) — ship this with `@Deprecated`, or don't ship it on the public surface at all.** The KDoc is thorough, but KDoc is invisible at call sites and nothing here is compiler-enforced — no `@Deprecated` exists at any layer (grep-verified). Two points:

1. `@Deprecated("Redundant; storeVoteTxHash already records the hash and marks submitted", level = DeprecationLevel.WARNING)` costs one line, keeps zodl-android compiling (warning only), and is the only mitigation the next accidental caller actually sees.
2. Stronger: `VotingSdk` is a brand-new public interface in this PR, and the app has to migrate onto it anyway — so this PR *is* the coordinated-change window the Rust-side doc says removal requires. The public surface could simply omit `markVoteSubmitted` (keep it on the internal `TypesafeVotingDb` if staging needs it); the app's call is documented as always the harmless no-op case, i.e. a two-line deletion in `SubmitVotesUseCase`.

Minimum ask before merge: the annotation.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/VotingSdkImpl.kt
```diff
@@ -0,0 +1,403 @@
...
+
+    // Probing availability warms the (expensive) Halo2 proving caches as a side effect, so the
+    // result is computed at most once per process and cached here rather than on every call.
+    // Any failure -- not just UnsatisfiedLinkError -- means unavailable: NativeLibraryLoader
+    // wraps a failed System.loadLibrary in AssertionError, not UnsatisfiedLinkError, so a
+    // `!is UnsatisfiedLinkError` check would previously report "available" for exactly the
+    // missing-native-library case this gate exists to catch.
+    override suspend fun isAvailable(): Boolean =
+        cachedIsAvailable ?: isAvailableMutex.withLock {
+            cachedIsAvailable ?: runCatching { backend.warmProvingCaches() }.isSuccess.also {
```
nesence-m: **Minor — `runCatching` swallows `CancellationException` and memoizes `false` for the process lifetime.** `warmProvingCaches()` is a suspend call; if the first probing coroutine is cancelled mid-probe (screen rotation, scope teardown), the cancellation is caught, `isSuccess == false` is cached, and every later call reports voting unavailable despite a healthy native library. Memoizing genuine probe failures is a fine design choice; cancellation isn't a probe failure — rethrow it and leave the cache unset:

```kotlin
runCatching { backend.warmProvingCaches() }
    .onFailure { if (it is CancellationException) throw it }
    .isSuccess
```

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
```diff
@@ -321,7 +300,7 @@ internal interface TypesafeVotingDb {
      * Reconstructs the stored vote commitment, with fresh helper-share payloads.
      *
      * Reports null until the vote reaches the confirmed phase — its transaction hash recorded via
-     * [markVoteSubmitted] *and* its tree position via [recordVcPosition] — and for a vote that was
+     * [storeVoteTxHash] and its tree position via [recordVcPosition] — and for a vote that was
```
nesence-m: Nit (doc precision): the actual non-null gate crate-side is `commitment_bundle_json` + `vc_tree_position` only — tx-hash presence is never consulted (`record_vc_position` doesn't require it either). In the real flow the position is only recorded post-confirmation, so this reads correctly in practice; strictly, the [storeVoteTxHash] clause over-specifies. "typically" — or dropping that clause — would make it exact.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/VotingSdkImpl.kt
```diff
@@ -0,0 +1,403 @@
+package cash.z.ecc.android.sdk.internal
+
+import cash.z.ecc.android.sdk.VotingDbSession
+import cash.z.ecc.android.sdk.VotingSdk
+import cash.z.ecc.android.sdk.internal.model.voting.JniDelegationPhase
```
nesence-m: Nit: five unused imports in this file — `JniDelegationPhase` (L5), `JniRoundPhase` (L6), `VotingEncryptedShare` (L18), `VotingNoteScope` (L22), `VotingRoundPhase` (L23). None of these symbols appears elsewhere in the file.

nesence-m (dismissed):
Re-review at head `88065245`. Every blocking finding from my first pass verified fixed — against the head tree and the pinned rc.5 crate source, not just the fix-commit diffs: the confirmation-window null mapping matches rc.5's exact error text (crate `queries.rs:2481`, full-message match, can't swallow the adjacent real errors); the exact-64 seed check with a 32..63-gap test; toString redaction restored on all seven types including the internal `VotingNoteInfo`, with tests asserting sensitive values absent; `isAvailable()` inverted and properly memoized (double-checked lock); both `buildVoteCommitment` boundary checks and their tests; and the restored `DelegationProved` fast-fail is provably safe for multi-bundle flows — in rc.5, submission never advances the round phase (only proving→`DelegationProved` and vote-commit→`VoteReady` write it), so after bundle 0 is proved+submitted the round sits at `DelegationProved`, which the `>` check permits; the only rejected state is one where the crate itself inevitably fails *after* the minutes-long proof. RUSTFLAGS append handles the unset case cleanly. The two threads that were resolved without a named fix commit (provisional-payload KDoc on `VotingCommitResult`, `sighash`-vs-`tx1Effects` KDoc) turned out to be real fixes riding unadvertised in ff580be7 — both verified at head.

The new `storeKeystoneSignature` plumbing reviewed clean as new code: stricter-than-crate length validation (64/32/32 via `java_bytes_exact`), checked JNI conversions, FK-protected inserts, sibling-pattern-consistent at all four layers, and the crate-side preservation path (`clear_unsigned_delegation_setup_fields`'s `keystone_signatures` exclusion) verified real including the crate's own covering test. The stale-proofs-gap KDoc on `resetVotingSessionState` is accurate at both layers.

Hygiene: CI 23/23 green at head; no active `[patch.crates-io]` entries (only the commented template — releasable state per policy); `cargo tree -d` shows zero family-crate duplicates; `cargo tree -i` gives exactly one node each for `zcash_protocol`/`orchard`/`sapling-crypto`/`rusqlite`/`libsqlite3-sys`; `--locked` resolves clean.

Approving with comments. The inline `@Deprecated` ask on `markVoteSubmitted` is the one I'd like handled before merge (one annotation). Below the inline cut:

- `ORCHARD_WITNESS_PATH_DEPTH` (`voting/helpers.rs:123` — pre-existing line, so not inline-anchorable): same tree-vs-keys naming drift 109fb026 fixed elsewhere; it validates `authPath` witnesses over the *Ironwood* tree, so `NOTE_COMMITMENT_WITNESS_PATH_DEPTH` (or similar) would finish that rename. No correctness impact — the depths are identical by design.
- The confirmation-window fix remains a string match on crate-internal error text; any future rc rewording silently regresses to throwing. Fine for rc.5 and documented in the KDoc — worth a typed error upstream eventually.
- None of the 16 cleanup-tier items from the first pass were picked up (dead `nullifier` param, unused imports in `VotingSdkImpl.kt` — see inline, per-call `SecureRandom().generateSeed(32)`, triple read in `recoverCommittedVoteNative`, raw wire `String` phase). Still non-blocking.

Explicitly not preempted by this approval: Danny's open stored-secret API-shape thread — that's a design decision that should resolve on its own track.

🤖 Review prepared with [Claude Code](https://claude.com/claude-code); all findings human-triaged.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/VotingSdk.kt
noop-sk: Done — added @Deprecated(level = DeprecationLevel.WARNING) to markVoteSubmitted on the public VotingDbSession interface (VotingSdk.kt) and to its VotingDbSessionImpl override (VotingSdkImpl.kt), per the minimum ask. Commit b3034116.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/VotingSdkImpl.kt
noop-sk: Fixed exactly as suggested — isAvailable() now does .onFailure { if (it is CancellationException) throw it } before .isSuccess, using kotlinx.coroutines.CancellationException, so cancellation propagates without touching cachedIsAvailable. Genuine probe failures are still memoized as before (Fix (E) untouched). Added a regression test (isAvailable_rethrows_cancellation_and_does_not_memoize_it) asserting the exception propagates. Commit eb0f8f19.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/TypesafeVotingBackend.kt
noop-sk: Dropped the storeVoteTxHash clause rather than softening it with "typically" — reads cleaner and matches the actual crate-side gate (commitment_bundle_json + vc_tree_position only). KDoc now reads "Reports null until the vote reaches the confirmed phase — its tree position recorded via [recordVcPosition] — and for a vote that was never stored." Commit e177a881.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/VotingSdkImpl.kt
noop-sk: Removed all 5 — JniDelegationPhase, JniRoundPhase, VotingEncryptedShare, VotingNoteScope, VotingRoundPhase. Confirmed unused via grep and a clean compile (:sdk-lib:compileDebugKotlin, ktlint, detektAll all pass). Commit 315496c7.

nesence-m (approved):
Lgtm

PR #2169, opened 2026-08-12
Fix spendable balance pinned at zero after Orchard→Ironwood migration (MOB-1667)

nesence-m:
Fixes [MOB-1667](https://linear.app/zodl/issue/MOB-1667/unable-to-sync-on-android-following-update-to-v390): after updating to Zashi 3.9.0 (SDK 3.0.1), users who had manually migrated Orchard→Ironwood saw a correct total balance but spendable pinned at 0 indefinitely, with a `SYNCED` status and no way to send.

Three defects chained into that state; each commit fixes one:

1. **Bound the `readyToBroadcast` sync-block leg** (`5aefa9f0`). Unlike the other two legs of `isSyncBlockedNow` (both self-expiring), the readiness leg had no time bound — and its only escape, stale-plan expiry, is evaluated against the *scanned* tip, which cannot advance while the block itself keeps sync paused. A migration driver the OS never runs again (aggressive OEM worker scheduling — the reporter is on Xiaomi/MIUI) paused sync forever. Readiness alone now blocks for at most three privacy sync buffers (~30 min mainnet) of continuous untouched readiness, tracked by a persisted stamp that every live `executeNextPendingTransfer` attempt re-arms — a working driver keeps its full decorrelation window; only a dead one lets the cap elapse. The actual send stays protected by the unchanged in-flight and post-broadcast buffer legs.
2. **`pause()` keeps observing; `resume()` restarts a stopped engine** (`ae1ee028`, `851f1f16`). The poll loop performs only local JNI reads — the network session it was meant to decorrelate was never gated here at all — so stopping it merely froze the balances, heights and status the host renders, behind a status forced to `SYNCED`. Pausing now gates just `syncBurst` and transaction resubmission, polling continues, status is truthful, and `resume()` restarts an engine that a preceding `onBackground()` stopped instead of polling a dead session. `restoreAfterBurst()` follows the same rule for a pause landing mid-burst.
3. **The stale-tip spendable mask fails open after 15 minutes of live stale ticks** (`403725b9`). `tipFresh` latches only after a fully successful serial network prologue, and the engine forces `is_recovering = 0` in error states — closing the mask's one escape exactly when the engine is wedged. Any persistent early-prologue failure pinned spendable at zero with no exit. Showing spendable against an unverified tip cannot lose funds (a stale-tip spend fails at propose or broadcast); hiding it indefinitely bricks the wallet.

Recovery for currently-wedged users once released: the existing stuck `STEP_BROADCAST` state arms the cap on first poll, sync resumes within ~30 min even if WorkManager never fires, balances are live immediately, and `tipFresh` latches on the first completed prologue.

Kotlin-only; no Rust/FFI, schema, or dependency changes. `sdk-lib` unit suites (261 tests), `detektAll`, `ktlint`, sdk-incubator-lib/lightwallet-client-lib suites and the demo-app compile are green. Note for reviewers: app `maint/v3.9.x` does not compile against this branch's *base* either (pre-existing voting-API drift in `VotingCryptoClient.kt`), so no end-to-end app assemble was possible from this SDK line.

Deferred follow-ups (out of scope here): JNI-layer independent tip corroboration (needs a Tor-policy review); a first-class `PAUSED` `Synchronizer.Status` (public API change); gating the native session itself during decorrelation windows; app-side expedited/foreground migration worker; a persisted tip-freshness grace window if the tick-based fail-open proves insufficient.

> **Note**

# Author

# Reviewer

### DIFF
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/OrchardMigrationSdkImpl.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/OrchardMigrationSdkImpl.kt
@@ -918,6 +918,7 @@ internal class OrchardMigrationSdkImpl(
         options: NetworkPrivacyOptions,
         useEstimatedTip: Boolean,
     ): TransferAttemptOutcome {
+        reStampSyncBlockReadySince()
         val preparation =
             logged("executeNextPendingTransfer.prepare") {
                 prepareNextPendingTransfer(options, useEstimatedTip)
@@ -1189,6 +1190,8 @@ internal class OrchardMigrationSdkImpl(
                 }
             }
         val nowEpochSeconds = Clock.System.now().epochSeconds
+        val readyToBroadcastWithinCap =
+            boundedReadyToBroadcast(preferenceProvider, readyToBroadcast, nowEpochSeconds)
         val resumeAtEpochSeconds =
             preferenceProvider.getString(EncryptedPreferenceKeys.MIGRATION_SYNC_RESUME_AT.key)?.toLongOrNull()
         val bufferActive = resumeAtEpochSeconds != null && resumeAtEpochSeconds > nowEpochSeconds
@@ -1196,7 +1199,71 @@ internal class OrchardMigrationSdkImpl(
             preferenceProvider.getString(EncryptedPreferenceKeys.MIGRATION_BROADCAST_IN_FLIGHT_UNTIL.key)?.toLongOrNull()
                 ?: 0L
         val broadcastInFlight = isBroadcastInFlight(nowEpochSeconds, inFlightUntilEpochSeconds)
-        return readyToBroadcast || bufferActive || broadcastInFlight
+        return readyToBroadcastWithinCap || bufferActive || broadcastInFlight
+    }
+
+    /**
+     * Caps how long a `STEP_BROADCAST` readiness alone may keep sync blocked, using the
+     * [EncryptedPreferenceKeys.MIGRATION_SYNC_BLOCK_READY_SINCE] stamp.
+     *
+     * The other two legs of [isSyncBlockedNow] self-expire; this one cannot. A transfer stays ready
+     * to broadcast until some driver broadcasts it, and the plan's own stale-plan expiry is
+     * evaluated against the SCANNED tip — which only advances while sync runs, i.e. exactly what
+     * this block stops. A driver that never runs again (a killed background worker, as on the
+     * aggressive OEM schedulers behind MOB-1667) therefore pauses sync forever, and a permanently
+     * paused sync never refreshes the tip.
+     *
+     * [MIGRATION_SYNC_BLOCK_READY_SINCE][EncryptedPreferenceKeys.MIGRATION_SYNC_BLOCK_READY_SINCE]
+     * is cleared as soon as readiness goes away and re-armed by every live
+     * [executeNextPendingTransfer] attempt, so a working driver keeps the block indefinitely and
+     * only a dead one lets the cap elapse.
+     */
+    private suspend fun boundedReadyToBroadcast(
+        preferenceProvider: PreferenceProvider,
+        readyToBroadcast: Boolean,
+        nowEpochSeconds: Long,
+    ): Boolean {
+        val key = EncryptedPreferenceKeys.MIGRATION_SYNC_BLOCK_READY_SINCE.key
+        val stamp = preferenceProvider.getString(key)
+        val readySinceEpochSeconds = stamp?.toLongOrNull()
+        return when {
+            !readyToBroadcast -> {
+                /*
+                 * Written only when there is something to clear: this poll runs every
+                 * SYNC_BLOCK_TICK, and a write wakes every preference observer.
+                 */
+                if (!stamp.isNullOrEmpty()) preferenceProvider.putString(key, "")
+                false
+            }
+
+            readySinceEpochSeconds == null -> {
+                preferenceProvider.putString(key, nowEpochSeconds.toString())
+                true
+            }
+
+            else -> {
+                isReadyToBroadcastBlockWithinCap(
+                    nowEpochSeconds = nowEpochSeconds,
+                    readySinceEpochSeconds = readySinceEpochSeconds,
+                    capSeconds =
+                        privacySyncBufferDuration().inWholeSeconds * READY_TO_BROADCAST_BLOCK_CAP_MULTIPLIER,
+                )
+            }
+        }
+    }
+
+    /**
+     * Re-arms [EncryptedPreferenceKeys.MIGRATION_SYNC_BLOCK_READY_SINCE] at the start of every
+     * transfer attempt, so [boundedReadyToBroadcast]'s cap measures "how long has a ready transfer
+     * gone untouched by any driver", not "how long has it been ready". A driver that keeps
+     * attempting keeps sync blocked for as long as it needs.
+     */
+    private suspend fun reStampSyncBlockReadySince() {
+        val nowEpochSeconds = Clock.System.now().epochSeconds
+        preferenceProviderHolder().putString(
+            EncryptedPreferenceKeys.MIGRATION_SYNC_BLOCK_READY_SINCE.key,
+            nowEpochSeconds.toString(),
+        )
     }
 
     private suspend fun isReadyToBroadcast(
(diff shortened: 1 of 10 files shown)


noop-sk (commented):
Overall a solid, well-documented fix for all three chained defects, and the CHANGELOG/tests are thorough. Left inline comments — one real regression risk in `resume()`, plus a few smaller observations worth a second look.

sdk-lib/src/main/java/com/zodl/slipstream/SlipstreamSynchronizer.kt
```diff
@@ -1402,20 +1406,37 @@ class SlipstreamSynchronizer internal constructor(
...
+     */
     override fun resume() {
         if (closed.get()) return
         migrationPaused.update { false }
         launchGuarded("resume") {
             if (!awaitPrepared()) return@launchGuarded
+            if (inForeground.get() && !engine.isRunning) {
+                engine.start(ufvk = null, birthday = startBirthday.value)
+            }
             engine.startPolling()
```
noop-sk, line 1444: `engine.startPolling()` here is unconditional, but the doc comment above claims *"A backgrounded wallet is left stopped - the next onForeground restarts it."* The code doesn't do that — only the `engine.start(...)` call is gated by `inForeground.get()`, `startPolling()` is not.

`SlipstreamEngine.startPolling()` has no `isRunning` guard (only `handle == 0L`), so this restarts the 2s poll loop (`tick()` → `snapshot`/`drainEvents`/`walletSummary` over JNI) against an engine that `onBackground()` just stopped via `engine.stopPolling(); engine.stop()`.

This is directly reachable: `WalletCoordinator.kt` drives `pause()`/`resume()` straight off the `isSyncBlocked` flow, independent of app foreground state:
```kotlin
isSyncBlocked.distinctUntilChanged().collect { blocked ->
    if (blocked) closeableSynchronizer.pause() else closeableSynchronizer.resume()
}
```
This very PR adds the mechanism (the readiness cap) by which `isSyncBlocked` can flip `true → false` on its own, in the background — the exact MOB-1667 scenario. When that happens while the app is backgrounded, `resume()` restarts the poll loop against a stopped native session indefinitely, undermining the battery/network intent of `onBackground()`.

Compare with `restoreAfterBurst()` a bit further down, which handles this correctly:
```kotlin
!inForeground.get() -> { engine.stopPolling(); engine.stop() }
else -> { engine.startPolling() }
```
Suggest gating `startPolling()` here the same way `start()` is gated (or mirroring `restoreAfterBurst`'s branch), so a background `resume()` leaves the engine genuinely stopped.

sdk-lib/src/test/java/com/zodl/slipstream/SlipstreamSynchronizerLifecycleTest.kt
```diff
@@ -412,28 +446,84 @@ class SlipstreamSynchronizerLifecycleTest {
...
+        val synchronizer = buildSynchronizer(engine = engine, key = key)
+        try {
+            `when`(engine.isRunning).thenReturn(false)
+            synchronizer.pause()
+            clearInvocations(engine)
+
+            synchronizer.resume()
+
+            runBlocking {
+                verify(engine, timeout(TIMEOUT_MS)).startPolling()
```
noop-sk: This test's name ("leaves a stopped engine stopped") and the fact that `buildSynchronizer()` never calls `onForeground()` (so `inForeground` is `false` throughout) actually demonstrate the `resume()` issue flagged on `SlipstreamSynchronizer.kt`: `startPolling()` is asserted to be called even while backgrounded. I'd expect this test to assert `verify(engine, never()).startPolling()` once the underlying `resume()` gating is fixed.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/OrchardMigrationSdkImpl.kt
```diff
@@ -1189,14 +1190,80 @@ internal class OrchardMigrationSdkImpl(
...
+        }
+    }
+
+    /**
+     * Re-arms [EncryptedPreferenceKeys.MIGRATION_SYNC_BLOCK_READY_SINCE] at the start of every
+     * transfer attempt, so [boundedReadyToBroadcast]'s cap measures "how long has a ready transfer
+     * gone untouched by any driver", not "how long has it been ready". A driver that keeps
+     * attempting keeps sync blocked for as long as it needs.
+     */
+    private suspend fun reStampSyncBlockReadySince() {
```
noop-sk, line 1266: Minor design note, not necessarily a bug: `reStampSyncBlockReadySince()` re-arms the stamp on *every* `executeNextPendingTransfer` attempt regardless of outcome (including `NothingDue`/failed broadcasts). That means as long as the driver process itself keeps running periodically, the readiness leg of the sync block never expires — even if broadcasting has been failing continuously for a long time (e.g. a persistent network outage), since the cap is designed to catch "driver stopped running entirely," not "driver runs but never succeeds." That matches the documented MOB-1667 root cause (killed background worker), so probably fine as-is, but might be worth a one-line callout in the doc comment so a future reader doesn't assume the cap also bounds a live-but-perpetually-failing driver.

sdk-lib/src/main/java/com/zodl/slipstream/internal/SlipstreamEngine.kt
```diff
@@ -48,6 +48,13 @@ internal class SlipstreamEngine(
     private var lastStartBirthday: Long = 0L
     private var running = false
 
+    /**
+     * Consecutive ticks whose snapshot reported a stale tip; feeds [effectiveTipFresh]'s fail-open.
+     * Counted in ticks rather than wall-clock time on purpose - it must measure how long the mask
+     * has been applied to LIVE observations, and no tick means no observation to mask.
+     */
+    private var staleTipTicks = 0
```
noop-sk, line 56: `staleTipTicks` is in-memory only and resets on process restart (deliberately, per the comment above). Worth double-checking this composes correctly with the `resume()` behavior flagged elsewhere in this PR: if the poll loop gets restarted against a stopped/not-actually-`start()`-ed engine (see the `SlipstreamSynchronizer.resume()` comment), the ticks it produces may not represent real network observations, which could skew this counter and delay (or prematurely trigger) the 15-minute fail-open in ways this specific interaction isn't tested for.

sdk-lib/src/test/java/com/zodl/slipstream/SlipstreamSynchronizerLifecycleTest.kt
```diff
@@ -1504,6 +1633,18 @@ class SlipstreamSynchronizerLifecycleTest {
...
 
+    /**
+     * The `onTick` callback the synchronizer's `init` installs on the engine. A mock records the
+     * setter call but its getter answers `null`, so the lambda has to be captured from the call
+     * itself rather than read back.
+     */
+    @Suppress("UNCHECKED_CAST")
+    private fun captureOnTick(engine: SlipstreamEngine): suspend (SlipstreamSnapshot) -> Unit {
+        val captor = ArgumentCaptor.forClass(Any::class.java)
+        verify(engine).onTick = captor.capture() as? (suspend (SlipstreamSnapshot) -> Unit)
```
noop-sk, line 1651: Nit: `ArgumentCaptor.forClass(Any::class.java)` + unchecked cast to capture the `onTick` setter works, but is a bit fragile (silently produces a bad cast exception if the setter signature ever changes shape). Not blocking — just flagging in case there's an existing helper elsewhere in the test suite for capturing Kotlin lambda-typed property setters that's less ad hoc.

sdk-lib/src/main/java/com/zodl/slipstream/SlipstreamSynchronizer.kt
nesence-m: Good catch — fixed in d244bdbe: `resume()` now early-returns before touching the engine when backgrounded, mirroring `restoreAfterBurst()`. A background resume clears only the pause flag; the next `onForeground()` restarts the engine and polling.

sdk-lib/src/test/java/com/zodl/slipstream/SlipstreamSynchronizerLifecycleTest.kt
nesence-m: Flipped in d244bdbe — the test now asserts `never()).startPolling()` (and the KDoc says why), matching the fixed `resume()` gating.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/OrchardMigrationSdkImpl.kt
nesence-m: Agreed, that asymmetry is intentional — added the callout to `reStampSyncBlockReadySince()`'s KDoc in d244bdbe: the cap bounds a driver that stopped running entirely, never a live-but-perpetually-failing one, and each individual attempt is still bounded by the in-flight and privacy-buffer legs.

sdk-lib/src/main/java/com/zodl/slipstream/internal/SlipstreamEngine.kt
nesence-m: The `resume()` gating fix in d244bdbe closes the interaction you're describing: polling can no longer be restarted against a stopped engine, so stale ticks only accrue while the poll loop reads a genuinely started session (foreground resume starts the engine before polling; background resume leaves both stopped).

sdk-lib/src/test/java/com/zodl/slipstream/SlipstreamSynchronizerLifecycleTest.kt
nesence-m: Leaving this one as-is: type erasure makes any lambda-typed captor unchecked in the end, and the helper at least centralizes the cast in a single documented place. Happy to revisit if a typed pattern lands in the suite.

Chlup (approved):
utACK

PR #2174, opened 2026-08-14
MOB-1541: Tolerate unknown-pool subtree root fetch failures without retry

### LINKED ISSUE #2058: Ironwood subtree-root fetch retries, then silently degrades, against a lightwalletd that does not know the pool
Follow-up from the review of #2051.

`CompactBlockProcessor.getSubtreeRoots` now fetches subtree roots for three pools:

```kotlin
val saplingSubtreeRootList = fetchSubtreeRoots(saplingStartIndex, ShieldedProtocolEnum.SAPLING)
val orchardSubtreeRootList = fetchSubtreeRoots(orchardStartIndex, ShieldedProtocolEnum.ORCHARD)
val ironwoodSubtreeRootList = fetchSubtreeRoots(ironwoodStartIndex, ShieldedProtocolEnum.IRONWOOD)
```

`fetchSubtreeRoots` wraps each call in `retryUpToAndContinue(GET_SUBTREE_ROOTS_RETRIES)`, which — per its implementation in `sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/WalletService.kt` — swallows the exception once retries are exhausted rather than rethrowing. The failure does set `result = GetSubtreeRootsResult.OtherFailure(...)`, but the trailing

```kotlin
if (saplingSubtreeRootList.isNotEmpty()) {
    result = GetSubtreeRootsResult.SpendBeforeSync(...)
}
```

overwrites it whenever the Sapling fetch succeeded.

### Consequence

Against a lightwalletd that does not recognise `ShieldedProtocol.ironwood` — i.e. any server that has not upgraded — each spend-before-sync initialisation will:

1. retry the Ironwood request `GET_SUBTREE_ROOTS_RETRIES` times with exponential backoff, then
2. proceed with an empty Ironwood root list, as though the request had succeeded and the pool were simply empty.

The wallet still syncs, but it pays the full retry backoff on every SbS init and logs errors that read as fatal but are not.

### Not a regression, but newly material

`maint/v2.6.x` had the identical overwrite for Orchard before the merge, and the loop refactor in #2051 is behaviour-preserving — I verified both. What changes is materiality: Orchard-unaware servers effectively no longer exist, whereas Ironwood-unaware servers are currently the norm.

### Suggestions

Either would address it:

- Distinguish "server does not know this pool" from "fetch failed", so the former short-circuits without the retry loop; or
- gate the Ironwood fetch on NU6.3 activation for the configured network, the same way the comment above the `if` already reasons about Orchard and NU5.

### LINKED ISSUE #2061: Stop tolerating missing Ironwood subtree roots once lightwalletd support is widespread

nesence-m:
## Summary

Against a lightwalletd that doesn't recognize the Ironwood pool (or whose backing node predates
NU6.3), `CompactBlockProcessor.getSubtreeRoots` was burning all `GET_SUBTREE_ROOTS_RETRIES`
attempts with exponential backoff on every spend-before-sync init and logging fatal-looking
errors. Two behavior changes fix this:

- **First-attempt tolerance of unknown-pool responses.** A new `isUnknownPoolFailure()` classifier
  recognizes the gRPC failure shapes a server returns when it doesn't know about a requested pool
  (UNKNOWN/2 + "unrecognized shielded protocol", INVALID_ARGUMENT/3 + "shielded protocol",
  UNIMPLEMENTED/12), and tolerates them: logged at info level, empty root list, no retry. The
  description-gated shapes apply uniformly across all three pools; the UNIMPLEMENTED tolerance
  exempts Sapling, since Sapling is the baseline pool every spend-before-sync-capable server must
  serve — a Sapling UNIMPLEMENTED means the server doesn't implement `GetSubtreeRoots` at all and
  stays a genuine failure.
- **Failures are no longer masked by Sapling success.** Previously, a genuine Orchard fetch
  failure was recorded but then silently discarded whenever the Sapling root list was non-empty,
  producing a false-positive `SpendBeforeSync` result. Fetch outcomes for all three pools are now
  combined explicitly: any `FailureConnection` wins, otherwise the first recorded failure wins,
  otherwise the result is derived from the Sapling root list as before.

This also removes the transitional Ironwood-only blanket failure tolerance tracked in #2061, since
Ironwood failures are now handled the same way as Sapling/Orchard failures.

Adds seven new `CompactBlockProcessorTest` tests covering: unknown-pool tolerance without retry
(all three failure shapes), a Sapling UNIMPLEMENTED surfacing as a genuine failure, an unmasked
genuine Orchard failure, a transient-then-success Sapling fetch, empty Sapling roots yielding
`Linear`, all-attempts-unavailable yielding `FailureConnection`, and connection-failure precedence
over other failures.

Closes #2058
Closes #2061

> **Note**

# Author

# Reviewer

### DIFF
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
@@ -1207,7 +1207,9 @@ class CompactBlockProcessor internal constructor(
          *   (2) with that description.
          * - An upgraded lightwalletd whose backing node is pre-NU6.3:
          *   `status.Errorf(codes.InvalidArgument, "GetSubtreeRoots: bad shielded protocol specifier error: ...")`,
-         *   which surfaces as INVALID_ARGUMENT (3).
+         *   which surfaces as INVALID_ARGUMENT (3) and is matched on the `"bad shielded protocol specifier"`
+         *   substring, not the more generic `"shielded protocol"`, so an INVALID_ARGUMENT failure for an
+         *   unrelated reason isn't mistaken for an unrecognized pool.
          * - UNIMPLEMENTED (12) is matched defensively for other indexers (e.g. Zaino) that haven't implemented
          *   the pool at all.
          *
@@ -1221,7 +1223,7 @@ class CompactBlockProcessor internal constructor(
                 }
 
                 GRPC_CODE_INVALID_ARGUMENT -> {
-                    description?.contains("shielded protocol", ignoreCase = true) == true
+                    description?.contains("bad shielded protocol specifier", ignoreCase = true) == true
                 }
 
                 GRPC_CODE_UNIMPLEMENTED -> {
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
@@ -1181,6 +1181,59 @@ class CompactBlockProcessor internal constructor(
         internal const val GET_SUBTREE_ROOTS_RETRIES = 3
 
         /**
+         * gRPC status code for UNKNOWN failures, i.e. the server returned a generic error with no matching
+         * gRPC status. sdk-lib has no `io.grpc.Status` on its compile classpath, so this mirrors
+         * `io.grpc.Status.Code.UNKNOWN`'s numeric value.
+         */
+        private const val GRPC_CODE_UNKNOWN = 2
+
+        /**
+         * gRPC status code for INVALID_ARGUMENT failures. Mirrors `io.grpc.Status.Code.INVALID_ARGUMENT`.
+         */
+        private const val GRPC_CODE_INVALID_ARGUMENT = 3
+
+        /**
+         * gRPC status code for UNIMPLEMENTED failures. Mirrors `io.grpc.Status.Code.UNIMPLEMENTED`.
+         */
+        private const val GRPC_CODE_UNIMPLEMENTED = 12
+
+        /**
+         * Classifies a subtree roots [Response.Failure] as coming from a server that doesn't recognize the
+         * requested shielded protocol, rather than as a genuine fetch failure. Covers the failure shapes
+         * empirically observed against lightwalletd:
+         *
+         * - A deployed pre-Ironwood lightwalletd (v0.4.x): its `GetSubtreeRoots` handler's pool switch returns
+         *   a plain Go error (`errors.New("unrecognized shielded protocol")`), which surfaces as gRPC UNKNOWN
+         *   (2) with that description.
+         * - An upgraded lightwalletd whose backing node is pre-NU6.3:
+         *   `status.Errorf(codes.InvalidArgument, "GetSubtreeRoots: bad shielded protocol specifier error: ...")`,
+         *   which surfaces as INVALID_ARGUMENT (3).
+         * - UNIMPLEMENTED (12) is matched defensively for other indexers (e.g. Zaino) that haven't implemented
+         *   the pool at all.
+         *
+         * UNKNOWN is gRPC's catch-all, so the code alone can't distinguish an unrecognized pool from any other
+         * unknown error; the description text is required to discriminate.
+         */
+        private fun Response.Failure<*>.isUnknownPoolFailure(): Boolean =
+            when (code) {
+                GRPC_CODE_UNKNOWN -> {
+                    description?.contains("unrecognized shielded protocol", ignoreCase = true) == true
+                }
+
+                GRPC_CODE_INVALID_ARGUMENT -> {
+                    description?.contains("shielded protocol", ignoreCase = true) == true
+                }
+
+                GRPC_CODE_UNIMPLEMENTED -> {
+                    true
+                }
+
+                else -> {
+                    false
+                }
+            }
+
+        /**
          * The theoretical maximum number of blocks in a reorg, due to other bottlenecks in the protocol design.
          */
         internal const val MAX_REORG_SIZE = 100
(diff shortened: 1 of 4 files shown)


nesence-m:
Rebased onto current `maint/v3.0.x` (past the 3.0.2 release merge, MOB-1667, and MOB-1717) and tightened the `INVALID_ARGUMENT` unknown-pool match from the generic "shielded protocol" substring to lightwalletd's exact "bad shielded protocol specifier" text per the review note, as a separate commit.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessor.kt
```diff
@@ -1180,6 +1180,61 @@ class CompactBlockProcessor internal constructor(
...
+         *   which surfaces as INVALID_ARGUMENT (3) and is matched on the `"bad shielded protocol specifier"`
+         *   substring, not the more generic `"shielded protocol"`, so an INVALID_ARGUMENT failure for an
+         *   unrelated reason isn't mistaken for an unrecognized pool.
+         * - UNIMPLEMENTED (12) is matched defensively for other indexers (e.g. Zaino) that haven't implemented
+         *   the pool at all.
+         *
+         * UNKNOWN is gRPC's catch-all, so the code alone can't distinguish an unrecognized pool from any other
+         * unknown error; the description text is required to discriminate.
+         */
+        private fun Response.Failure<*>.isUnknownPoolFailure(): Boolean =
```
noop-sk: **UNIMPLEMENTED tolerance also applies to Sapling**

`isUnknownPoolFailure()` is now invoked identically for all three pools via the shared `fetchSubtreeRoots` helper, and for `GRPC_CODE_UNIMPLEMENTED` it returns `true` unconditionally with no description check. That's the intended behavior for Orchard/Ironwood on indexers that haven't implemented those pools, but it also covers Sapling: a legacy lightwalletd that never implemented the `GetSubtreeRoots` RPC at all would return UNIMPLEMENTED for Sapling too, which previously surfaced as a genuine `OtherFailure` and now gets silently tolerated as "no Sapling roots", producing a false-positive `Linear` result instead of signaling that spend-before-sync isn't supported by that server. The new tests only exercise the unknown-pool failure shapes against Ironwood, not Sapling.

nesence-m: Good catch — fixed in 63efad75. `isUnknownPoolFailure` now takes the pool and exempts Sapling from the UNIMPLEMENTED tolerance: a Sapling UNIMPLEMENTED (server never implemented `GetSubtreeRoots` at all) is treated as a genuine failure again — retried and surfaced as `OtherFailure` — while Orchard/Ironwood keep the defensive tolerance. Added `sapling_unimplemented_is_a_genuine_failure_not_an_unknown_pool` covering the Sapling path (3 attempts, `OtherFailure` result). The two description-gated shapes remain uniform across pools since their server-side messages are inherently pool-mismatch responses. Also rebased onto current `maint/v3.0.x` past the Slipstream removal.

sdk-lib/src/test/java/cash/z/ecc/android/sdk/block/processor/CompactBlockProcessorTest.kt
```diff
@@ -432,6 +444,283 @@ class CompactBlockProcessorTest {
         }
     }
 
+    @Test
+    fun unknown_pool_failure_is_tolerated_without_retry() =
```
noop-sk, line 448: **Test only exercises the Ironwood path**

This test proves `UNKNOWN`(2)/`INVALID_ARGUMENT`(3)/`UNIMPLEMENTED`(12) tolerance via Ironwood only. Since the PR's stated behavior is that `UNKNOWN`/`INVALID_ARGUMENT` now apply uniformly across all three pools (only `UNIMPLEMENTED` gets the Sapling carve-out), there's no test pinning down "Sapling + code 2/3 → tolerated" alongside "Sapling + code 12 → genuine failure" side by side. Worth adding given this is exactly the axis the exemption fix needed to get right — a future edit to the `when` block could flip the wrong code without a test catching it.

noop-sk (dismissed):
Re-reviewed following the Sapling-exemption fix. Confirmed correct: `shieldedProtocol != ShieldedProtocolEnum.SAPLING` in the UNIMPLEMENTED branch (CompactBlockProcessor.kt) now returns `false` (genuine failure) only for Sapling, `true` for Orchard/Ironwood — exactly what I asked for, with a dedicated regression test (`sapling_unimplemented_is_a_genuine_failure_not_an_unknown_pool`). The redesigned `PoolFetchOutcome` sealed type also correctly fixes the underlying 'masked by Sapling success' bug this PR set out to solve. No Rust/JNI/PCZT changes, no supply-chain concerns, CI green, mergeable.

One non-blocking suggestion left inline on the test coverage — happy to approve as-is if you'd rather follow up separately.

noop-sk (approved):
Verified — the new commit adds exactly the two tests I asked for (sapling_unknown_code_is_tolerated_as_an_unknown_pool_failure, sapling_invalid_argument_is_tolerated_as_an_unknown_pool_failure), sitting right alongside the existing sapling_unimplemented_is_a_genuine_failure test — Sapling + code 2/3 tolerated, Sapling + code 12 genuine failure, all three pinned. CI green, mergeable. Ready.

Issue #2078, opened 2026-07-27
round_consensus_branch_id derives the branch from the caller's network, not the round's stored one


nuttycom:
`round_consensus_branch_id` (`backend-lib/src/main/rust/voting/delegation.rs:174-188`) loads the round's snapshot height with `voting::storage::queries::load_round_params` and then resolves the consensus branch using the `network` the JNI caller supplied:

```rust
voting::delegate::branch_id_for_height(network, snapshot_height)
```

`zcash_voting` persists the network alongside the round, and exposes it through `VotingDb::get_round_state`, which returns both the stored network and the snapshot height. `load_round_params_with_network` returns both in a single query as well. Either would let the branch be derived entirely from the round rather than partly from caller-supplied input.

## Severity: low, and it fails closed today

This is not currently exploitable. `build_governance_pczt` re-validates downstream: `validate_network_matches_round` rejects a network that disagrees with the stored round before `validate_consensus_branch_id_for_round` compares the branch id (upstream `storage/operations.rs:35-53`, `:567-575`). A caller passing the wrong network gets an error, not a wrong PCZT.

The objection is structural rather than behavioural: a value that the database already knows authoritatively is being taken from the caller instead, and the guarantee that this is safe lives in a different crate. Should upstream ever relax that validation, this becomes the place the mismatch slips through.

The crate's own migration notes also prefer `VotingDb::…` accessors over reaching into `storage::queries` directly.

## Suggested change

Derive both the snapshot height and the network from the round — `VotingDb::get_round_state`, or `load_round_params_with_network` if the lower-level call is preferred — and drop the reliance on the caller-supplied `network` for branch resolution. Note there is a pre-existing direct `load_round_params` call at `delegation.rs:514` that may be worth the same treatment.

Deferred from the review of #2075 as non-blocking.

---
Filed with Claude Code assistance.

PR #2180, opened 2026-08-17
Derive the governance PCZT branch id from the round, not the caller's network

dannywillems:
Closes [MOB-1561](https://linear.app/zodl/issue/MOB-1561/round-consensus-branch-id-derives-the-branch-from-the-callers-network) / #2078.

## The defect

`build_governance_pczt_for_bundle` resolved the consensus branch id as:

```rust
voting::delegate::branch_id_for_height(db.network, round_state.snapshot_height)
```

The snapshot height came from the round. The network did not: `VotingDbHandle::network` is whatever `network_id` the JNI caller passed to `openVotingDbNative`, stored once at open time and never revalidated against the database's contents. Two different networks were in play for one round:

| | source | authority |
|---|---|---|
| `db.network` | JNI `network_id` at open time | whatever the caller said |
| `round_state.network` | the `rounds` table, written at `init_round` | the database |

`zcash_voting` persists the network alongside the round and returns it on `RoundState`, so both inputs to the resolution can come from the round.

Note the issue text is stale in two ways: there is no `round_consensus_branch_id` function (the code is in the private helper `build_governance_pczt_for_bundle`), and the `snapshot_height` half was already fixed. Only the network argument remained.

## The change

Extract `consensus_branch_id_for_round`, which reads both `network` and `snapshot_height` off `get_round_state` and resolves the branch from those. One call site.

The handle's network is still **checked** against the round rather than simply dropped. A disagreement means the database was opened for the wrong network, so the voting hotkey, and hence `DelegationKeys::network`, is wrong for this round too. Upstream rejects that mismatch as well (`validate_network_matches_round`, consulted from `build_governance_pczt`), so this only turns a downstream error into a local one today. It is checked here so the invariant does not depend on a guarantee that lives in another crate, which was the issue's actual objection.

**Constraint for any follow-up:** `DelegationKeys::network` must keep deriving from the handle. The upstream check compares it against the stored round network, and it only has content because the two values arrive from independent sources. Deriving both from the round would make the comparison vacuously true and silently delete a fail-closed guard.

## Reachability

Not exploitable today, and it failed closed. The mismatch is reachable across handles rather than within one: `open_managed_db` rejects a differing network only while a handle for the same path is still live in-process, so a fresh handle opened over an existing database with another `network_id` is not caught.

## Not changed

The other direct `storage::queries::load_round_params` call, in `require_witnesses_match_bundle`, is left as is. It reads only `nc_root`, no network is involved, and the migration note's preference for `VotingDb::...` accessors cannot be honoured there because `VotingDb` exposes no method returning `VotingRoundParams`.

## Tests

Two added, both in `voting::delegation::tests`:

- `consensus_branch_id_for_round_resolves_against_the_rounds_stored_network` asserts the result equals the Regtest-derived branch, and via `assert_ne!` that the mainnet-derived value at the same height differs. That inequality is what makes the second test meaningful.
- `consensus_branch_id_for_round_rejects_a_handle_opened_for_another_network` is the regression test. Under the old code this returned a mainnet branch id with no error, so `unwrap_err()` would have panicked.

## Verification

- `cargo test --lib voting::` green (14/14)
- `cargo fmt -- --check` clean
- `cargo clippy --all-targets -- -D warnings` clean

No changelog entry: the Unreleased section is user-facing API deltas, and this is invisible to callers on every supported path. Happy to add one if preferred.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,4 +104,4 @@ backup-dbs/
 
 # Design docs, implementation plans, and other working notes. These are
 # conversation artifacts rather than repository history.
-.plans/
+.plans
--- a/backend-lib/src/main/rust/voting/delegation.rs
+++ b/backend-lib/src/main/rust/voting/delegation.rs
@@ -164,15 +164,7 @@ fn build_governance_pczt_for_bundle(
     )
     .map_err(|e| anyhow!("DelegationKeys::with_voting_hotkey: {}", e))?;
 
-    // build_governance_pczt now validates the branch id against the round's
-    // own snapshot_height (see lwd::branch_id_for_height), so it must be
-    // resolved per round instead of the old hardcoded Nu6 constant.
-    let round_state = db
-        .get_round_state(round_id)
-        .map_err(|e| anyhow!("get_round_state: {}", e))?;
-    let consensus_branch_id =
-        voting::delegate::branch_id_for_height(db.network, round_state.snapshot_height)
-            .map_err(|e| anyhow!("branch_id_for_height: {}", e))?;
+    let consensus_branch_id = consensus_branch_id_for_round(db, round_id, db.network)?;
 
     let pczt = db
         .build_governance_pczt(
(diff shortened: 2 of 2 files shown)


dannywillems (commented):
Self-utACK

PR #2183, opened 2026-08-18
Remove accidentally-merged voting from maint/v3.0.x

noop-sk:
## Summary

`zcash_voting` was accidentally reenabled on `maint/v3.0.x` via #2157 and
#2161. `zcash_voting` work belongs on `main`; the 3.0.x maintenance line
was never meant to carry it. Root cause (per the team thread): the real
need behind #2157 was verifying the CHP delegation fix on a real device
against the 3.0.x-based app, which never required actually merging to
`maint/v3.0.x`.

**Surgical, dependency-only fix** (not a full revert of #2157/#2161):
reverting those merges wholesale would also throw away the legitimate
`tx1_effects`/`DelegationPhase` fixes they carried and unnecessarily
diverge this branch's voting source from `main`'s own line. Instead:

- Comment out `zcash_voting`/`pasta_curves`/`ff` in `Cargo.toml`, drop
  `orchard`'s `unstable-voting-circuits` feature — matches the
  "to re-enable, uncomment these three" comment that was already in the
  file (and that the accidental re-enable had drifted out of sync with)
- `Cargo.lock` regenerated (`cargo check`): drops `zcash_voting`,
  `voting-circuits`, `vote-commitment-tree(-client)`, `pir-client`,
  `pir-types`, `valar-spiral-rs`, `valar-ypir`, `imt-tree`,
  `hyper-rustls` — 10 crates fully out of the resolved dependency graph.
  This was the actual binary-size/build-time concern raised in the
  thread, and it's fully addressed here — same win as a full revert.
- Drop the `--cfg zcash_voting` RUSTFLAGS thread-through in
  `build.gradle.kts` — load-bearing, not cosmetic: left in place, every
  Gradle-driven `cargoBuild` would try to compile `mod voting` against
  the now-absent crate and fail.

`mod voting` in `lib.rs` stays behind `#[cfg(zcash_voting)]`, so none of
`voting/delegation.rs`, `voting/helpers.rs`, `voting/rounds.rs`, or the
Kotlin JNI/API surface (`VotingRustBackend.kt`, `JniConstants.kt`,
`JniVotingModels.kt`, `TypesafeVotingBackend*.kt`) needed to change —
they're simply not compiled in this default config, same as before
MOB-1545 reintroduced the module. Only exception: `VotingRustBackendTest`
actually exercises the native symbols at runtime, so it needs `@Ignore`
again now that they're not linked; everything else in that file is
untouched.

**Deliberately not touched:** `.github/workflows/pull-request.yml`'s
60min CI timeout bump (and its now-stale "this branch re-enables
shielded voting" comment). It's inaccurate now but harmless — a looser
timeout than needed, not a failure — and reverting it isn't required
for the dependency removal to work. Left alone to keep this diff to
exactly what the fix needs; happy to clean it up separately if wanted.

**Net effect for the app: none.** `VotingCryptoClient.kt`'s existing
`pirDepth`/`pirTier0Layers`/`pirTier1Layers` call-site shim (added in
zodl-android commit `227ee1e09` to match #2157's SDK signature) keeps
compiling as-is, since that signature didn't change. No companion app
PR is needed — verified with a full
`:app:assembleZcashmainnetInternalDebug` build against this branch via
an included-build composite (app on `maint/v3.9.x`, SDK on this
branch) — **BUILD SUCCESSFUL**, APK produced.

## ⚠️ Forward-merge caution

`#2157`/`#2161` never landed on `main` — `main` has its own, further-
developed `zcash_voting` line. When this branch is eventually cascaded
forward (`maint/v3.0.x` → newer maint → `main`), the diff here is small
and dependency-only (`Cargo.toml`/`Cargo.lock`/`build.gradle.kts`/one
`@Ignore`), so the collision surface with `main`'s active voting work is
much smaller than a full revert would have been — but still worth a
manual look at those files during that merge rather than trusting it
blind.

## Test plan

- [x] `cargo check` / `cargo check --all-targets` on `backend-lib`
      (default, non-voting config) — clean
- [x] Confirmed via `Cargo.lock` diff that all 10 voting-related crates
      drop out of the resolved graph
- [x] No production Kotlin/Rust voting API/logic files touched
      (`git diff --stat` scoped to `backend-lib/src/main`/`sdk-lib` is
      empty)
- [x] Full app build (`:app:assembleZcashmainnetInternalDebug`) against
      this branch via composite build — BUILD SUCCESSFUL
- [ ] CI green on this PR

### DIFF
--- a/backend-lib/Cargo.lock
+++ b/backend-lib/Cargo.lock
@@ -2311,22 +2311,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "hyper-rustls"
-version = "0.27.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "33ca68d021ef39cf6463ab54c1d0f5daf03377b70561305bb89a8f83aab66e0f"
-dependencies = [
- "http",
- "hyper",
- "hyper-util",
- "rustls",
- "tokio",
- "tokio-rustls",
- "tower-service",
- "webpki-roots",
-]
-
-[[package]]
 name = "hyper-timeout"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2388,20 +2372,6 @@ name = "ident_case"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b9e0384b61958566e926dc50660321d12159025e767c18e043daf26b70104c39"
-
-[[package]]
-name = "imt-tree"
-version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0aad3e3ab5a46f4a08d00f5525b86cd0d5204cba64ba68f46e20a590e1808aec"
-dependencies = [
- "anyhow",
- "ff",
- "halo2_gadgets",
- "hex",
- "pasta_curves",
- "rayon",
-]
 
 [[package]]
 name = "incrementalmerkletree"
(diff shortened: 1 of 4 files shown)


nuttycom (approved):
utACK

PR #2184, opened 2026-08-18
fix(sync): stop Tor circuit-build noise from skewing fastest-server ranking

noop-sk:
## Summary
- Fastest-server candidate ranking averaged a cold circuit-build measurement (`getServerInfoDuration`) together with a warm round-trip measurement (`getLatestBlockHeightDuration`), baking Tor circuit-build noise into the ranking metric. Live-verified as the dominant cause of an isSpendable-oscillation symptom (PRO-389): 4/4 consecutive app launches on a 7-candidate mainnet config each picked a different "fastest" server, forcing a real endpoint switch — and full Synchronizer teardown/rebuild — almost every launch. Ranking on the warm measurement alone: 3/4 launches after the first settled as a no-op.
- Also share one isolated Tor circuit across all fastest-server candidates in a single `FastestServerFetcher.invoke()` call instead of each candidate building its own — an independently-justified fix for wasted concurrent circuit-build churn (up to 7 simultaneous circuit builds per evaluation on mainnet).

This is a first step addressing the Tor-enabled sync instability reported in PRO-389/MOB-1726/MOB-1727 — it fixes the noisy ranking that was causing unnecessary real server switches. It does not address every symptom in that investigation.

## Test plan
- [x] Compiles (`:sdk-lib:compileDebugKotlin`)
- [x] Live A/B tested on-device (isSpendable-transition timing, Tor + automatic server enabled): before fix, 4/4 launches triggered a real switch; after fix, 3/4 settled as a no-op

### DIFF
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/FastestServerFetcher.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/FastestServerFetcher.kt
@@ -42,14 +42,22 @@ internal class FastestServerFetcher(
         flow {
             emit(FastestServersResult.Measuring)
 
-            val serversByRpcMeanLatency =
+            // MOB-1728: candidates share no Tor circuit -- each derives its own via
+            // walletClientFactory.create(endpoint), same as before this file was touched. An
+            // earlier version of this fix shared one isolated circuit across every candidate, on
+            // the theory that too many concurrent Tor connections were the problem; that turned
+            // out not to be it (see ValidateServerResult.rankingLatency's kdoc for what actually
+            // was: the noisy ranking metric), the sharing was reverted, and parallelMapNotNull
+            // below still runs every candidate concurrently regardless -- N independent circuits
+            // built at once, same as it would have been without ever touching this file at all.
+            val serversByRankingLatency =
                 servers
                     .parallelMapNotNull {
                         validateServerEndpointAndMeasure(it)
                     }.sortedBy {
-                        it.meanDuration
+                        it.rankingLatency
                     }.mapIndexedNotNull { index, result ->
-                        if (index <= K - 1 || result.meanDuration <= LATENCY_THRESHOLD) {
+                        if (index <= K - 1 || result.rankingLatency <= LATENCY_THRESHOLD) {
                             Twig.debug { "Fastest Server: '${result.endpoint}' VALIDATED by SORTING by RPC latency" }
                             result
                         } else {
@@ -59,13 +67,13 @@ internal class FastestServerFetcher(
                     }
 
             Twig.debug {
-                "Fastest Server: '${serversByRpcMeanLatency.map { it.endpoint }}' VALIDATED by MEASURING RPC latency"
+                "Fastest Server: '${serversByRankingLatency.map { it.endpoint }}' VALIDATED by MEASURING RPC latency"
             }
 
-            emit(FastestServersResult.Validating(serversByRpcMeanLatency.map { it.endpoint }.take(K)))
+            emit(FastestServersResult.Validating(serversByRankingLatency.map { it.endpoint }.take(K)))
 
             val serversByGetBlockRangeTimeout =
-                serversByRpcMeanLatency
+                serversByRankingLatency
                     .asFlow()
                     .mapNotNull { result ->
                         result.use {
@@ -115,33 +123,34 @@ internal class FastestServerFetcher(
             }
         }
 
-        val lightWalletClient = kotlin.runCatching { walletClientFactory.create(endpoint) }.getOrNull() ?: return null
+        val lightWalletClient =
+            kotlin.runCatching { walletClientFactory.create(endpoint) }.getOrNull()
+                ?: return null
 
-        val remoteInfo: LightWalletEndpointInfoUnsafe?
-        val getServerInfoDuration =
-            measureTime {
-                // 5 seconds timeout in case server is very unresponsive
-                remoteInfo =
-                    withTimeoutOrNull(5.seconds) {
-                        when (
-                            val response =
-                                lightWalletClient.getServerInfo(
-                                    sdkFlags ifTor
-                                        ServiceMode.Group(
-                                            "validateServerEndpointAndMeasure(${endpoint.host}:${endpoint.port})"
-                                        )
+        // Not timed: MOB-1728 ranks candidates on getLatestBlockHeightDuration alone (see
+        // ValidateServerResult.rankingLatency's kdoc for why), so this call's own duration isn't
+        // needed -- only remoteInfo, for the checks below.
+        // 5 seconds timeout in case server is very unresponsive
+        val remoteInfo: LightWalletEndpointInfoUnsafe? =
+            withTimeoutOrNull(5.seconds) {
+                when (
+                    val response =
+                        lightWalletClient.getServerInfo(
+                            sdkFlags ifTor
+                                ServiceMode.Group(
+                                    "validateServerEndpointAndMeasure(${endpoint.host}:${endpoint.port})"
                                 )
-                        ) {
-                            is Response.Success -> {
-                                response.result
-                            }
-
-                            is Response.Failure -> {
-                                logRuledOut("getServerInfo failed", response.toThrowable())
-                                null
-                            }
-                        }
+                        )
+                ) {
+                    is Response.Success -> {
+                        response.result
                     }
+
+                    is Response.Failure -> {
+                        logRuledOut("getServerInfo failed", response.toThrowable())
+                        null
+                    }
+                }
             }
 
         if (remoteInfo == null) {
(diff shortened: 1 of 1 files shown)


noop-sk:
Good catch, confirmed — added \`88f546e1\`.

\`CombinedWalletClientImpl.dispose()\` unconditionally disposed its \`isolatedTorClient\`, and \`LazyTorClient.dispose()\` is one-shot/permanent by design. Since every candidate in \`FastestServerFetcher\` shares the same instance and runs concurrently via \`parallelMapNotNull\`, the first candidate to finish and \`dispose()\` — a ruled-out one, the common/expected outcome — permanently killed the shared circuit for every sibling still probing, sending them all into \`UninitializedTorClientException\` and getting them ruled out too. Confirmed by reading the dispose paths directly, not just from the report.

Fix: \`CombinedWalletClientImpl\` now takes an \`ownsIsolatedTorClient\` flag (default \`true\`, matching every existing call site that derives its own dedicated client). \`FastestServerFetcher\` passes \`false\` for its shared client and takes ownership of disposing it itself, exactly once, in a \`finally\` after every candidate (validation + getBlockRange stages) has finished with it.

On the missing unit tests for \`FastestServerFetcher\`/\`WalletClientFactory\` — agreed that's a real, pre-existing gap (confirmed no test file exists for either today), not something introduced by this PR. Leaving that out of scope here to keep this PR focused on the lifecycle fix; happy to open a follow-up if that's preferred instead.

noop-sk:
Ran an independent review pass (Fable) over the full diff and verified its findings myself before acting — added \`f61f1625\`:

**Fixed:**
- CI-blocking: \`detektAll\` (the actual repo-wide task, not \`detekt\`) failed on 5 \`MaxLineLength\` violations — pre-existing \`Twig.debug\` lines pushed one indent level deeper by the new \`try\`/\`finally\`. Reflowed; also had to suppress a resulting \`LongMethod\` on \`invoke()\` (same pattern already used on \`validateServerEndpointAndMeasure\` in this file).
- The \`finally { sharedIsolatedTorClient?.dispose() }\` from the last commit silently no-ops on flow cancellation (e.g. navigating away mid-probe): \`LazyTorClient.dispose()\` suspends via \`TorClient.dispose()\`'s \`withContext(Dispatchers.IO)\`, and \`withContext\` in an already-cancelled coroutine throws immediately without running its block — skipping the native Tor runtime teardown this \`finally\` exists to guarantee. Wrapped in \`withContext(NonCancellable)\`.
- Comments in this file and \`CombinedWalletClientImpl\` asserted the per-candidate probes run *concurrently* ("N concurrent circuit builds in one burst", "under real contention"). Verified they don't: \`parallelMapNotNull\`'s per-element \`coroutineScope { async { ... } }\` blocks until that element finishes before the next one starts, despite the name — this is pre-existing, unchanged by this PR. The bugs/fixes here (noisy ranking, shared-circuit ownership) hold regardless, but corrected the wording — it's N sequential circuit-build costs adding up, not concurrent contention.
- Now that ranking no longer averages \`getServerInfoDuration\` in: dropped the now-dead duration measurement, renamed \`meanDuration\`/\`serversByRpcMeanLatency\` (no longer means of anything) to \`rankingLatency\`/\`serversByRankingLatency\`.
- Softened the "nothing is lost" privacy framing on sharing one isolated circuit across candidates — true for the candidates themselves, but an observing exit does see one circuit toward all N hosts instead of N separate ones. Noted as an accepted tradeoff rather than a non-issue.

**Deliberately left as follow-ups (pre-existing, not introduced by this PR, out of scope here):**
- \`parallelMapNotNull\` is misleadingly named — it's sequential, not parallel. Actually fixing it would be a real behavioral change (probes would start genuinely overlapping against the shared Tor client) that deserves its own PR and re-testing, not folded into a lifecycle-fix PR.
- Two pre-existing \`ValidateServerResult\` leaks in the same function (candidates ruled out by the latency-sort filter, and any left over after \`take(K)\` short-circuits stage 2) never call \`dispose()\`. Same shape on \`maint/v3.0.x\` today; happy to file a follow-up ticket if useful.

Also independently confirmed the \`ownsIsolatedTorClient\` fix itself is complete — traced every \`dispose()\` call path and every \`create()\` call site again from scratch, no gaps found.

noop-sk:
Actually fixed \`parallelMapNotNull\` to be genuinely concurrent (\`64e140fa\`), addressing the follow-up flagged in the previous comment.

**Fix:** moved \`coroutineScope\` to wrap the whole \`map\`, not create one per element — previously each candidate's own \`coroutineScope { async { ... } }\` suspended on that one child before returning, so \`map\` never started the next candidate until the previous one's \`block()\` fully completed, despite the function's name.

**Live verification (6 trials total):**
- Confirmed via logs: overlapping Tor connection attempts to multiple candidate hosts now interleave (previously strictly sequential) — one trial completed the full probe in ~1s vs. several seconds per candidate before.
- No deadlock or crash across any trial.
- Two of six trials had some/all candidates fail (\`CircTimeout\`, one \`UninitializedTorClientException\`/hostname-lookup error). Ran a controlled A/B for this specifically — stashed this exact commit, rebuilt the pre-parallelization (sequential) version, ran 3 more trials on the same device/session: the *same* \`CircTimeout\` errors show up there too, at a similar rate, including one sequential run that hadn't even finished probing after 25s. So this is a pre-existing characteristic of the emulator's Tor path, not a regression from parallelizing — confirmed empirically rather than assumed.

Also updated \`invoke()\`'s top comment (now accurate again: N candidates really do build Tor circuits concurrently, which is what actually motivates the shared-circuit fix from the earlier commits here) and \`parallelMapNotNull\`'s own kdoc.

noop-sk:
Reverted the shared-Tor-circuit fix entirely (\`3f61028a\`) after re-examining why it existed.

Its own original commit message already flagged it as unconfirmed: *"Live A/B tested... no measurable difference on its own — the isSpendable-oscillation symptom is dominated by a separate mechanism [the ranking fix]."* It was kept anyway as a speculative, "independently-justified" efficiency improvement — but today it cost 3 follow-up bugfixes (ownership/disposal, cancellation-safety, then genuine parallelization) to get correct, for a benefit that was never actually confirmed to matter for the reported issue. The real, live-verified root cause is the ranking metric baking in cold-circuit-build noise (previous commit) — nothing to do with how many Tor circuits get built or whether they're shared.

\`FastestServerFetcher.kt\`/\`WalletClientFactory.kt\`/\`CombinedWalletClientImpl.kt\` are now byte-for-byte identical to \`maint/v3.0.x\` except for the ranking fix — confirmed via diff.

Kept \`parallelMapNotNull\`'s concurrency fix from two commits ago — candidates now genuinely probe in parallel, each still building its own independent circuit (same as pre-MOB-1728, just concurrent now instead of accidentally sequential). Live-tested 3x on-device after this revert: **7/7 candidates validated in every single run**, full probe completing in ~2.5-2.8s — actually more consistent than the shared-circuit parallel version from earlier today (which had one run return zero candidates). So dropping the sharing didn't just remove risk, it measurably improved reliability here.

PR is now much smaller in scope: the ranking fix (the actual confirmed root-cause fix) plus a genuine, independently-useful concurrency correctness fix to \`parallelMapNotNull\`.

nesence-m (commented):
Verified the premise against the actual client code before judging the fix: in `CombinedWalletClientImpl`, `ServiceMode.Group` routes through `getOrPut(serviceMode) { createTorWalletClient() }`, and both probes use the identical group key — so `getServerInfo` builds the circuit (cold, variable) and `getLatestBlockHeight` reuses it (warm). Ranking on `rankingLatency = getLatestBlockHeightDuration` alone genuinely isolates the circuit-build cost, `getServerInfo` still runs first as the warm-up (the first-request bias hasn't just moved), and non-Tor mode arguably improves too. Core fix is correct; CI green; nothing touches the JNI/Cargo surface.

Concerns worth a second look before approval:

1. **`FastestServerFetcher.kt:45-52` — PR description contradicts the code.** The body's second bullet claims the PR "share[s] one isolated Tor circuit across all fastest-server candidates", but the comment at these lines says sharing was tried, found not to be the cause, and reverted — the diff contains no sharing. The merge record would misdescribe what ships; please update the body.

2. **`FastestServerFetcher.kt:252-255` — the `parallelMapNotNull` rewrite is a second, real behavior change, undersold.** The old `map { coroutineScope { async { block(it) } } }` was de facto sequential (`coroutineScope` suspends until its child completes) — verified against the pre-PR body. The new form is genuinely concurrent: wall-time drops from sum to max (good), but all candidates now measure their warm round-trip while up to 6 other circuits build and probe concurrently over the same device link and one shared base Tor runtime — cross-candidate contention is a new noise source in the very metric now ranked on. Your live A/B ran on this build so it's acceptance-tested, but this deserves explicit mention in the PR body/CHANGELOG as its own change (the comment at 51-52 currently describes the intended-but-never-shipped parallel behavior as the status quo).

3. **`FastestServerFetcher.kt:183-210` — no timeout on the ranked call (recommend follow-up).** `getServerInfo` is 5s-capped; `getLatestBlockHeight` has none, and the ranking now rests solely on it — one hung warm circuit blocks `awaitAll` indefinitely and the whole fastest-server flow stalls with no `Done` emission. A `withTimeoutOrNull` ruling the candidate out on expiry would bound it. Pre-existing gap, but the PR raises its weight.

4. **`FastestServerFetcher.kt:60,296` — `LATENCY_THRESHOLD` not recalibrated.** 300ms was tuned against `mean(cold, warm)`; warm-only values are systematically smaller, so the beyond-K cutoff is effectively loosened. Likely fine (stage 2 still `.take(K)`s), but worth being a conscious decision.

5. Nit (pre-existing): `mapIndexedNotNull` drop paths and `.take(K)` cancellation leave `ValidateServerResult` clients undisposed — heavier per leak under Tor; a follow-up dispose pass would help.

6. No tests (the sequential-vs-parallel helper is cheaply unit-testable with fake suspending blocks) and no `CHANGELOG.md` entry despite consumer-observable server-selection behavior changing — suggest a `### Fixed` entry.

PR #2192, opened 2026-08-21
MOB-1750: lower Orchard migration dust threshold to 0.0001 ZEC

noop-sk:
## Summary
- `MIGRATION_DUST_THRESHOLD_ZATOSHI` 100,000 -> 10,000 zatoshi (0.001 -> 0.0001 ZEC), the
  lower bound app-side MOB-1750 uses to detect a sub-0.01 ZEC Orchard residue for
  restored/self-migrated wallets.
- Companion app-side PR: zodl-inc/zodl-android (bugfix/MOB-1750 -> maint/v3.10.x).

## Test plan
- [x] `./gradlew ktlint detektAll` clean
- [x] `./gradlew :sdk-lib:test` — full sdk-lib unit test suite passes, including the
      `migrationDustThresholdZatoshi` passthrough test with the updated value
- [x] Verified no other stale `100_000`/`0.001 ZEC` reference remains anywhere in
      `backend-lib`/`sdk-lib` tied to migration dust semantics (adversarial review pass)

### DIFF
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/MigrationRustBackend.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/MigrationRustBackend.kt
@@ -480,12 +480,12 @@ class MigrationRustBackend private constructor() {
 
     /**
      * The zatoshi value below which a leftover post-migration Orchard balance is treated as dust
-     * rather than a residual worth migrating in its own transfer — see
-     * `MIGRATION_DUST_THRESHOLD_ZATOSHI` in `migration.rs`. A fixed protocol-level constant, not
-     * derived from any wallet/account state — [SdkDispatchers.CPU_BOUND] (2026-08-07 blocking-
-     * without-reason audit), not [SdkDispatchers.DATABASE_IO]: this never touches a database, so
-     * routing it through the SDK's single shared DB-IO thread bought nothing but contention with
-     * genuine migration DB reads/writes.
+     * rather than a residual worth migrating in its own transfer — `2 * MARGINAL_FEE`, see
+     * `migration_dust_threshold_zatoshi()` in `migration.rs`. Not derived from any wallet/account
+     * state — [SdkDispatchers.CPU_BOUND] (2026-08-07 blocking-without-reason audit), not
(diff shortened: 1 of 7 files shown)


nesence-m (dismissed):
Minimal and fully self-consistent: the Rust constant, both KDoc mentions, and every test value move together, and a sweep of backend-lib/sdk-lib confirms no stale 100_000 dust reference remains. Companion app PR zodl-inc/zodl-android#2458 reads this at runtime via migrationDustThresholdZatoshi(), so it needs an SDK release from this line — noted over there.

noop-sk:
Added a follow-up commit per the team Slack discussion on the dust/migratable-total calculation (Kris Nuttycombe's algorithm): `migratableOrchardTotal()`, a new per-note total (`sum(value - MARGINAL_FEE)` over notes whose value exceeds `MARGINAL_FEE`), which the app now compares against `migrationDustThresholdZatoshi()` instead of the raw aggregate Orchard balance.

`select_unspent_notes` already applies the `value > MARGINAL_FEE` filter at the SQL layer, so this only sums what it returns — no new engine/SQL logic. `MIGRATION_DUST_THRESHOLD_ZATOSHI` is unchanged (still `2 * MARGINAL_FEE`).

Flagging since this PR was already approved — happy to answer any questions on the new addition.

noop-sk:
One more small follow-up commit: MIGRATION_DUST_THRESHOLD_ZATOSHI was still a bare hardcoded 10_000 literal that only happened to numerically match 2 * MARGINAL_FEE. Since migratableOrchardTotal() already compares against exactly that value, the threshold now computes 2 * MARGINAL_FEE directly (via a plain fn — Zatoshis::into_u64() isn't const, so it can't stay a compile-time const) instead of duplicating the number. It now tracks ZIP-317's marginal fee automatically instead of silently drifting if that ever changes.

nesence-m (dismissed):

noop-sk:
Fixed a CI failure: \`test_android_modules_unit\` failed because \`FakeTypesafeMigrationBackend\` (a hand-rolled test double in \`OrchardMigrationSdkImplTest.kt\` implementing the full \`TypesafeMigrationBackend\` interface directly) was missing an override for the new \`migratableOrchardTotal()\` method. Added the same \`error("Unused")\` stub the other unexercised methods there use.

nesence-m (approved):

PR #2193, opened 2026-08-21
[MOB-1760] Size Keystone migration runs by signing-round capacity, not note count

noop-sk:
## Summary

A Keystone-signed migration run was planned via the crate's plain 50-note default (`engine::plan_migration`), independent of the 96-action-per-round QR-scanning budget a Keystone device can actually sign in one interaction. A run's action count follows the wallet's fragmentation, not its note count, so a "single" run could still need several manual QR-scan rounds to sign.

- `Backend::is_keystone()` resolves from the account's `key_source` (matching zashi-android's `AccountDataSource.importKeystoneAccount` tag, `KEYSTONE_KEYSOURCE`).
- `compute_plan`/`estimateMigrationRunCountNative` now share one `run_sizing_for` helper, so the previewed run count can never drift from what actually gets planned:
  - Keystone accounts → `RunSizing::Signer(RunSigningCapacity::KEYSTONE)` — a run is sized to fit one 96-action signing round.
  - zodl's own in-process signer → `RunSizing::Notes(ZODL_MAX_PREPARED_NOTES_PER_RUN)` (200, raised from the crate's Keystone-oriented 50-note default — no per-round cost to bound, so a larger cap just means fewer background sync/broadcast campaigns for the same wallet).
- `KEYSTONE_KEY_SOURCE` is now a named constant in `migration_engine.rs` with a doc comment flagging the cross-repo string-matching risk (no compiler-enforced link to the app's own constant) for future maintainers.

Adapted from the reference spike (`spike/keystone-signer-capacity-sizing`, `331bb2e8f`) against the now-stable `zcash_pool_migration` 0.1.0 crate — no `[patch.crates-io]` pin needed, every API used (`plan_migration_sized_with`, `estimate_migration_runs_sized_with`, `RunSizing`, `RunSigningCapacity::KEYSTONE`) ships in the published crate.

Screenshot 2026-08-21 at 13 44 26

## Test plan

- [x] `cargo test --all-features` — 98 passed, 0 failed, 15 ignored (pre-existing `#[ignore = "requires MIGRATION_TEST_WALLET_DB"]` live-wallet tests; ran all 15 manually against a real on-device wallet DB pulled during this investigation, including the new `backend_is_keystone_reflects_the_accounts_key_source`).
- [x] `cargo clippy --all-features -- -D warnings` — clean.
- [x] `cargo fmt -- --check` — clean.
- [x] `./gradlew ktlint detektAll` — clean (no Kotlin files touched).
- [x] Two new unit tests pin the `run_sizing_for` mapping directly (`keystone_accounts_use_signer_capacity`, `non_keystone_accounts_use_the_zodl_note_cap`) — no wallet fixture needed, targets the exact "swap KEYSTONE for a different capacity" regression class.
- [x] Live-verified end to end on a real Keystone-imported mainnet account (DIM build): pulled the on-device wallet DB, ran `compute_plan` directly against it, confirmed `signing_round_count(KEYSTONE) == 1` for a fresh plan (previously multi-round under the old note-cap-only sizing on the same account's earlier, pre-fix committed migration).

### DIFF
--- a/backend-lib/src/main/rust/migration.rs
+++ b/backend-lib/src/main/rust/migration.rs
@@ -43,6 +43,7 @@ use jni::{
 use prost::Message;
 use rand::rngs::OsRng;
 use rusqlite::Connection;
+use std::num::NonZeroUsize;
 use std::ptr;
 
 use zcash_client_backend::data_api::wallet::input_selection::LockFilter;
@@ -63,12 +64,13 @@ use zcash_pool_migration::{
     engine::{
         self, MigrationCrypto, MigrationPlan, MigrationState, MigrationTransaction,
         MigrationTransferId, MigrationTxKind, MigrationTxState, PoolMigrationRead,
-        PoolMigrationWrite, ProveError,
+        PoolMigrationWrite, ProveError, RunSizing,
     },
-    preparation::{PrepInput, PreparationPlan},
+    preparation::{PrepInput, PreparationPlan, default_portfolio},
     satisfiability::{
         AdvanceConfig, DuenessTargets, ReorgSettleDepth, ReplanThreshold, advance_migration,
     },
+    signing_rounds::RunSigningCapacity,
     state::{AdvanceStep, NextAction, StepKind},
 };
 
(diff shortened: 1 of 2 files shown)


backend-lib/src/main/rust/migration.rs
```diff
@@ -112,6 +114,63 @@ const JNI_KEYSTONE_BATCH_SIGNED_PCZTS: &str =
 /// or account state, so it needs no database access to read.
 pub const MIGRATION_DUST_THRESHOLD_ZATOSHI: u64 = 100_000;
 
+/// The per-run prepared-note cap for zodl's own in-process signer, well above the crate's
+/// [`MIGRATION_MAX_PREPARED_NOTES_PER_RUN`](zcash_pool_migration::denomination::MIGRATION_MAX_PREPARED_NOTES_PER_RUN)
+/// default of 50. That default exists to bound a run's transaction/proving cost for a signer that
+/// must sign it within a per-round action budget (Keystone); zodl's own signer has no such round
+/// to bound, so a larger cap just means fewer runs (and so fewer background sync/broadcast
+/// campaigns) for the same wallet, at the cost of a longer single planning/proving pass.
+const ZODL_MAX_PREPARED_NOTES_PER_RUN: NonZeroUsize = match NonZeroUsize::new(200) {
```
nesence-m, line 123: **Hot-wallet cap raise validation**

Beyond the Keystone fix, this raises the non-Keystone cap from the crate's 50-note default to 200 — a behavioral change for every hot-wallet account (longer single planning/proving pass, smaller run-count estimates surfacing in the UI). Was 200 exercised against a real device for planning/proving time and memory, and does a larger run still fit the background sync/broadcast campaign budget? The live end-to-end verification described in the PR covered the Keystone path.

nesence-m (approved):
The `run_sizing_for` single-knob design does exactly what it promises — preview and plan can no longer drift — and I verified the `"keystone"` string against the app side (`AccountDataSource.kt`'s `KEYSTONE_KEYSOURCE`): it matches, and the case-insensitive comparison here is a safe superset of the app's own checks. One question inline about the bundled hot-wallet cap raise. Small logistics note: #2192 edits the adjacent dust-threshold lines of `migration.rs` on the same base, so whichever lands second needs a trivial rebase.

backend-lib/src/main/rust/migration.rs
noop-sk: Fair question — no, 200 wasn't load-tested against a real device for proving time/memory. It's deliberately just a "high enough" ceiling, not a validated number: for zodl's own in-process signer there's no per-round signing cost to bound (unlike Keystone), so the goal is simply to avoid needing multiple migration runs at all for a typical wallet — 200 is picked generously for that, not because 200 preparations in one pass was measured safe. Agreed this deserves a real load test before we'd want to push it higher or rely on it for a large/heavily-fragmented wallet; happy to file a follow-up for that rather than block this PR on it, since the Keystone fix (this PR's actual subject) doesn't depend on the exact zodl-side number.

PR #2195, opened 2026-08-21
MOB-1751: Add Rust-backed cross-chain payment URI parser

zcash-harry:
## Summary

- expose one PaymentUriParser API for Bitcoin, Ethereum, Litecoin, and Solana
- decode the shared versioned librustzcash result for every supported scheme
- preserve the existing EIP-681 result shape and the upstream eip681 parser logic
- preserve exact decimal amounts as strings
- keep the JNI bridge to a single shared parse_to_json call

## Dependency

- https://github.com/zcash/librustzcash/pull/2989

## Testing

- ./scripts/ci-local.sh quick passed
- managed Pixel 2 API 36 sdk-lib instrumentation suite passed: 129 tests executed
- PaymentUriParser instrumentation verifies Ethereum recipient and atomic-value output from the common path

AI disclosure: Implemented with OpenAI Codex assistance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `PaymentUriParser`, a Rust-backed API that validates Bitcoin, Ethereum, Litecoin, and Solana
+  payment URIs and returns typed payment request models without floating-point amount conversion.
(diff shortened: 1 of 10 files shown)


zcash-harry:
Closing this in favor of #2196 — same commits, but pushed directly to a branch on this repo instead of a personal fork. No functional reason for the fork hop; I have direct push access here. Sorry for the churn — the review discussion here is still fully readable, just continuing on the new PR going forward.

PR #2196, opened 2026-08-22
MOB-1751: Add Rust-backed cross-chain payment URI parser

zcash-harry:
## Summary

- expose one PaymentUriParser API for Bitcoin, Ethereum, Litecoin, and Solana
- decode the shared versioned librustzcash result for every supported scheme
- preserve the existing EIP-681 result shape and the upstream eip681 parser logic
- preserve exact decimal amounts as strings
- keep the JNI bridge to a single shared parse_to_json call

## Dependency

- https://github.com/zcash/librustzcash/pull/2989

## Testing

- ./scripts/ci-local.sh quick passed
- managed Pixel 2 API 36 sdk-lib instrumentation suite passed
- PaymentUriParser instrumentation covers every result variant (Bitcoin, Litecoin, Solana transfer/transaction, Ethereum native/ERC-20/unrecognised), testnet/regtest network decoding, and full field assertions per case

## Note

Supersedes #2195, which was opened from a personal fork despite direct push access to this repo. Same commits, this time on a branch pushed directly here. Closing #2195 in favor of this PR.

AI disclosure: Implemented with OpenAI Codex and Claude Code assistance.


sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
```diff
@@ -0,0 +1,140 @@
...
+            message = optionalString("message")
+        )
+
+    private fun JSONObject.toSolanaTransfer() =
+        SolanaPayTransferRequest(
+            recipient = PaymentUriAddress(getString("recipient")),
+            amount = optionalString("amount")?.let(::PaymentUriAmount),
+            splToken = optionalString("spl_token")?.let(::PaymentUriAddress),
+            references =
+                getJSONArray("references").let { values ->
```
noop-sk: `references` is read via the required `getJSONArray("references")`, while every sibling field in this class goes through the `optionalString()`-style null guard (see `amount`, `splToken`, `label`, `message`, `memo` above/below). Solana Pay's `reference` query param is optional per spec — if the `payment_uri` crate ever omits the `references` key instead of emitting an empty array for a no-reference transfer, this throws `JSONException`, which the outer `catch` on line 35 silently converts into a generic `InvalidPaymentUriException()`, rejecting an otherwise-valid Solana Pay link with no diagnostic trail.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
```diff
@@ -0,0 +1,140 @@
...
+ * result into these Kotlin types.
+ */
+class PaymentUriParser private constructor(
+    private val paymentUri: PaymentUri
+) {
+    /** Parses and validates a Bitcoin, Ethereum, Litecoin, or Solana payment URI. */
+    fun parse(input: String): PaymentUriRequest =
+        try {
+            JSONObject(paymentUri.parse(input)).toPaymentRequest()
+        } catch (_: Exception) {
```
noop-sk: `parse()` catches bare `Exception` and rethrows a fixed `InvalidPaymentUriException()`, discarding the original cause entirely. A JNI `RuntimeException` from a native-library fault, a `JSONException` from an unexpected schema (see the `references` comment on `toSolanaTransfer()`), or a genuine `NullPointerException` from a future bug in the decode helpers all surface identically to a user pasting a typo'd URI — no message, type, or stack trace survives, making real SDK/environment defects indistinguishable from bad user input in logs and crash reports.

backend-lib/src/main/rust/payment_uri.rs
```diff
@@ -0,0 +1,25 @@
...
+pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustPaymentUriTool_parsePaymentUri<
+    'local,
+>(
+    mut env: JNIEnv<'local>,
+    _: JClass<'local>,
+    input: JString<'local>,
+) -> jstring {
+    let result = catch_unwind(&mut env, |env| {
+        let input = java_string_to_rust(env, &input)?;
+        let json = parse_to_json(&input).map_err(|_| anyhow::anyhow!("Invalid payment URI"))?;
```
noop-sk: `parse_to_json`'s real error is discarded via `map_err(|_| anyhow::anyhow!("Invalid payment URI"))`. Combined with the Kotlin-side `catch (_: Exception)` in `PaymentUriParser.parse()`, any genuine internal failure in the `payment_uri` crate (not a malformed user URI) is reported with zero diagnostic content anywhere in the stack, making it impossible to distinguish a real bug from bad input.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/PaymentUriRequest.kt
```diff
@@ -0,0 +1,103 @@
...
+    val amount: PaymentUriAmount?,
+    val splToken: PaymentUriAddress?,
+    val references: List<PaymentUriAddress>,
+    val label: String?,
+    val message: String?,
+    val memo: String?
+)
+
+/** Parsed EIP-681 payment request. */
+sealed interface Eip681PaymentRequest {
```
noop-sk, line 51: This introduces a second, independently maintained Ethereum/EIP-681 decoder. `eip681.rs` / `RustEip681Tool` / `JniEip681TransactionRequest` already exist on `main` and marshal the identical `Native`/`Erc20`/`Unrecognised` field set directly across JNI with compile-time field checking and no JSON hop, wrapping the same underlying `eip681` crate. Now two decoders exist for the same protocol: a future EIP-681 fix applied to one bridge and not the other would let the same `ethereum:` URI decode differently depending on which entry point a caller uses, with nothing to catch the drift at compile time.

backend-lib/Cargo.toml
```diff
@@ -59,7 +59,8 @@ zcash_proofs = "0.30"
 zcash_protocol = "0.10"
 zcash_script = "0.4"
 zip32 = "0.2"
-eip681 = "0.1.0"
+eip681 = { git = "https://github.com/zcash/librustzcash.git", rev = "3460a6a7be8d3e86b7556b38738e05e95a940f68" }
```
noop-sk: `eip681` and `payment_uri` (this line and the next) are pinned by raw git rev to a commit tied to an unmerged upstream PR (zcash/librustzcash#2989), outside this file's own `[patch.crates-io]` convention used elsewhere for exactly this kind of temporary pin. If #2989 is rebased/force-pushed before merge, or its branch is deleted and GC reclaims the now-unreferenced commit, this rev becomes unreachable — breaking `cargo fetch`/reproducible builds for anyone re-cloning after that point, with no marker flagging it as a temporary pin to replace once librustzcash cuts a release.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
```diff
@@ -0,0 +1,140 @@
...
+    private fun String.toPaymentUriNetwork() =
+        when (this) {
+            "mainnet" -> PaymentUriNetwork.Mainnet
+            "testnet" -> PaymentUriNetwork.Testnet
+            "regtest" -> PaymentUriNetwork.Regtest
+            else -> throw InvalidPaymentUriException()
+        }
+
+    companion object {
+        private const val ENCODED_VERSION = 1
```
noop-sk, line 182: `ENCODED_VERSION = 1` is the contract for the Rust crate's JSON envelope version, but it lives only as a private Kotlin constant with no shared source of truth or documented bump procedure. If the `payment_uri` crate's JSON schema changes and its emitted version increments, the only symptom is every `parse()` call throwing the same generic `InvalidPaymentUriException` as plain bad user input — no explicit version-mismatch signal anywhere in the stack.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
```diff
@@ -0,0 +1,140 @@
...
+            references =
+                getJSONArray("references").let { values ->
+                    List(values.length()) { PaymentUriAddress(values.getString(it)) }
+                },
+            label = optionalString("label"),
+            message = optionalString("message"),
+            memo = optionalString("memo")
+        )
+
+    private fun JSONObject.toEthereumNativeRequest() =
```
noop-sk: `toEthereumNativeRequest()` and `toEthereumErc20Request()` (below) copy-paste identical extraction of `schemaPrefix`, `hasPay`, and `chainId`, diverging only on the recipient/token/value fields. Not a bug today, but a future common-field addition to both EIP-681 variants requires updating both functions in lockstep — easy to update one and forget the other, producing a compiling-but-silently-incomplete decode for whichever variant was missed.

backend-lib/src/main/rust/lib.rs
```diff
@@ -107,6 +107,7 @@ use crate::utils::{
 };
 
 mod eip681;
+mod payment_uri;
```
noop-sk: `mod payment_uri;` is inserted between `mod eip681;` and `mod migration;`, breaking this file's existing alphabetical module ordering (correct placement would be between `migration_send_max` and `tor`). Not a runtime issue, just worth fixing to avoid needless merge noise on this frequently-touched file for anyone scanning it alphabetically.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
```diff
@@ -0,0 +1,140 @@
...
+        Eip681PaymentRequest.Erc20(
+            schemaPrefix = getString("schema_prefix"),
+            hasPay = getBoolean("has_pay"),
+            chainId = optionalString("chain_id"),
+            tokenContractAddress = PaymentUriAddress(getString("token_contract_address")),
+            recipientAddress = PaymentUriAddress(getString("recipient_address")),
+            valueHex = getString("value_hex")
+        )
+
+    private fun JSONObject.optionalString(name: String): String? =
```
noop-sk, line 161: `optionalString()` reimplements a nullable-JSONObject-field idiom already written ad hoc elsewhere in the codebase (an `isNull`-then-`getString` pattern). Low cost today, but each new consumer re-deriving its own null-check idiom instead of sharing one canonical helper risks a subtly different variant (e.g. `optString` instead of the `isNull` guard) silently diverging in one file without review catching it as "the same helper".

noop-sk (commented):
Automated review pass (Claude Code) on the SDK-side of the cross-chain payment URI parser. Filed as individual inline comments below; summary: the most significant issues are a second, independently-maintained EIP-681 decoder duplicating the existing typed-JNI `RustEip681Tool`/`eip681.rs` bridge on `main`, and two-layer error swallowing (Rust `map_err` + Kotlin `catch Exception`) that makes real bugs indistinguishable from bad user input.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
noop-sk: Fixed — `references` is now read via `optJSONArray("references")` with a safe default, no longer throws on a missing key.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
noop-sk: Fixed — `parse()` now rethrows with `cause = e`, preserving the original exception. (Note: this now interacts with a new double-wrap issue when the caught exception is itself an internally-thrown `InvalidPaymentUriException` — flagged separately.)

backend-lib/src/main/rust/payment_uri.rs
noop-sk: Fixed — the Rust side now categorizes the error via `error_kind()` and includes it in the message instead of a flat generic string.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/PaymentUriRequest.kt
noop-sk: Fixed — the duplicate EIP-681 JNI bridge (`RustEip681Tool`/`JniEip681TransactionRequest`) was removed; no dangling references remain.

backend-lib/Cargo.toml
noop-sk: Fixed — the git-rev pin now points at librustzcash#2989's actual merge commit on `main` (permanent, reachable) instead of the PR branch tip, with a TODO flagging it as temporary pending a crates.io release.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
noop-sk: Still open — `ENCODED_VERSION` is unchanged, still a bare private constant with no link to the `payment_uri` crate's envelope version.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
noop-sk: Fixed — both `toEthereumNativeRequest()`/`toEthereumErc20Request()` now share a `commonEip681Fields()` extension for `schemaPrefix`/`hasPay`/`chainId`.

backend-lib/src/main/rust/lib.rs
noop-sk: Fixed — `mod payment_uri;` now sits in correct alphabetical order between `migration_send_max` and `tor`.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
noop-sk: Still open — `optionalString()` is unchanged, still a local reimplementation rather than a shared helper.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
noop-sk: Note: new PRs against this repo (zcash/zcash-android-wallet-sdk) are now deprecated — SDK development is moving to zodl-inc/zodl-android-wallet-sdk. Please retarget/recreate this PR against the new repo (candidate/maint line matching v3.1.x there) rather than landing it here.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
zcash-harry: Good catch — fixed. parse() now catches InvalidPaymentUriException first and rethrows it as-is before the broader catch, so toPaymentRequest()'s own version-mismatch/unrecognized-type/unrecognized-network throws no longer get wrapped a second time.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
zcash-harry: Added a doc comment on ENCODED_VERSION linking directly to payment_uri::JSON_VERSION in librustzcash's lib.rs as the actual source of truth, with a note to bump both together. Didn't add a runtime cross-check via FFI since there's no existing precedent for that kind of round-trip in this codebase and it felt like more machinery than the actual risk (a schema bump) warrants -- let me know if you think it's worth it.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/PaymentUriParser.kt
zcash-harry: I looked for the JSONObject-nullable-field idiom you're referring to elsewhere in the codebase and couldn't find one -- the only other isNull( usages I found are Cursor.isNull (a completely different API, for SQLite queries in TransactionTable.kt/TxOutputsView.kt), not JSONObject.isNull. Since I can't find an actual second JSONObject call site to unify with, I'm leaving optionalString() as the sole implementation for now -- happy to extract a shared helper if you can point me at the other one you had in mind.

zcash-harry:
Retargeting per the review note above -- SDK development has moved to zodl-inc/zodl-android-wallet-sdk. Superseded by zodl-inc/zodl-android-wallet-sdk#7 (squashed, targeting maint/v3.1.x there). Closing this one.

PR #2197, opened 2026-08-24
MOB-1694: reject out-of-range block heights at the JNI boundary

nesence-m:
## What
The JNI entry points `branchIdForHeight` and `putUtxo` converted an incoming `jlong` block height with `height as u32`. Negative values two's-complement-wrap (e.g. -1 -> 4294967295) and values above `u32::MAX` truncate. The same unchecked-cast pattern was also present in `migration.rs`'s pool-migration entry points (`recordTransferResultNative`, `hasOverdueTransfersNative`, `nextDueTransferNative`, `nextStepNative` for block heights; the `proposal_handle` parameter shared by several commit entry points; and `buildKeystoneSignBatchQrPartsNative`'s `max_fragment_len`), which are now fixed here too.

## Why it matters
- `branchIdForHeight`: `BranchId::for_height` could silently select the wrong consensus branch id for a malformed height.
- `putUtxo`: a wrapped/truncated height would be persisted into the wallet DB as a bogus UTXO height.
- The migration entry points could similarly misjudge transfer timing, or misinterpret a `proposal_handle`/fragment length as an unrelated value.

To be clear about severity: this is defense-in-depth at the raw `Backend`/`MigrationRustBackend` JNI boundary, not a fix for a currently-reachable app crash or misbehavior. The Kotlin `TypesafeBackend` layer already rejects out-of-range heights before they reach these entry points, via `BlockHeight`'s `require(...)` range check, so the app itself cannot currently trigger this path with a bad value.

This is Least Authority audit finding MOB-1694 (parent MOB-1687, label I-SECURITY, risk: High): https://linear.app/zodl/issue/MOB-1694/g-block-height-cast-to-u32-allows-wrap-and-truncation-and-misbinds

## Fix
All these call sites now use the checked `BlockHeight::try_from(height)` conversion (or the equivalent checked conversion for the handle/length parameters), matching the idiom already used elsewhere in these files. An out-of-range value now propagates through the existing anyhow + `unwrap_exc_or` plumbing and surfaces to Kotlin as a thrown `RuntimeException` instead of silently producing a wrong result. No Kotlin-side changes were needed - `RustBackend.getBranchIdForHeight` just forwards the `Long`.

`putUtxo` has no Linear ticket of its own but is the same bug class under parent MOB-1687; included here per reviewer/user decision. The `migration.rs` sibling sites were flagged by reviewer noop-sk on this PR and are fixed in the same branch.

## Out of scope
The smaller `index as u32` outpoint cast in `putUtxo` is a different field (not a height) and is left as a possible follow-up. The `parse_network(network_id as u32)` cast sites in both `lib.rs` and `migration.rs` are intentionally left alone here - they're covered by the separate `bugfix/MOB-1764-jni-cast-sweep` branch, and fixing them here would create a merge collision.

## Testing
`cargo check` passes in `backend-lib` with zero `Cargo.lock` diff. No new Rust unit tests: these JNI functions require a live `JNIEnv` and have no existing Rust unit-test harness in `lib.rs`/`migration.rs`; the conversion itself is upstream `zcash_protocol` behavior already exercised there. CI's existing gates cover compilation across targets.

### DIFF
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -2091,7 +2091,10 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_putUtxo<'
                     .map_err(|_| anyhow!("Invalid UTXO value"))?,
                 script_pubkey,
             },
-            Some(BlockHeight::try_from(height)?),
+            Some(
+                BlockHeight::try_from(height)
+                    .map_err(|_| anyhow!("Invalid block height: {}", height))?,
+            ),
             // This JNI entry point has no account context to attribute the UTXO to.
             None,
             None,
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -2091,7 +2091,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_putUtxo<'
                     .map_err(|_| anyhow!("Invalid UTXO value"))?,
                 script_pubkey,
             },
-            Some(BlockHeight::from(height as u32)),
+            Some(BlockHeight::try_from(height)?),
             // This JNI entry point has no account context to attribute the UTXO to.
             None,
             None,
@@ -2752,7 +2752,9 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_branchIdF
     let res = panic::catch_unwind(|| {
         let _span = tracing::info_span!("RustBackend.branchIdForHeight").entered();
         let network = parse_network(network_id as u32)?;
-        let branch: BranchId = BranchId::for_height(&network, BlockHeight::from(height as u32));
+        let height = BlockHeight::try_from(height)
+            .map_err(|_| anyhow!("Invalid block height: {}", height))?;
+        let branch: BranchId = BranchId::for_height(&network, height);
         let branch_id: u32 = u32::from(branch);
         debug!(
             "For height {} found consensus branch {:?} with id {}",
--- a/backend-lib/src/main/rust/migration.rs
+++ b/backend-lib/src/main/rust/migration.rs
@@ -410,6 +410,15 @@ fn target_height(wallet: &Wallet) -> anyhow::Result<BlockHeight> {
     Ok(tip + 1)
 }
 
+/// Converts a non-negative `jlong` tip height reported by a caller into a [`BlockHeight`],
+/// erroring instead of silently truncating a value above the `u32` range into a different
+/// height. Callers own any negative-sentinel handling ("no estimate available") themselves —
+/// this only decodes the case where the caller has already established the value is meant to
+/// be taken literally.
+fn decode_tip_height(height: jlong) -> anyhow::Result<BlockHeight> {
+    BlockHeight::try_from(height).map_err(|_| anyhow!("Invalid tip height: {}", height))
+}
+
 /// The wallet's real, currently-witnessable anchor height (the same one ordinary, non-migration
 /// sends use, via `get_target_and_anchor_heights`) — NOT just "chain tip minus one", which isn't
 /// necessarily checkpointed (confirmed live: `root_at_checkpoint_id` returned `None` for a raw
@@ -747,6 +756,12 @@ fn decode_transfer_id(id: jlong) -> anyhow::Result<MigrationTransferId> {
     Ok(MigrationTransferId::new(idx))
 }
 
+/// Decodes the `proposalHandle` field Kotlin echoes back into a cache lookup key, rejecting a
+/// negative handle rather than reinterpreting its bit pattern as a large `u64`.
+fn decode_plan_handle(handle: jlong) -> anyhow::Result<crate::migration_plan_cache::PlanHandle> {
+    u64::try_from(handle).map_err(|_| anyhow!("Invalid proposal handle: {}", handle))
+}
+
 /// One preparation (note-split) transaction entry, ready to encode into [`JniPreparationStep`].
 ///
 /// - `id` is the stable [`MigrationTransferId`] the engine assigns at commit time (same ordinal as
@@ -1061,7 +1076,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_MigrationRustBackend_
     let res = catch_unwind(&mut env, |env| {
         let (_network, wallet, _store_conn) = open(env, db_data, network_id)?;
         let account = crate::account_id_from_jni(env, account_uuid)?;
-        let plan_handle = proposal_handle as u64;
+        let plan_handle = decode_plan_handle(proposal_handle)?;
         let migration_plan = crate::migration_plan_cache::get(account, plan_handle)?;
         let tip = wallet
             .chain_height()
@@ -1392,7 +1407,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_MigrationRustBackend_
                 store_conn: &mut store_conn,
                 target,
             },
-            proposal_handle as u64,
+            decode_plan_handle(proposal_handle)?,
             |network, target, backend, migration_plan, rng| {
                 let state = engine::commit_preparation(
                     network,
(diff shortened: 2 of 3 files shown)


backend-lib/src/main/rust/lib.rs
```diff
@@ -2752,7 +2752,9 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_branchIdF
     let res = panic::catch_unwind(|| {
         let _span = tracing::info_span!("RustBackend.branchIdForHeight").entered();
         let network = parse_network(network_id as u32)?;
-        let branch: BranchId = BranchId::for_height(&network, BlockHeight::from(height as u32));
+        let height = BlockHeight::try_from(height)
```
noop-sk, line 2751: @nullcopy — this fix's correctness rests entirely on `zcash_protocol::consensus::BlockHeight::try_from(i64)`'s exact contract (upstream crate, unmodified here). Can you confirm it rejects negative and `> u32::MAX` values correctly (inclusive/exclusive at the boundary), and that this bare `?` still surfaces a useful `Display` message rather than an opaque one (unlike this line, which has an explicit `map_err`)?

noop-sk (commented):
Reviewed via the SDK app-consumer/JNI-boundary lens. The two named call sites (`putUtxo`, `branchIdForHeight`) are fixed correctly and idiomatically, CI is green, and I confirmed every remaining `height: jlong` parameter in this file now routes through `BlockHeight::try_from`. Not blocking, but two things worth addressing before treating MOB-1694/MOB-1687 as fully closed:

**Same bug class survives in a sibling file.** `backend-lib/src/main/rust/migration.rs` (`MigrationRustBackend`, untouched by this PR) has the identical unchecked-cast pattern at 4 sibling JNI entry points, all still live on `maint/v3.1.x` after this merges: `recordTransferResultNative` (`BlockHeight::from_u32(observed_tip as u32)`), `hasOverdueTransfersNative` and `nextDueTransferNative` (`BlockHeight::from(estimated_tip as u32)`), and `nextStepNative` (`BlockHeight::from_u32(estimated_tip as u32) + 1`). All four guard the lower bound already but not the upper — a value above `u32::MAX` still silently truncates. Severity is lower here (chain-tip estimates computed internally, not attacker-supplied consensus data like `branchIdForHeight`/`putUtxo`), but it's the same audit finding's bug class left half-covered. Worth a follow-up ticket/PR rather than closing the audit item as done.

**No new tests for the boundary values.** The existing `BranchIdTest.kt` only exercises `getBranchIdForHeight` with known-valid historical heights, never `putUtxo`, and never a negative/zero/`> u32::MAX` height. Understood there's no `JNIEnv`-based Rust unit-test harness in this file, but an instrumented test for the boundary values on the Kotlin side would lock in the new behavior.

Also worth noting for context, not a defect: I traced the actual app call sites (`CompactBlockProcessor.kt`, `TransactionEncoderImpl.kt`) and `TypesafeBackend`'s Kotlin wrapper already rejects out-of-range heights via `BlockHeight`'s own `require(UINT_RANGE.contains(value))` before reaching the raw JNI call — so this fix is defense-in-depth for the raw `Backend` interface, not a fix for a currently-reachable app crash. Might be worth tightening the PR description if it implies more real-world impact than currently exists.

nesence-m:
Thanks for catching this - the sibling sites in migration.rs are now fixed in this PR too. Converted the same unchecked-cast pattern to the checked BlockHeight::try_from idiom at:

- recordTransferResultNative
- hasOverdueTransfersNative
- nextDueTransferNative
- nextStepNative

While I was in there I also found and fixed the same pattern on two non-height parameters shared by nearby entry points: the proposal_handle: jlong cast (as u64) used by the migration commit entry points, and buildKeystoneSignBatchQrPartsNative's max_fragment_len: jint cast (as usize).

Left the two parse_network(network_id as u32) sites alone deliberately - those are covered by bugfix/MOB-1764-jni-cast-sweep, and touching them here would collide with that branch.

On boundary-value tests: agreed on the rationale, but I'm not adding them here. TypesafeBackend/BlockHeight's require(UINT_RANGE...) already pins the app-reachable behavior on the Kotlin side, and this file has no JNIEnv-based Rust test harness to exercise these JNI entry points directly, so there's no natural place to add a boundary-value test for this fix without building new test infrastructure.

Also tightened the PR description per your impact-claim note - it now explicitly frames this as defense-in-depth at the raw JNI boundary rather than a fix for a currently-reachable app crash.

backend-lib/src/main/rust/migration.rs
```diff
@@ -410,6 +410,15 @@ fn target_height(wallet: &Wallet) -> anyhow::Result<BlockHeight> {
     Ok(tip + 1)
 }
 
+/// Converts a non-negative `jlong` tip height reported by a caller into a [`BlockHeight`],
+/// erroring instead of silently truncating a value above the `u32` range into a different
+/// height. Callers own any negative-sentinel handling ("no estimate available") themselves —
+/// this only decodes the case where the caller has already established the value is meant to
+/// be taken literally.
+fn decode_tip_height(height: jlong) -> anyhow::Result<BlockHeight> {
```
noop-sk, line 427: **Now unit-testable, unlike the JNI entry points**

The boundary-value test decline made sense for the `extern "C"` functions (no JNIEnv-based test harness), but `decode_tip_height`/`decode_plan_handle`/`decode_max_fragment_len` are plain functions taking primitive `jlong`/`jint` and returning `Result` — no `JNIEnv` involved, so `assert!(decode_tip_height(-1).is_err())` and similar are trivially addable to this file's existing `#[cfg(test)]` module without any harness. Worth adding a handful of cases given these three helpers are now the actual gate for every height/handle/length value crossing this boundary.

noop-sk (commented):
Verified the migration.rs fix — all 4 named sites now route through `decode_tip_height`, plus the extra scope Milan flagged (6 `proposal_handle as u64` sites via `decode_plan_handle`, `max_fragment_len as usize` via `decode_max_fragment_len`) checks out as legitimate, not overclaimed; swept the rest of migration.rs myself and didn't find any remaining instance of the pattern. `parse_network` sites confirmed genuinely deferred to the other branch as stated. PR description and CHANGELOG updated as described.

Two small things still open, not blocking:

1. The @nullcopy question on this PR (lib.rs, the `branchIdForHeight`/`putUtxo` thread) still has no reply from anyone. I independently checked the underlying fact against upstream `zcash_protocol::consensus::BlockHeight`'s `TryFrom<i64>` impl and it does correctly bound-check both negative and `> u32::MAX` — so a reply should be quick. The one open half of that question — `putUtxo`'s bare `?` giving a generic "out of range integral type conversion attempted" message with no height value, unlike `branchIdForHeight`'s explicit `anyhow!("Invalid block height: {}", height)` — is still a real, unaddressed inconsistency.
2. Left one inline note on the new decode helpers re: test coverage.

Also noticed this PR currently shows `mergeable: CONFLICTING` — will need a rebase before it can land, unrelated to the review content.

backend-lib/src/main/rust/migration.rs
nesence-m: Added in 5bb33d7c — boundary tests for all three helpers in migration.rs's `#[cfg(test)]` module, each matched to the helper's actual contract (`decode_tip_height` 0..=u32::MAX, `decode_plan_handle` rejects negatives, `decode_max_fragment_len` rejects non-positives). 13 cases, all passing under `cargo test`.

backend-lib/src/main/rust/lib.rs
nesence-m: The putUtxo half of this is fixed in 5bb33d7c — the bare `?` now maps to the same explicit `anyhow!("Invalid block height: {}", height)` as branchIdForHeight. The `try_from` contract confirmation stays with @nullcopy.

nesence-m:
Moved to https://github.com/zodl-inc/zodl-android-wallet-sdk/pull/10 as part of the repo migration to zodl-inc.

backend-lib/src/main/rust/lib.rs
nullcopy: Yes, `BlockHeight::try_from(i64)` rejects negative and `> u32::MAX` heights, and the error should have a meaningful `Display` value.

PR #2198, opened 2026-08-24
[MOB-1689] Harden secret-key handling across the JNI boundary

nesence-m:
## Summary

Hardening for the Least Authority audit finding tracked as [MOB-1689](https://linear.app/zodl/issue/MOB-1689/b-high-privilege-keys-exposed-across-jni-boundary) ("High-Privilege Keys Exposed Across JNI Boundary"): JNI exports that derive spending and ZIP-325 metadata keys returned them to Kotlin as raw `byte[]`, leaving unwiped copies of `UnifiedSpendingKey` bytes and metadata keys on the JVM heap, outside Rust's secrecy controls.

**No public API changes.** This PR shrinks the exposure window at every point the SDK controls:

- **Rust (`backend-lib/src/main/rust`)**: all outbound key material now travels through `secrecy::SecretVec`-owned buffers that zeroize on drop (`encode_metadata_key`, private-use metadata keys, arbitrary wallet/account keys — matching the pattern `encode_usk`/`deriveSpendingKey` already used), and the inbound metadata-key halves of `derivePrivateUseMetadataKey` now go through `secret_from_jni` instead of an unprotected `Vec<u8>`. No `Cargo.toml` changes.
- **Kotlin**: new internal `ByteArray.clearContents()`/`useAndClear()` utility; every transient SDK-created copy of key material is zeroed in a `finally` once the JNI call completes — spending-key derivation, full-viewing-key derivation, metadata-key derivation, `createProposedTransactions`, `createAccountAndGetSpendingKey` (previously the returned `JniAccountUsk` bytes were abandoned to the GC), and both Orchard-migration signing paths. Caller-supplied seed arrays and arrays returned to the consumer are never touched.
- **Disclosure fixes**: `JniMetadataKey` gains the redacted `toString()` every other JNI key model already had (plus corrected `require` messages); `AccountMetadataKey.toUnsafe()` now returns defensive copies instead of its live internal arrays; `AccountCreateSetup.toString()` no longer hex-dumps the seed via `FirstClassByteArray`; KDoc added on `deriveArbitraryWalletKey`/`deriveArbitraryAccountKey` advising callers to zero returned bytes.

### Honest limitations (stated deliberately)

- JVM-side wiping is **best-effort**: GC compaction and JIT can leave unreachable copies that application code cannot clear.
- The upstream `zip32` key structs themselves are not zeroized on drop; the `SecretVec` wrappers cover the owned transit buffers.
- The seed and derived keys still cross the JNI boundary by API design. Full elimination — opaque key handles held in a Rust-side registry (modeled on the existing `migration_plan_cache.rs` pattern) so keys never enter the JVM heap — is a breaking API change deferred to a follow-up.
- The Slipstream call sites (`SlipstreamSpendService.createProposedTransactions`, `SlipstreamBroadcaster.createProposedTransactions`) were in scope on this PR's base, `maint/v3.1.x`, today — not a hypothetical future backport — and are now fixed with the same `clearContents()` pattern used elsewhere in this PR.

## Test plan

- `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test` (backend-lib): clean; 74 passed, 15 ignored (env-gated), 0 failed.
- `cargo tree -d`: no duplicated `zcash_*`/`orchard`/`sapling-crypto`/`rusqlite` crates; `Cargo.lock` untouched.
- `./gradlew ktlint detektAll lint`: clean (detekt report empty; only pre-existing baseline lint warnings).
- `./gradlew :backend-lib:testDebugUnitTest :sdk-lib:testDebugUnitTest`: 205 tests, 0 failures — including new unit tests for the wipe utility, redacted `toString()`s, `toUnsafe()` defensive copies, and transient-wipe assertions on `TypesafeBackendImpl.createProposedTransactions` / `DerivationToolExt.deriveUnifiedSpendingKey` via capturing fakes.
- `./gradlew :backend-lib:assembleDebug`: Rust cross-compile green for arm/arm64/x86/x86_64.
- androidTest (connected) suites not run locally — no device attached; CI covers them. The seed-non-mutation contract (`create_spending_key_does_not_mutate_passed_bytes`) is preserved by design (caller arrays are never wiped).

### DIFF
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/ByteArrayExt.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/ByteArrayExt.kt
@@ -12,7 +12,14 @@ fun ByteArray.clearContents() {
 
 /**
  * Runs [block] with this array and zeroes it afterwards, including when [block]
- * throws. The array must not be retained past [block]'s return.
+ * throws.
+ *
+ * [block] must not retain the array past its return - no assigning it to a field, closing
+ * over it in a callback, handing it to another coroutine, or returning it as (part of) [R].
+ * This array is zeroed unconditionally once [block] returns or throws, so any alias [block]
+ * kept escapes with its contents silently wiped out from under it. There is no runtime check
+ * for this: the no-aliasing contract is convention-only, enforced by caller discipline, not by
+ * this function.
  */
 inline fun <R> ByteArray.useAndClear(block: (ByteArray) -> R): R =
     try {
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/ByteArrayExt.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/ByteArrayExt.kt
@@ -1,0 +1,22 @@
+@file:Suppress("ktlint:standard:filename")
+
+package cash.z.ecc.android.sdk.internal.ext
+
+/**
+ * Overwrites this array with zeros. Best effort only: the JVM may hold unreachable
+ * copies (GC compaction, JIT) that cannot be cleared from here.
+ */
+fun ByteArray.clearContents() {
+    fill(0)
+}
+
+/**
+ * Runs [block] with this array and zeroes it afterwards, including when [block]
+ * throws. The array must not be retained past [block]'s return.
+ */
+inline fun <R> ByteArray.useAndClear(block: (ByteArray) -> R): R =
+    try {
+        block(this)
+    } finally {
+        fill(0)
+    }
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -2932,12 +2932,14 @@ fn encode_metadata_key<'a>(
     env: &mut JNIEnv<'a>,
     key: zip32::registered::SecretKey,
 ) -> anyhow::Result<JObject<'a>> {
+    let sk = SecretVec::new(key.data().to_vec());
+    let chain_code = SecretVec::new(key.chain_code().as_bytes().to_vec());
     Ok(env.new_object(
         "cash/z/ecc/android/sdk/internal/model/JniMetadataKey",
         "([B[B)V",
         &[
-            (&env.byte_array_from_slice(key.data())?).into(),
-            (&env.byte_array_from_slice(key.chain_code().as_bytes())?).into(),
+            (&env.byte_array_from_slice(sk.expose_secret())?).into(),
+            (&env.byte_array_from_slice(chain_code.expose_secret())?).into(),
         ],
     )?)
 }
@@ -2989,20 +2991,22 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustDerivationTool_de
 ) -> jobjectArray {
     let res = catch_unwind(&mut env, |env| {
         let _span = tracing::info_span!("RustDerivationTool.derivePrivateUseMetadataKey").entered();
-        let account_metadata_key_sk = utils::java_bytes_to_rust(env, &account_metadata_key_sk)?;
-        let account_metadata_key_c = utils::java_bytes_to_rust(env, &account_metadata_key_c)?;
+        let account_metadata_key_sk = secret_from_jni(env, account_metadata_key_sk)?;
+        let account_metadata_key_c = secret_from_jni(env, account_metadata_key_c)?;
         let ufvk_string = utils::java_nullable_string_to_rust(env, &ufvk_string)?;
         let private_use_subject = utils::java_bytes_to_rust(env, &private_use_subject)?;
         let network = parse_network(network_id as u32)?;
 
         let account_metadata_key = {
             let sk = account_metadata_key_sk
+                .expose_secret()
                 .as_slice()
                 .try_into()
                 .map_err(|_| anyhow!("Incorrect length for account_metadata_key_sk"))?;
 
             let chain_code = ChainCode::new(
                 account_metadata_key_c
+                    .expose_secret()
                     .as_slice()
                     .try_into()
                     .map_err(|_| anyhow!("Incorrect length for account_metadata_key_c"))?,
@@ -3054,7 +3058,8 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustDerivationTool_de
 
         Ok(
             utils::rust_vec_to_java(env, private_use_keys, "[B", |env, key| {
-                utils::rust_bytes_to_java(env, key.data())
+                let key_bytes = SecretVec::new(key.data().to_vec());
+                utils::rust_bytes_to_java(env, key_bytes.expose_secret())
             })?
             .into_raw(),
         )
@@ -3079,8 +3084,9 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustDerivationTool_de
 
         let key =
             zip32::arbitrary::SecretKey::from_path(&context_string, seed.expose_secret(), &[]);
+        let key_bytes = SecretVec::new(key.data().to_vec());
 
-        Ok(utils::rust_bytes_to_java(&env, key.data())?.into_raw())
+        Ok(utils::rust_bytes_to_java(&env, key_bytes.expose_secret())?.into_raw())
     });
     unwrap_exc_or(&mut env, res, ptr::null_mut())
 }
@@ -3113,8 +3119,9 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustDerivationTool_de
                 ChildIndex::hardened(account.into()),
             ],
         );
+        let key_bytes = SecretVec::new(key.data().to_vec());
 
-        Ok(utils::rust_bytes_to_java(&env, key.data())?.into_raw())
+        Ok(utils::rust_bytes_to_java(&env, key_bytes.expose_secret())?.into_raw())
     });
     unwrap_exc_or(&mut env, res, ptr::null_mut())
 }
--- a/sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/FirstClassByteArray.kt
+++ b/sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/FirstClassByteArray.kt
@@ -2,6 +2,11 @@ package cash.z.ecc.android.sdk.model
 
 import cash.z.ecc.android.sdk.ext.toHex
 
+/**
+ * [toString] hex-dumps [byteArray], which is appropriate for non-secret data (transaction
+ * ids, raw transactions, and the like). Types that embed secret material in a
+ * [FirstClassByteArray] must override [toString] themselves to redact it.
+ */
 class FirstClassByteArray(
     val byteArray: ByteArray
 ) {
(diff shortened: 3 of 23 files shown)


noop-sk (commented):
Reviewed the app-consumer/Kotlin lane and traced the aliasing safety of every construct-then-wipe site this PR touches (`UnifiedSpendingKey`, `AccountUsk`, `AccountMetadataKey`) — all defensively `.copyOf()` before the source array is zeroed, so the wipe-after-return bug this pattern could introduce doesn't currently exist. CI is green, no supply-chain concerns (no Cargo/Gradle/CI changes). Left two Kotlin-lane notes inline and two core-team questions on the Rust side.

**The one thing I'd want addressed before this leaves draft:** `slipstream-lib/src/main/java/com/zodl/slipstream/internal/spend/SlipstreamSpendService.kt:128` and `slipstream-lib/.../SlipstreamBroadcaster.kt:54` both call `backend.createProposedTransactions(proposal.toUnsafe(), usk.copyBytes())` with the exact unwiped pattern this PR just fixed in `TypesafeBackendImpl.createProposedTransactions` — confirmed both files are unchanged at this PR's head and live on `maint/v3.1.x`, the branch this PR targets today. The PR's "Honest limitations" section frames this as "relevant only if a backport is ever wanted," but it's in scope on the current branch right now, not a hypothetical backport target — and it's the same audit-relevant leak class (MOB-1689) this PR is fixing everywhere else. Since neither file is touched by this diff I can't leave an inline comment on them, but wanted to flag it here before this comes out of draft.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/ByteArrayExt.kt
```diff
@@ -0,0 +1,22 @@
...
+ */
+fun ByteArray.clearContents() {
+    fill(0)
+}
+
+/**
+ * Runs [block] with this array and zeroes it afterwards, including when [block]
+ * throws. The array must not be retained past [block]'s return.
+ */
+inline fun <R> ByteArray.useAndClear(block: (ByteArray) -> R): R =
```
noop-sk, line 24: **Safety is convention-only, not enforced**

`useAndClear`'s safety depends entirely on `block` never retaining the array reference past return (the doc says so), but nothing enforces that. It holds for every current call site because the wrapping constructors copy defensively before this runs — but a future caller that aliases instead of copies would silently zero out a "live" returned key with no compile-time or runtime signal. Worth a stronger warning in the doc, or a cheap runtime assertion (e.g. re-check a canary byte after `block` returns) rather than relying purely on the convention.

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/FirstClassByteArray.kt
```diff
@@ -2,6 +2,11 @@ package cash.z.ecc.android.sdk.model
 
 import cash.z.ecc.android.sdk.ext.toHex
 
+/**
+ * [toString] hex-dumps [byteArray], which is appropriate for non-secret data (transaction
+ * ids, raw transactions, and the like). Types that embed secret material in a
+ * [FirstClassByteArray] must override [toString] themselves to redact it.
```
noop-sk, line 8: **Redaction is a convention with no compiler enforcement**

Every current secret-bearing subclass (`UnifiedSpendingKey`, `AccountMetadataKey`, `AccountCreateSetup`) does override `toString()` correctly, but nothing catches a future type that wraps secret bytes in `FirstClassByteArray` and forgets to. Given this PR exists specifically because of an audit finding about leaking key material, worth considering whether this needs more than a doc comment — e.g. a marker interface/sealed hierarchy that forces the override, or at minimum a lint rule.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2932,12 +2932,14 @@ fn encode_metadata_key<'a>(
     env: &mut JNIEnv<'a>,
     key: zip32::registered::SecretKey,
 ) -> anyhow::Result<JObject<'a>> {
+    let sk = SecretVec::new(key.data().to_vec());
+    let chain_code = SecretVec::new(key.chain_code().as_bytes().to_vec());
```
noop-sk, line 2936: @nullcopy — the PR body notes `zip32::registered::SecretKey`'s own backing memory isn't zeroized on drop; only the `SecretVec` copies taken from `key.data()`/`key.chain_code().as_bytes()` are protected. Is this an accepted residual risk given upstream `zip32` doesn't implement `Zeroize`, or is there appetite to raise it with the crate or patch it locally? Matters to us since it's the exact exposure class the Least Authority finding is about.

backend-lib/src/main/rust/lib.rs
```diff
@@ -2989,20 +2991,22 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustDerivationTool_de
...
-        let account_metadata_key_sk = utils::java_bytes_to_rust(env, &account_metadata_key_sk)?;
-        let account_metadata_key_c = utils::java_bytes_to_rust(env, &account_metadata_key_c)?;
+        let account_metadata_key_sk = secret_from_jni(env, account_metadata_key_sk)?;
+        let account_metadata_key_c = secret_from_jni(env, account_metadata_key_c)?;
         let ufvk_string = utils::java_nullable_string_to_rust(env, &ufvk_string)?;
         let private_use_subject = utils::java_bytes_to_rust(env, &private_use_subject)?;
         let network = parse_network(network_id as u32)?;
 
         let account_metadata_key = {
             let sk = account_metadata_key_sk
```
noop-sk, line 3001: @nullcopy — `account_metadata_key_sk.expose_secret().as_slice().try_into()` produces a fresh, un-wrapped `[u8; 32]` stack copy (feeding `ChainCode::new(...)` etc.) that sits outside the `SecretVec` these bytes were just extracted from. Does anything downstream (`ChainCode`, `SecretKey::from_bytes`) zeroize its own inputs, or is this a second unprotected transit copy that the `SecretVec` wrapping here doesn't actually cover?

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/ext/ByteArrayExt.kt
nesence-m: Strengthened the KDoc to spell out the no-aliasing contract explicitly: block must never retain the array past return (no field, no captured callback, no handing it to another coroutine, no returning it as part of R), the wipe fires unconditionally on return or throw, and there's no runtime check for any of this — it's convention-only. Deliberately skipping a runtime canary here: every current caller already copies defensively before calling in, so a canary would add a per-call cost against a bug class that's better caught in review. Happy to revisit if a non-copying caller ever shows up. Commit ab72c87e ("[MOB-1689] Wipe the two remaining un-wiped Slipstream USK copies").

sdk-lib/src/main/java/cash/z/ecc/android/sdk/model/FirstClassByteArray.kt
nesence-m: Keeping the doc-comment approach in this PR rather than adding enforcement here. A marker-interface/lint check that catches unredacted FirstClassByteArray subclasses is a good follow-up, but it's out of scope for this audit fix — I'll note it for a follow-up ticket.

nesence-m:
Addressed review 5007485464's before-undraft item: the two Slipstream `createProposedTransactions` call sites (`slipstream-lib/src/main/java/com/zodl/slipstream/internal/spend/SlipstreamSpendService.kt` and `.../SlipstreamBroadcaster.kt`) now wipe the USK copy with the same `clearContents()` pattern used elsewhere in this PR, in commit ab72c87e ("[MOB-1689] Wipe the two remaining un-wiped Slipstream USK copies"). Added `SlipstreamSpendServiceOrderingTest` coverage asserting the array is zeroed once the flow completes. Also corrected the "Honest limitations" section to state these sites were in scope on `maint/v3.1.x` today, not a hypothetical backport, as requested.

The two zeroization questions at `lib.rs:2936` and `lib.rs:3001` remain open for @nullcopy.

noop-sk (commented):
Verified the Slipstream wipe fix — both `SlipstreamSpendService` and this file now wrap only the `backend.createProposedTransactions` call in the try/finally, correctly leaving the subsequent emit/notify calls outside it ("flow-safe" as described). The `useAndClear` KDoc strengthening and the `SlipstreamSpendServiceOrderingTest` wipe assertion (with a non-empty `uskBytes` fixture, so the assertion is meaningful) are real, not cosmetic. Deferring `FirstClassByteArray`'s redaction-enforcement gap to a follow-up ticket is a reasonable call for this PR's scope, and flagging `SlipstreamSynchronizer.createAccount` (raw seed bytes, same exposure class) as a known follow-up rather than silently leaving it is appreciated — confirmed that site is real and unwiped.

One gap found while re-checking: this file's own wipe fix has no dedicated test — `SlipstreamSpendServiceOrderingTest` only covers `SlipstreamSpendService`, and there's no equivalent test file for `SlipstreamBroadcaster`. Left a note inline. Not blocking the draft, but worth closing before this comes out of it, alongside the two still-open @nullcopy questions on lib.rs.

slipstream-lib/src/main/java/com/zodl/slipstream/internal/spend/SlipstreamBroadcaster.kt
```diff
@@ -51,7 +52,13 @@ internal class SlipstreamBroadcaster(
         usk: UnifiedSpendingKey
     ): List<CreatedTransaction> {
         SaplingParams.ensureDownloaded(saplingParamsDir)
-        val txIds = backend.createProposedTransactions(proposal.toUnsafe(), usk.copyBytes())
+        val uskBytes = usk.copyBytes()
+        val txIds =
+            try {
+                backend.createProposedTransactions(proposal.toUnsafe(), uskBytes)
+            } finally {
+                uskBytes.clearContents()
```
noop-sk, line 65: **No test coverage for this wipe**

The identical fix in `SlipstreamSpendService` got a real regression test (`SlipstreamSpendServiceOrderingTest`, asserting `uskBytes` content-equals a zeroed array after the call). This file has no test file at all, so this wipe has zero automated coverage — worth a small test mirroring that one before this leaves draft.

nesence-m: Added in dc5c8356 — `SlipstreamBroadcasterTest` mirrors `SlipstreamSpendServiceOrderingTest` (same fakes and conventions, non-empty USK fixture) and asserts the wipe on both the success and backend-failure paths, via an injectable `ensureSaplingParams` seam so tests never touch the network.

nesence-m:
Moved to https://github.com/zodl-inc/zodl-android-wallet-sdk/pull/9 as part of the repo migration to zodl-inc.

backend-lib/src/main/rust/lib.rs
nullcopy: My gut reaction is that the zip32 crate should zeroize this, but I'll have to look into it to see if there's a reason why they didn't zeroize.

backend-lib/src/main/rust/lib.rs
nullcopy: I'll dig in to see, but same as above, my intuition is that this should all be zeroized

backend-lib/src/main/rust/lib.rs
nullcopy: I've created the following implement zeroization upstream: https://github.com/zcash/zip32/issues/33

PR soon

backend-lib/src/main/rust/lib.rs
nullcopy: I've created the following issue to fix the zeroization gap in librustzcash: https://github.com/zcash/librustzcash/issues/3013

PR soon

PR #2199, opened 2026-08-24
MOB-1764: convert all JNI numeric inputs with checked conversions

nesence-m:
## What

Sweeps the remaining unchecked `as u32` / `as u64` casts on caller-supplied JNI numeric arguments in `backend-lib/src/main/rust/lib.rs` (and two knock-on call sites in `migration.rs`):

1. **`putUtxo` output index (real bug fix).** A negative `jint` output index currently wraps to a huge `u32` via `index as u32` and gets persisted to the wallet DB. It now goes through `u32::try_from(index)` and returns `Err(anyhow!("Invalid UTXO output index: {}", index))` on failure.
2. **`parse_network` / `parse_tor_dormant_mode` signature change.** Both helpers now take `i32` (matching the `jint` JNI type) and do the checked `u32::try_from` conversion internally, instead of every call site doing an unchecked `as u32` before calling in. All ~50 call sites for `parse_network` and the one for `parse_tor_dormant_mode` were updated to drop the now-unnecessary cast.
3. **Subtree-roots start-index guards normalized (behavior-preserving).** The `sapling_start_index` / `orchard_start_index` / `ironwood_start_index` guards in the subtree-roots entry point were rewritten from an `if x >= 0 { x as u64 } else { return Err(...) }` shape to `u64::try_from(x).map_err(|_| anyhow!(...))?`, with the exact existing error message strings preserved.

The `accounts > 0` guard in `deriveUnifiedFullViewingKeysFromSeed` was deliberately left untouched — it enforces strictly positive, which `try_from` cannot express.

## Why

Out-of-range values passed into these JNI entry points currently wrap silently instead of failing loudly. This complements #2197 (MOB-1694), which fixed the two `height as u32` casts; this PR sweeps the rest of the unchecked-cast surface in `backend-lib`.

## Out of scope

The `tor_runtime as usize` / `lwd_conn as usize` pointer round-trips are deliberately left alone — those are pointer handles, not JNI-input wrap risks, and are tracked separately as MOB-1766. The `rate.scale() as i32` and the float `(offset_blocks * 75.0).round() as u32` casts are also untouched, for the same reason (not caller-supplied JNI input).

Linear: https://linear.app/zodl/issue/MOB-1764/sweep-remaining-unchecked-jni-numeric-casts-in-backend-lib

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   subtypes the upstream engine uses, instead of letting the raw Rust `RuntimeException` escape
   (MOB-1723).
 
+### Fixed
+- All JNI entry points now convert caller-supplied numeric arguments (network ids, the UTXO
+  output index, Tor dormant mode) with checked conversions instead of unchecked casts, so
+  out-of-range values fail with an exception instead of silently wrapping (MOB-1764).
+
 ## [3.1.0] - 2026-08-20
 
 ### Added
--- a/backend-lib/src/main/rust/lib.rs
+++ b/backend-lib/src/main/rust/lib.rs
@@ -283,7 +283,7 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_RustBackend_initDataD
     network_id: jint,
 ) -> jint {
     let res = catch_unwind(&mut env, |env| {
-        let network = parse_network(network_id as u32)?;
+        let network = parse_network(network_id)?;
         let mut db_data = wallet_db(env, network, db_data)
             .map_err(|e| anyhow!("Error while opening data DB: {}", e))?;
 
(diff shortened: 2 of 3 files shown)


noop-sk (approved):

noop-sk (approved):
Reviewed as a follow-up to #2197's JNI-boundary hardening. All four checked conversions (`parse_network`, `parse_tor_dormant_mode`, `putUtxo` index, subtree-root indices) are correct and behavior-preserving — traced the `putUtxo` path back to `CompactBlockProcessor.processUtxoResult`, confirming this closes a real reachable issue (a malformed/negative index from lightwalletd's raw UTXO response would previously wrap into a huge unsigned value before being persisted). No Kotlin/JNI-signature changes, no supply-chain concerns, CI green.

Two small non-blocking notes if you want to fold them in before merge, otherwise happy for a fast-follow:
- `parse_network`/`parse_tor_dormant_mode` give a shorter error message for a negative input (fails at the `u32::try_from` step) than for an in-range-but-unrecognized positive one (falls through to the match's `_` arm with a fuller message) — same error class, two message shapes.
- Both are plain free functions with no `JNIEnv` dependency, so `parse_network(-1).is_err()" and similar could be unit tested directly in the existing `#[cfg(test)]` module in this file, unlike the JNI entry points which need a live JNI harness.

PR #2200, opened 2026-08-24
MOB-1765: validate numeric arguments at the RustBackend boundary

nesence-m:
## Summary

- Adds argument-range validation at the raw Kotlin `Backend`/`RustBackend` boundary
  (`backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/Backend.kt` and
  `.../internal/jni/RustBackend.kt`) so out-of-range values throw `IllegalArgumentException`
  at the call site, before crossing the JNI boundary.
- Mirrors the same checks in `FakeRustBackend` so the fake and real implementations behave
  identically.

### Layered-defense rationale

- The typed `TypesafeBackend` path in `sdk-lib` is already safe: its `BlockHeight` model
  `require`s the UInt32 range in `init`.
- The gap was the raw surface in `backend-lib`: `Backend`/`RustBackend` take plain
  `Long`/`Int` with no validation before crossing JNI. `backend-lib` cannot depend on
  `sdk-lib`'s model type, so a new shared type wasn't an option — `require(...)` at the
  Kotlin entry points is the chosen fix.
- This PR adds the **first line of defense** at the Kotlin/JNI boundary. The Rust-side checks
  from #2197 (MOB-1694) and the parallel MOB-1764 sweep remain the **second line of defense**
  inside the Rust layer itself.

### Methods covered

Block heights (must be in `0..0xFFFF_FFFFL`): `getBranchIdForHeight`, `rewindToHeight`,
`updateChainTip`, `findBlockMetadata`, `rewindBlockMetadataToHeight`, `putUtxo` (height),
`scanBlocks` (fromHeight).

Indices/counts (must be nonnegative): `putUtxo` (index), `putSubtreeRoots` (sapling/orchard/
ironwood start indices), `getMemoAsUtf8` (outputIndex), `scanBlocks` (limit).

## Test plan

- [x] `./gradlew ktlint` — passes
- [x] `./gradlew detektAll` — passes
- [x] `./gradlew :backend-lib:compileDebugKotlin` — compiles
- [x] `./gradlew :backend-lib:testDebugUnitTest` — full suite green, including the new
      `FakeRustBackendValidationTest` (negative height throws, `0xFFFF_FFFFL` accepted,
      `0x1_0000_0000L` throws, negative index throws)

Linear: https://linear.app/zodl/issue/MOB-1765/validate-heightcount-arguments-at-the-rustbackend-kotlin-boundary

### DIFF
--- a/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt
+++ b/backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt
@@ -4,6 +4,7 @@ import cash.z.ecc.android.sdk.internal.Backend
 import cash.z.ecc.android.sdk.internal.SdkDispatchers
 import cash.z.ecc.android.sdk.internal.ext.deleteRecursivelySuspend
 import cash.z.ecc.android.sdk.internal.ext.deleteSuspend
+import cash.z.ecc.android.sdk.internal.ext.isInUIntRange
 import cash.z.ecc.android.sdk.internal.model.JniAccount
 import cash.z.ecc.android.sdk.internal.model.JniAccountUsk
 import cash.z.ecc.android.sdk.internal.model.JniBlockMeta
@@ -60,6 +61,43 @@ class RustBackend private constructor(
             }
         }
         return cacheClearResult && dataClearResult
+    }
+
+    /**
+     * Validates that [height] is within the unsigned 32-bit range accepted by the Rust layer
+     * before it crosses the JNI boundary.
+     *
+     * @throws IllegalArgumentException if [height] is outside the valid UInt range.
+     */
+    private fun requireValidHeight(
+        height: Long,
+        name: String
+    ) {
+        require(height.isInUIntRange()) { "$name $height is outside of allowed UInt range" }
+    }
+
+    /**
+     * Validates that [value] is nonnegative before it crosses the JNI boundary.
+     *
+     * @throws IllegalArgumentException if [value] is negative.
+     */
+    private fun requireNonNegative(
+        value: Long,
+        name: String
+    ) {
+        require(value >= 0) { "$name $value must be nonnegative" }
+    }
+
+    /**
+     * Validates that [value] is nonnegative before it crosses the JNI boundary.
+     *
+     * @throws IllegalArgumentException if [value] is negative.
+     */
+    private fun requireNonNegative(
+        value: Int,
+        name: String
+    ) {
+        require(value >= 0) { "$name $value must be nonnegative" }
     }
 
     //
@@ -209,6 +247,7 @@ class RustBackend private constructor(
         protocol: Int,
         outputIndex: Int
     ) = withContext(SdkDispatchers.DATABASE_IO) {
+        requireNonNegative(outputIndex, "outputIndex")
         getMemoAsUtf8(
             dataDbFile.absolutePath,
             txId,
@@ -239,6 +278,7 @@ class RustBackend private constructor(
 
     override suspend fun findBlockMetadata(height: Long) =
         withContext(SdkDispatchers.DATABASE_IO) {
+            requireValidHeight(height, "height")
             findBlockMetadata(
                 fsBlockDbRoot.absolutePath,
                 height
@@ -247,6 +287,7 @@ class RustBackend private constructor(
 
     override suspend fun rewindBlockMetadataToHeight(height: Long) =
         withContext(SdkDispatchers.DATABASE_IO) {
+            requireValidHeight(height, "height")
             rewindBlockMetadataToHeight(
                 fsBlockDbRoot.absolutePath,
                 height
@@ -264,9 +305,12 @@ class RustBackend private constructor(
 
     /**
      * Rewinds the data database to at most the given height.
+     *
+     * @throws IllegalArgumentException if [height] is outside the valid UInt range.
      */
     override suspend fun rewindToHeight(height: Long): JniRewindResult =
         withContext(SdkDispatchers.DATABASE_IO) {
+            requireValidHeight(height, "height")
             rewindToHeight(
                 dataDbFile.absolutePath,
                 height,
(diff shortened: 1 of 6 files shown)


backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt
```diff
@@ -62,6 +63,43 @@ class RustBackend private constructor(
         return cacheClearResult && dataClearResult
     }
 
+    /**
+     * Validates that [height] is within the unsigned 32-bit range accepted by the Rust layer
+     * before it crosses the JNI boundary.
+     *
+     * @throws IllegalArgumentException if [height] is outside the valid UInt range.
+     */
+    private fun requireValidHeight(
```
noop-sk: **Duplicated validation helpers**

`requireValidHeight`/`requireNonNegative` are copy-pasted verbatim in both `RustBackend` and `FakeRustBackend`. Understood `backend-lib` can't depend on `sdk-lib` for a shared model type, but these could still live in one internal utility (e.g. alongside `isInUIntRange()` in `NumberExt.kt`) shared by both `Backend` implementations, rather than risking the two copies drifting apart.

backend-lib/src/test/java/cash/z/ecc/android/sdk/internal/jni/FakeRustBackendValidationTest.kt
```diff
@@ -0,0 +1,57 @@
...
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertFailsWith
+import kotlin.test.assertIs
+
+/**
+ * Exercises the [Backend][cash.z.ecc.android.sdk.internal.Backend] boundary argument validation
+ * against [FakeRustBackend], which is constructible without the native library.
+ */
+class FakeRustBackendValidationTest {
```
noop-sk, line 14: **Coverage only exercises 2 of the 8 guarded call sites**

Tests here cover `rewindToHeight` (height boundary) and `putUtxo` (negative index only). The other 6 wired-in call sites — `getBranchIdForHeight`, `updateChainTip`, `findBlockMetadata`, `rewindBlockMetadataToHeight`, `scanBlocks` (`fromHeight`+`limit`), `putSubtreeRoots` (3 indices), `getMemoAsUtf8` (`outputIndex`), and `putUtxo`'s own `height` — have no test asserting the guard actually fires. A wiring slip at any of those (wrong parameter validated, or a call accidentally left unguarded) would go undetected.

noop-sk (commented):
Reviewed via the JNI-boundary/app-consumer lens, alongside #2197 and #2199 (same batch). This is clean, narrowly-scoped Kotlin-only hardening — `require()` guards at the raw `Backend`/`RustBackend` boundary, mirrored identically in `FakeRustBackend`, correctly reusing the existing `Long.isInUIntRange()` extension. Not blocking, three things worth a look:

**Scope gap (informational, not a live vulnerability).** Three other same-shaped parameters at this boundary were left out: `decryptAndStoreTransaction`'s `minedHeight: Long?`, `createAccount`/`importAccountUfvk`'s `recoverUntil: Long?`, and `importAccountUfvk`'s `zip32AccountIndex: Long?`. Checked the Rust side (`lib.rs`) for all three — they're already checked conversions there, so no live issue — but worth a conscious call on whether a fast-follow should extend the same first-line-of-defense treatment for consistency with this PR's own stated scope.

**CHANGELOG conflict heads-up.** #2197, #2199, and this PR all insert a new bullet at the same anchor (right after the MOB-1723 line, before the `[3.1.0]` header) on the same base branch. Whichever two merge second will hit a trivial-but-real git conflict there — worth flagging to whoever sequences the merges.

Left one inline note on the duplicated validation helpers and one on test coverage.

backend-lib/src/main/java/cash/z/ecc/android/sdk/internal/jni/RustBackend.kt
nesence-m: Moved `requireValidHeight`/`requireNonNegative` into `NumberExt.kt` as internal top-level functions alongside `isInUIntRange`, and deleted both private copies in `RustBackend` and `FakeRustBackend` in favor of importing the shared ones. See 99ad7206.

backend-lib/src/test/java/cash/z/ecc/android/sdk/internal/jni/FakeRustBackendValidationTest.kt
nesence-m: Extended coverage to every remaining guarded site on `FakeRustBackend`: `getBranchIdForHeight`, `updateChainTip`, `findBlockMetadata`, `rewindBlockMetadataToHeight`, `scanBlocks` (`fromHeight` and `limit`), `putSubtreeRoots` (all three start indices), `getMemoAsUtf8` (`outputIndex`), and `putUtxo`'s `height` parameter, plus boundary acceptance (0 / UInt max) for the sites that do not stub out to `error(...)`. See 99ad7206.

nesence-m:
Addressed both inline items in 99ad7206: moved the duplicated `requireValidHeight`/`requireNonNegative` helpers into `NumberExt.kt` as shared internal functions, and extended `FakeRustBackendValidationTest` to cover all 8 guarded call sites.

On the scope gap you flagged (`decryptAndStoreTransaction.minedHeight`, `recoverUntil`, `zip32AccountIndex`): leaving these for a fast-follow. The Rust side already does checked conversions on these, so this PR stays at its stated scope rather than growing.

Also noting the CHANGELOG-anchor conflict with #2197/#2199 — trivial, whoever merges later can resolve it at rebase.

noop-sk (approved):
Verified both original findings are genuinely fixed, not just claimed: the duplicated helpers are now a single shared implementation in NumberExt.kt (confirmed both RustBackend and FakeRustBackend just import and call it), and the validation test file grew from 4 to 23 cases covering all 8 originally-unguarded call sites — cross-checked each site by name. The nullable-param scope gap and the shared CHANGELOG-anchor conflict with #2197/#2199 are correctly acknowledged as accepted/deferred rather than silently dropped. CI green, no new issues found in the refactor itself.

PR #2202, opened 2026-08-24
Surface setupBundles' dropped dust-note count through the SDK

dannywillems:
zcash_voting's BundleLayout/ChunkResult already compute dropped_count
(notes whose bundle fell below BALLOT_DIVISOR and was discarded), but the
JNI layer threw it away when building JniBundleSetupResult, so callers of
computeBundleSetup/setupBundles had no way to know dust had silently
reduced their eligible voting weight. Thread it through: Rust's
bundle_setup_from_notes now returns a named NoteBundleSetup struct
(instead of a positional tuple) carrying dropped_count, the JNI ctor
signature grew an Int, and JniBundleSetupResult/VotingBundleSetupResult
gained a droppedCount field mapped straight through.

Fixes MOB-1560.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


dannywillems (commented):
self utACK e5fe510f2c1ba0c277ae329fd79ed422ef1a4a26

Issue #2204, opened 2026-08-24
Release zcash-android-wallet-sdk version 4.0.0


nuttycom:
This release includes support for:

* Zcash Ironwood migration
* Spend Before Sync
* Coin-holder polling

PR #2211, opened 2026-08-31
Cascade Orchard migration fixes from maint/v3.1.x into candidate/v4.0.0

noop-sk:
## Summary

Forward-cascades the Orchard migration bugfixes from `maint/v3.1.x` onto the new `candidate/v4.0.0` line (cut from `main` after #2209 merged `maint/v3.0.x` and removed Slipstream), ahead of the 4.0.0 release tracked in #2204.

Carries:
- `d4753efe` [MOB-1760] Size Keystone migration runs by signing-round capacity, not note count
- `094bb99e` MOB-1750: lower the Orchard migration dust threshold to 0.0001 ZEC
- `e4be0926` Add `migratableOrchardTotal()`: per-note dust total per Kris Nuttycombe's formula
- `0324548a` Derive `MIGRATION_DUST_THRESHOLD_ZATOSHI` from `MARGINAL_FEE` instead of a bare literal
- `da9aef46` Fix `FakeTypesafeMigrationBackend` missing `migratableOrchardTotal` override

## What was intentionally left out

`maint/v3.1.x` also carries `77961391` (adapt to `zcash_pool_migration` 0.1.0 stable's `MigrationCrypto` shape). Not cherry-picked here — `main`/`candidate/v4.0.0` already contains an equivalent, independently-written adaptation to the same 0.1.0 API (infallible `orchard_fvk()`, no spending key held on `Backend`) from the librustzcash bump that landed as part of #2209. Picking it would have reintroduced the old shape and conflicted with the current one.

The rest of `maint/v3.1.x`'s delta over `main` (PRO-97 Sapling params race fix, MOB-1764/1765 JNI hardening, MOB-1723 typed insufficient-funds) is unrelated to migration and intentionally not included in this PR.

## Conflict resolution

One conflict, in `backend-lib/src/main/rust/migration_engine.rs`: MOB-1760's `is_keystone` needs the `account_row` binding that `main`'s independently-adapted `Backend::new` had inlined away. Resolved by keeping the `account_row` binding and deriving both `orchard_fvk` and `is_keystone` from it, ahead of the (unchanged) `PoolMigrations::for_account` store setup.

## Verification

- `cargo check` on `backend-lib` passes clean.
- `cargo test --lib migration` — 55 passed, 0 failed (15 ignored, require a real wallet DB), including `run_sizing_for_tests::keystone_accounts_use_signer_capacity` / `non_keystone_accounts_use_the_zodl_note_cap`.
- No Slipstream dependency in any of the picked commits (confirmed via diff inspection) — safe to carry onto the Slipstream-free `main`/`candidate/v4.0.0`.

### DIFF
--- a/backend-lib/src/main/rust/migration.rs
+++ b/backend-lib/src/main/rust/migration.rs
@@ -46,11 +46,12 @@ use rusqlite::Connection;
 use std::num::NonZeroUsize;
 use std::ptr;
 
-use zcash_client_backend::data_api::wallet::input_selection::LockFilter;
+use zcash_client_backend::data_api::wallet::input_selection::{LockFilter, LockedInputPolicy};
 use zcash_client_backend::data_api::{InputSource, OutputLockStore, WalletRead};
 use zcash_client_backend::wallet::{LockOwner, OutputRef};
 use zcash_client_sqlite::AccountUuid;
 use zcash_client_sqlite::util::SystemClock;
+use zcash_primitives::transaction::fees::zip317::MARGINAL_FEE;
 use zcash_protocol::consensus::{
     BLOCKS_PER_HOUR, BlockHeight, Network, NetworkConstants, Parameters,
 };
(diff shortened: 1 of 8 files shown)


backend-lib/src/main/rust/migration.rs
```diff
@@ -3260,18 +3343,74 @@ pub extern "C" fn Java_cash_z_ecc_android_sdk_internal_jni_MigrationRustBackend_
...
+                LockFilter::Policy(&LockedInputPolicy::Exclude),
+            )
+            .map_err(|e| anyhow!("Error reading migratable Orchard total: {}", e))?;
+
+        let marginal_fee = u64::from(MARGINAL_FEE);
+        let total: u64 = received
+            .orchard()
+            .iter()
+            .map(|rn| rn.note().value().inner().saturating_sub(marginal_fee))
+            .sum();
```
nuttycom: This is incorrect in the case that there is only a single Orchard note; a better approach here would be to use the ZIP 317 `FeeRule` from `zcash_primitives` to compute the fee. In the one-note case, this will underestimate the fee.

nuttycom (approved):
utACK a0989050030b2b7537187e15bf32bf64b8999c31

nuttycom (approved):
re-utACK (clippy fix)

nuttycom (approved):

